PR #33, opened 2024-05-08
chore: clear warnings, bump up datafusion version to 37.1.0

mesejo:


src/common/data_type.rs
```diff
@@ -309,6 +313,7 @@ impl DataTypeMap {
             ScalarValue::DurationMillisecond(_) => Ok(DataType::Duration(TimeUnit::Millisecond)),
             ScalarValue::DurationMicrosecond(_) => Ok(DataType::Duration(TimeUnit::Microsecond)),
             ScalarValue::DurationNanosecond(_) => Ok(DataType::Duration(TimeUnit::Nanosecond)),
+            _ => todo!(),
```
dlovell, line 316: is there an alternative to panicking here?

mesejo: Yes, change it to return Err

dlovell (approved):

PR #40, opened 2024-05-14
chore(predict): objective function is being deprecated

hussainsultan:
loss function used by XGBoost is `reg::squarederror`, formerly called `reg:linear`


mesejo (approved):
LGTM

PR #49, opened 2024-05-26
feat: add better snowflake caching

dlovell:
feat: add snowflake_utils
test: add test for snowflake cache invalidation

### DIFF
--- a/python/letsql/backends/let/tests/conftest.py
+++ b/python/letsql/backends/let/tests/conftest.py
@@ -1,11 +1,27 @@
+import os
+from contextlib import contextmanager
 from typing import Any
 
 import ibis
 import pandas as pd
 import pandas.testing as tm
 import pytest
+import sqlglot as sg
+import sqlglot.expressions as sge
+from ibis.util import gen_name
 
 import letsql as ls
+import letsql.common.utils.snowflake_utils as SU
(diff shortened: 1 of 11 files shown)


python/letsql/backends/let/tests/conftest.py
```diff
@@ -1,11 +1,27 @@
...
 import ibis
 import pandas as pd
 import pandas.testing as tm
 import pytest
+import sqlglot as sg
+import sqlglot.expressions as sge
+from ibis.util import gen_name
 
 import letsql as ls
+import letsql.common.utils.snowflake_utils as SU
```
dlovell, line 14: @mesejo this import ultimately requires `snowflake.connector` to be present and will fail if not. should this be restructured so the test suite run doesn't require `snowflake.connector`?

dlovell:
@mesejo I'll wait till #50 is merged so I can modify and rebase on it

python/letsql/backends/let/tests/conftest.py
mesejo: Yes, as you mentioned, perhaps we need to create a test matrix for each backend, and create tests only for the snowflake backend

mesejo (approved):

PR #50, opened 2024-05-26
ref(pyproject): make engines optional extras

dlovell:


.github/workflows/ci-test.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         uses: snok/install-poetry@v1
 
       - name: Poetry install
-        run: poetry install
+        run: poetry install --with="dev,test" --extras="duckdb snowflake datafusion"
```
dlovell, line 48: annoyingly, `--with` takes commas but not spaces and `--extras` takes spaces but not commas

pyproject.toml
```diff
@@ -48,30 +48,37 @@ packages = [
...
 connectorx = "^0.3.2"
-scikit-learn = "^1.4.0"
 psycopg2-binary = "^2.9.9"
 sqlalchemy = "^2.0.25"
-pytest-mock = "^3.14.0"
 pyarrow = "13.0.0"
-
+duckdb = {version = "^0.10.3", optional = true}
+snowflake-connector-python = {version = "^3.10.1", optional = true}
+datafusion = {version = "^34.0.0", optional = true}
```
mesejo, line 60: Do we need datafusion other than in tests?

mesejo (approved):
LGTM

pyproject.toml
dlovell: a user might want to have datafusion available as an engine, similar to duckdb and snowflake

PR #51, opened 2024-05-28
chore: add docs

mesejo:
This PR adds basic documentation so the user doesn't feel lost when using LETSQL


dlovell:
Is there a command for me to render / serve the docs?

dlovell:
we should add the `docs` group to the list of groups to install in `flake.nix`
```
diff --git a/flake.nix b/flake.nix
index eab92ab..e2ae9af 100644
--- a/flake.nix
+++ b/flake.nix
@@ -143,7 +143,7 @@
           src = pySrc;
           preferWheels = true;
           python = python';
-          groups = [ "dev" "test" ];
+          groups = [ "dev" "test" "docs" ];
         };
         myapp = (mkPoetryApplication (commonPoetryArgs // {
           buildInputs = pkgs.lib.optionals pkgs.stdenv.isDarwin [
```

PR #52, opened 2024-05-29
chore(nix): add letsql-quarto* commands

dlovell:


nix/commands.nix
```diff
@@ -85,8 +85,18 @@
     ${pkgs.git}/bin/git config blame.ignoreRevsFile "$ignore_revs_file"
   '';
 
+  letsql-quarto = pkgs.writeShellScriptBin "letsql-quarto" ''
+    repo_dir=$(git rev-parse --show-toplevel 2>/dev/null) || exit
+    pushd "$repo_dir/docs" || exit
+    ${pkgs.quarto}/bin/quarto "''${@}"
+  '';
+
+  letsql-quarto-render = pkgs.writeShellScriptBin "letsql-quarto-render" ''
```
mesejo, line 94: How does this line know that it is executing `quarto render`? 
Is this `"''${@}"` transforming `letsql-quarto-render` into `render`?

mesejo (approved):

nix/commands.nix
dlovell: I think you probably made this comment before I fixed the command by adding `render` as an argument just before `"''${@}"`. It was missing but is there now.

PR #60, opened 2024-05-31
feat: add array functions

mesejo:
- flatten
- range
- array_distinct
- unnest
- array_slice

### DIFF
--- a/python/letsql/tests/test_array.py
+++ b/python/letsql/tests/test_array.py
@@ -7,7 +7,7 @@ import pandas as pd
 import pytest
 
 from pytest import param
-from letsql.tests.util import assert_series_equal
+from letsql.tests.util import assert_series_equal, assert_frame_equal
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


python/letsql/tests/test_array.py
```diff
@@ -112,3 +126,70 @@ def test_array_remove(con):
...
+            id="not_null",
+        ),
+    ],
+)
+def test_array_unique(con, data, expected):
+    t = ibis.memtable(data)
+    expr = t.a.unique()
+    result = con.execute(expr).map(frozenset, na_action="ignore")
+    expected = pd.Series(expected, dtype="object").map(frozenset, na_action="ignore")
+    assert frozenset(result.values) == frozenset(expected.values)
```
dlovell, line 195: why do we use frozenset here?
could we test against the `pd.Series` result of `execute` directly?
```
In [20]: expr.execute().equals(pd.Series([[1, 3], [], [42], []]))
Out[20]: True
```

dlovell: why does `expected` contain `float`s rather than `int`s?

python/letsql/tests/test_array.py
```diff
@@ -112,3 +126,70 @@ def test_array_remove(con):
...
+def test_array_flatten(con, flatten_data, column, expected):
+    data = flatten_data[column]
+    t = ibis.memtable({column: data["data"]}, schema={column: data["type"]})
+    expr = t[column].flatten()
+    result = con.execute(expr)
+    assert_series_equal(
+        result.sort_values().reset_index(drop=True),
+        expected.sort_values().reset_index(drop=True),
+        check_names=False,
+    )
```
dlovell, line 149: could this use `pd.Series.equals` as well?
```
expr.execute().equals(pd.Series([["abc"], ["bcd"], ["def"]]))
```

python/letsql/tests/test_array.py
mesejo: Fixed: I'm using `assert_series_equal`

dlovell (approved):

PR #81, opened 2024-06-10
ls accessor

dlovell:


python/letsql/backends/let/hotfix.py
```diff
@@ -0,0 +1,165 @@
...
+    @property
+    def op(self):
+        return self.expr.op()
+
+    @property
+    def cached_nodes(self):
+        return self.op.find(CachedNode)
+
+    @property
+    def storage(self):
```
mesejo, line 37: How would this work with multiple storages?

python/letsql/backends/let/hotfix.py
```diff
@@ -0,0 +1,165 @@
...
+    def uncached_one(self):
+        if self.is_cached:
+            return self.op.parent.to_expr()
+        else:
+            return self.expr
+
+    def get_key(self):
+        if self.is_cached:
+            return self.storage.get_key(
+                self.ls_con._register_and_transform_cache_tables(
```
mesejo, line 128: Why the use of `_register_and_transform_cache_tables` in here?

mesejo (approved):

python/letsql/backends/let/hotfix.py
dlovell: currently, the hash depends on all downstream cached nodes being replaced by their cached location

python/letsql/backends/let/hotfix.py
dlovell: `storage` works like `is_cached` and only refers to the top level expr/op being referenced; `storages` / `has_cached` refers to the entire graph of this expr

PR #84, opened 2024-06-11
feat: add pyarrow udaf based on PyAggregator

mesejo:
see https://www.letsql.com/posts/xgboost-udaf-ibis/


dlovell:
are we demonstrating/testing udafs? i don't see any `group_by`s

PR #93, opened 2024-06-12
fix `ls.native_expr`

dlovell:


dlovell:
closing in favor of #94

PR #101, opened 2024-06-17
feat: make cache default path configurable

mesejo:


python/letsql/common/caching.py
```diff
@@ -105,7 +105,7 @@ class ParquetCacheStorage(CacheStorage):
     path = field(
         validator=instance_of(pathlib.Path),
         converter=abs_path_converter,
-        default=(pathlib.Path("~/.local/share/letsql")),
+        default=letsql.options.cache_default_path,
```
dlovell, line 108: rather than `default`, this `field` should use `factory` to dynamically get the `cache_default_path` value
something like `factory=functools.partial(letsql.options.get, "cache_default_path")` should work

https://www.attrs.org/en/stable/examples.html#defaults

python/letsql/backends/let/tests/test_cache.py
```diff
@@ -553,3 +555,32 @@ def test_replace_table_matching_kwargs(pg, ls_con, tmp_path):
...
 
     assert expr.ls.native_expr is not None
+
+
+def test_cache_default_path_set(pg, ls_con, tmp_path):
+    letsql.options.cache_default_path = tmp_path
+
+    storage = ParquetCacheStorage(
+        source=ls_con,
+        path=tmp_path,
```
dlovell, line 565: `path` should not be passed here so we can verify that modification of `letsql.options.cache_default_path` takes effect

mesejo: Done

python/letsql/config.py
```diff
@@ -0,0 +1,20 @@
...
+    """LETSQL configuration options
+
+    Attributes
+    ----------
+
+    cache_default_path : str
+
+    """
+
+    cache_default_path: Union[str, pathlib.Path] = pathlib.Path("~/.local/share/letsql")
```
dlovell, line 17: if you don't call `.expanduser()` on the path, this is a relative path with the top level dir of `$(pwd)/~` and not `/home/$USER`

mesejo: Should it call absolute as well?

dlovell (approved):

PR #103, opened 2024-06-18
feat(caching): add SnapshotStorage

dlovell:


mesejo (approved):

python/letsql/common/caching.py
```diff
@@ -155,3 +163,33 @@ def _put(self, key, value):
...
+                tokenized = dask.base.tokenize(expr)
+                return "-".join(("snapshot", tokenized))
+
+    @staticmethod
+    def normalize_backend(con):
+        name = con.name
+        if name in ("pandas", "duckdb", "datafusion"):
+            return (name, None)
+        else:
+            return normalize_backend(con)
```
mesejo, line 185: If the backend name is `postgres,` wouldn't this method reach `RecursionError`?

python/letsql/common/utils/dask_normalize.py
```diff
@@ -1,6 +1,25 @@
...
 
+
 dask.config.set({"tokenize.ensure-deterministic": True})
+
+
+@contextmanager
+def patch_normalize_token(*typs, f=toolz.functoolz.return_none):
+    with patch.dict(
+        dask.base.normalize_token._lookup,
+        {typ: Mock(side_effect=f) for typ in typs},
```
mesejo, line 22: Out of curiosity, why is the code using testing mocks?

mesejo (approved):

python/letsql/common/caching.py
dlovell: the `normalize_backend` it accesses is the one imported, `self.normalize_backend` **would** result in a `RecursionError`

python/letsql/common/utils/dask_normalize.py
dlovell: for `SnapshotStorage` to work, we need to remove some specific sources of non-determinism in `dask.base.normalize_token` but only during the context of specific invocations. `patch` fits this use and we already need this capability for tests.

PR #105, opened 2024-06-20
feat: api letsql api methods

mesejo:


python/letsql/backends/let/__init__.py
```diff
@@ -120,6 +120,19 @@ def replace_table(node, _, **_kwargs):
     def do_connect(
         self, config: Mapping[str, str | Path] | SessionContext | None = None
```
dlovell, line 121: should we remove `SessionContext` here?

python/letsql/config.py
```diff
@@ -1,22 +1,124 @@
...
 options = Options()
+
+
+def _backend_init():
+    if (backend := options.backend) is not None:
+        return backend
+
+    import letsql
+
+    options.backend = con = letsql.connect()
```
dlovell, line 123: should there only ever be one instance of `letsql.backends.let.Backend`?

mesejo: The user can create multiple instances of `letsql.backends.let.Backend` using the `letsql.connect` method, this instance would be the one used to execute the `api` methods. This is what ibis does, it is also a common pattern in the Python standard library. For example, the modules re and random create class instances and bound the api methods to those intances.

dlovell: are we testing using multiple `letsql.backends.let.Backend` instances?

mesejo: Done

dlovell (approved):

PR #110, opened 2024-07-08
chore: add unit tests based on workflow diagram

mesejo:
- create new unit test to detect and fix problems obtained from the workflow diagram analysis

- fixes the problem with using set or equals for testing equality among backends, currently the hashing of backends is based on string of the name of the class, which renders all backends equals

- fix problem normalize_duckdb_databasetable to handle SEQ_SCAN and duckdb tables

- fix node replacement, now it is a LETSQL op. Before calling .cache at the end of expr causes the expr source to become the source of the cache, leaving letsql and entering into ibis.

- also add tests to verify that execution happens in the native backend

workflow diagram reference: https://excalidraw.com/#json=rmcDKw_OB0hoZ7B2O4NKO,TXGXMJ38zb1Ly3NwXgaVcw

### DIFF
--- a/python/letsql/backends/let/tests/test_execute.py
+++ b/python/letsql/backends/let/tests/test_execute.py
@@ -1,3 +1,4 @@
+import itertools
 import random
 from pathlib import Path
 
@@ -5,6 +6,7 @@ import ibis
 import numpy as np
 import pandas as pd
 import pytest
+from ibis.expr import types as ir
 from pytest import param
 
 import letsql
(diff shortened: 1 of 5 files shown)


python/letsql/backends/let/tests/test_execute.py
```diff
@@ -85,13 +87,38 @@ def dirty_duckdb_con(csv_dir):
...
+
+
+@pytest.fixture(scope="function")
+def parquet_batting(parquet_dir):
+    return parquet_dir / "batting.parquet"
+
+
+@pytest.fixture(scope="function")
+def ls_batting(parquet_batting):
+    return letsql.connect().read_parquet(parquet_batting)
```
dlovell, line 114: could/should this be `ls.read_parquet`?

dlovell (approved):

PR #115, opened 2024-07-15
feat: wrap ibis backends

mesejo:
- add __getattr__ to letsql module, try to load letsql backends, otherwise default to ibis
	- use letsql postgres backend with _cached method for deferred cache
- update ruff to the latest version (to check, for instance, and type equality)
- use pytest-snapshot to test for values that do not change frequently. The added benefits of snapshot testing are that:
    - They are easy to create.
    - They are easy to update when the expected value of a test changes.    
  
  more on: https://pypi.org/project/pytest-snapshot/


requirements-dev.txt
```diff
@@ -55,12 +55,13 @@ pyarrow==13.0.0 ; python_version >= "3.10" and python_version < "4.0"
 pygments==2.17.2 ; python_version >= "3.10" and python_version < "4.0"
 pytest-cov==4.1.0 ; python_version >= "3.10" and python_version < "4.0"
 pytest-mock==3.14.0 ; python_version >= "3.10" and python_version < "4.0"
+pytest-snapshot==0.9.0 ; python_version >= "3.10" and python_version < "4.0"
```
dlovell, line 58: nice find :smiley: !

dlovell (approved):

PR #170, opened 2024-07-24
Predict udf

hussainsultan:
add `register_xgb_model` so that we can use the predict method with ibis expressions.


mesejo (approved):
LGTM

PR #208, opened 2024-08-12
feat/deferred pandas read

dlovell:


python/letsql/__init__.py
mesejo, line 1: We can add our own Pandas backend (with no need for hot-fixing). We would need to add an entry-point (plugin) for the letsql Pandas backend in the `pyproject.toml` the same way it was done for Snowflake and Postgres.

python/letsql/backends/pandas/executor.py
mesejo, line 1: Could we inherit here from the ibis Pandas backend, or are we doing this to prepare for the possible deletion?

dlovell:
close in favor of #315

PR #209, opened 2024-08-12
feat/no expr registration

dlovell:
demonstration script

```
import letsql as ls


pg = ls.postgres.connect(
    # FIXME: use dyndns to point examples.letsql.com to the gcp sql host
    host="34.135.241.141",
    user="letsql",
    password="letsql",
    database="letsql",
)
db = ls.duckdb.connect()


batting = pg.table("batting")
awards_players = db.register(
    ls.config.options.pins.get_path("awards_players"),
    table_name="awards_players",
)
left = batting[batting.yearID == 2015]
right = awards_players[awards_players.lgID == "NL"].drop("yearID", "lgID")
expr = left.join(right, ["playerID"], how="semi")


result = expr.execute()
print(result)
result = expr.to_pyarrow()
print(result)
result = expr.to_pyarrow_batches().read_pandas()
print(result)
print(tuple(dt.args for dt in expr.ls.native_dts))
```

### DIFF
--- a/examples/multi_engine.py
+++ b/examples/multi_engine.py
@@ -1,3 +1,5 @@
+import ibis
+
 import letsql as ls
 
 
--- a/examples/multi_engine.py
+++ b/examples/multi_engine.py
@@ -1,6 +1,5 @@
 import letsql as ls
 
-# this example also shows the no need for registering tables or expression directly in LETSQL
 
 pg = ls.postgres.connect_examples()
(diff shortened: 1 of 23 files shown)


mesejo (commented):
**Note:** What happens if one of the tables of the Expr to be executed is an InMemoryTable?

mesejo (approved):

PR #230, opened 2024-08-22
feat: reading from csv in HTTP, add basic s3 support

mesejo:


dlovell:
@mesejo LGTM. There are some issues created with caching that need to be addressed, but I'll create issues once this is merged.

dlovell (approved):

dlovell:
@mesejo maybe we should consider adding `xfail`ing tests for replicability of schema inference on urls

PR #275, opened 2024-09-24
chore: use default connection to avoid full slot error

mesejo:


mesejo:
The testing of examples is temporarily disabled

PR #278, opened 2024-10-06
ref: reparametrize caching into storage and invalidation strategy

dlovell:


python/letsql/backends/let/tests/snapshots/test_cache/test_udaf_caching/test_udaf_caching.txt
```diff
@@ -1 +1 @@
-letsql_cache-34a69b04834861e2f3d6fb81c8d607a7
\ No newline at end of file
+letsql_cache-letsql_cache-34a69b04834861e2f3d6fb81c8d607a7
```
mesejo, line 1: Should we use a modification strategy specific prefix?

mesejo (approved):

PR #279, opened 2024-10-06
feat: add pipe operator

mesejo:
Add the pipe operator with the ability to create partial ibis expressions, for example, with this PR now is possible to do:
```python
pipeline = (
        left
        | join(right, "a")
        | select("a", bx="b", cx="c")
        | mutate(a2=_.a * 2, bcx=_.bx + _.cx)
        | limit(5)
    )
```

as well as (the equivalent):

```python
pipeline = left | join(right, "a").select("a", bx="b", cx="c").mutate(
        a2=_.a * 2, bcx=_.bx + _.cx
    ).limit(5)
```

also the method `sql` is supported:

```python
pipeline = (
        penguins
        | sql(
            """
        SELECT island, mean(bill_length_mm) AS avg_bill_length
        FROM penguins
        GROUP BY 1
        ORDER BY 2 DESC
        """
        )
        | head(4)
    )
```

### DIFF
--- a/python/letsql/pipe/README.md
+++ b/python/letsql/pipe/README.md
@@ -67,7 +67,7 @@ multiple partial expressions in one large pipeline, for example:
     )
 ```
 
-Notice the use of each `|` pipe to "glue" together each partial expression.
+Notice the use of each `|` pipe to "glue" together each partial expression, it should be used for that. 
 
 ### Proposal: Pipe Operator should receive multiple arguments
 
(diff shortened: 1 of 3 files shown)


hussainsultan:
One behavior to note:
```python
 df = ibis.memtable(ls.examples.penguins.load())
 df.pipe(con.register, "df") | sql("select species from df") # this works
 expr = df | sql("select species from df")
```
```python
InvalidInputException: Invalid Input Error: Python Object "df" of type "Table" found on line "<ipython-input-34-c3e3fc7357e3>:1" not suitable for replacement scans.
Make sure that "df" is either a pandas.DataFrame, duckdb.DuckDBPyRelation, pyarrow Table, Dataset, RecordBatchReader, Scanner, or NumPy ndarrays with supported format

```

PR #282, opened 2024-10-11
wip: added RemoteTable

mesejo:


python/letsql/common/utils/dask_normalize_expr.py
```diff
@@ -55,7 +55,7 @@ def normalize_datafusion_databasetable(dt):
     if dt.source.name not in ("datafusion", "let"):
         raise ValueError
     ep_str = str(dt.source.con.table(dt.name).execution_plan())
-    if ep_str.startswith(("ParquetExec:", "CsvExec:")):
+    if ep_str.startswith(("ParquetExec:", "CsvExec:", "CustomExec")):
```
dlovell, line 58: I think we want the process the remote expr, not the `ep_str` in this case

python/letsql/backends/let/__init__.py
```diff
@@ -310,10 +343,12 @@ def fn(node, _, **kwargs):
     def _to_sqlglot(
         self, expr: ir.Expr, *, limit: str | None = None, params=None, **_: Any
     ):
-        op = expr.op()
-        out = op.map_clear(replace_cache_table)
+        expr = self._register_and_transform_cache_tables(expr)
```
dlovell, line 346: why did we move this here?

python/letsql/backends/let/__init__.py
```diff
@@ -290,8 +293,38 @@ def _get_source(self, expr: ir.Expr):
...
+                    if v in tables:
+                        name = f"{v.name}_{next(count)}"
+                        kwargs[k] = DatabaseTable(
+                            name,
+                            schema=v.schema,
+                            source=v.source,
+                            namespace=v.namespace,
+                        )
+                        remote: RemoteTable = tables[v]
+                        batches = remote.remote_expr.to_pyarrow_batches()
```
dlovell, line 311: I think this doesn't address the case where there are multiple passes on the table within the single engine expr.

mesejo:
Closed via #304

PR #315, opened 2024-10-25
deferred read

dlovell:


examples/deferred_read_csv.py
```diff
@@ -0,0 +1,49 @@
+import letsql as ls
+from letsql.common.utils.defer_utils import (
+    deferred_read_csv,
```
mesejo, line 3: Should we add `deferred_read_csv` as an api function?

dlovell: we can, though i would like it to be `read_csv` with "deferred" implied

related, we should not introspect on backends to detect what the csv/parquet source is and instead require users to use `deferred_read_{csv,parquet}` (or eventual api function `read_{csv,parquet}`) if they want to be able to cache the expr

python/letsql/backends/let/hotfix.py
```diff
@@ -126,7 +127,12 @@ def exists(self):
 def _letsql_find_backend(self, *, use_default=True):
     # FIXME: push this into LETSQLAccessor
     try:
-        current_backend = self._find_backend._original(self, use_default=use_default)
+        if tuple(op.source for op in self.op().find(Read)):
+            current_backend = letsql.options.backend
```
mesejo, line 131: I think `letsql.options.backend` may be null if it has not been initialized; in general, it is safer to do:
```python
    from letsql.config import _backend_init
    current_backend = _backend_init() 
```

python/letsql/common/utils/dask_normalize_expr.py
```diff
@@ -167,6 +171,44 @@ def normalize_module(module):
...
+def normalize_read(read):
+    path = dict(read.read_kwargs).get("path") or dict(read.read_kwargs).get("source")
+    if isinstance(path, (str, pathlib.Path)):
+        path = str(path)
+        if path.startswith("http") or path.startswith("https:"):
+            import requests
+
+            resp = requests.head(path)
+            resp.raise_for_status()
+            dct = {
```
mesejo, line 184: Is the variable `dct` a dictionary?

mesejo (approved):

python/letsql/common/utils/dask_normalize_expr.py
dlovell: yes

dlovell: rather, it should be (thought doesn't need to be)

PR #316, opened 2024-10-25
chore: use letsql execute/to_pyarrow/to_pyarrow_batches

mesejo:


dlovell (approved):
Thanks for this @mesejo !

PR #318, opened 2024-10-26
chore: ci check for right install

mesejo:
This PR only uses `pip` and `venv` to verify the library is installed correctly. 
1. The first step is to install `pytest` and `maturin`
2. Execute `maturin develop` to install letsql in development mode
3. Execute some of the examples to verify the correct installation 

This will fail if a `dependency` is specified in `poetry` and not in the `dependencies`.

### DIFF
--- a/.github/workflows/ci-test-library.yml
+++ b/.github/workflows/ci-test-library.yml
@@ -47,13 +47,19 @@ jobs:
         run: python -m venv .venv
 
       - name: install dependencies
-        run: .venv/bin/python -m pip install pytest maturin
+        run: |
+          source .venv/bin/activate
+          python -m pip install pytest maturin
 
       - name: develop
-        run: .venv/bin/python -m maturin develop
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test-library.yml
```diff
@@ -35,20 +35,29 @@ jobs:
...
-        uses: snok/install-poetry@v1
+      # check how to use virtual environment https://stackoverflow.com/q/74668349/4001592
+      - name: create virtual environment
+        run: python -m venv .venv
 
-      - name: Poetry install
-        run: poetry install
-        working-directory: ${{ github.workspace }}
+      - name: install dependencies
+        run: .venv/bin/python -m pip install pytest maturin
```
dlovell, line 50: @mesejo be aware that directly running a binary in the venv without activation will have issues if python needs access to any of the binaries provided in `.venv/bin`

dlovell (approved):

PR #346, opened 2024-11-12
chore: use bare datafusion

mesejo:


python/letsql/backends/datafusion/__init__.py
```diff
@@ -1,871 +1,5 @@
...
-
-        self._import_pyarrow()
-        import pyarrow.parquet as pq
-
-        with letsql.to_pyarrow_batches(expr, params=params) as batch_reader:
-            with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
-                for batch in batch_reader:
-                    writer.write_batch(batch)
+class Backend(IbisDatafusionBackend):
+    pass
```
dlovell, line 5: how is this used?

dlovell:
One of the reasons I didn't realize I wasn't actually using datafusion but our internal engine was that the underlying con didn't let me know. Can we change the name?

```
In [1]: import letsql as ls
l
In [2]: ls_con = ls.connect()

In [3]: df_con = ls.datafusion.connect()

In [4]: (type(ls_con.con), type(df_con.con))
Out[4]: (datafusion.SessionContext, datafusion.context.SessionContext)
```

pyproject.toml
```diff
@@ -54,7 +54,7 @@ duckb = [
     "duckdb>=0.10.3,<2 ; python_version >= '3.10' and python_version < '4.0'"
 ]
 datafusion = [
-    "datafusion>=34.0.0,<41; python_version >= '3.10' and python_version < '4.0'"
+    "datafusion>=34.0.0,<43; python_version >= '3.10' and python_version < '4.0'"
```
dlovell, line 57: why do we have a constraint on the datafusion version?

mesejo: I think is better to have it this way, having an open range means that the user may install a version of datafusion we haven't yet tested

dlovell: Ibis has a wider range than us
`datafusion = { version = ">=0.6,<41", optional = true }`
https://github.com/ibis-project/ibis/blob/9.5.0/pyproject.toml#L62
Can you point me to the version specific functionality we have?

dlovell (approved):

PR #375, opened 2024-11-22
ref: udf.agg.pandas_df: enable use as a decorator

dlovell:
and accept schema rather than expr


mesejo (commented):
This udf can be used even if the columns name do not match that of the schema?

mesejo (approved):

PR #389, opened 2024-12-05
feat: add segmentation

mesejo:


python/letsql/backends/let/__init__.py
```diff
@@ -242,12 +244,28 @@ def to_pyarrow_batches(
...
+        )
+
+        def clean_up():
+            for table, con in created.items():
+                try:
+                    con.drop_table(table)
+                except Exception:
+                    con.drop_view(table)
+
+        return rbr_wrapper(reader, clean_up)
```
dlovell, line 265: can we have `execute` and `to_pyarrow` use `to_pyarrow_batches`?

dlovell: if we do that, we can remove `_get_backend_and_expr`

python/letsql/backends/duckdb/__init__.py
```diff
@@ -17,5 +18,16 @@ def execute(
         limit: str | None = "default",
         **_: Any,
     ) -> Any:
-        expr = expr.op().replace(RemoteTableReplacer()).to_expr()
+        expr = expr.op().replace(replace_fix(RemoteTableReplacer())).to_expr()
```
dlovell, line 21: unless there's a compelling reason not to, I think all `execute` and `to_pyarrow` definitions should work by way of `to_pyarrow_batches`

python/letsql/common/utils/caching_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        else:
+            backends.add(table.source)
+
+    return (
+        backends.pop(),
+        has_unbound,
+    )  # TODO what happens if it has more than one backend
+
+
+def exists(node: CachedNode):
```
dlovell, line 60: please look at `LETSQLAccessor`'s `exists` implementation
if there's duplication between `caching_utils` and `LETSQLAccessor`, let's have `LETSQLAccessor` use `caching_utils`

python/letsql/backends/let/hotfix.py
```diff
@@ -40,7 +43,20 @@ def op(self):
 
     @property
     def cached_nodes(self):
-        return self.op.find(CachedNode)
+        def _find(node):
+            cached = node.find((CachedNode, RemoteTable))
+            if not cached:
+                yield None
+            else:
+                for no in cached:
```
dlovell, line 51: this function can start from `for no in node.find((CachedNode, RemoteTable))`, removing what is above, and then you can remove the `filter(None, ...`

python/letsql/common/caching.py
```diff
@@ -136,17 +137,24 @@ def get_key(self, expr: ir.Expr):
...
         else:
             return normalize_backend(con)
 
     @staticmethod
     def normalize_databasetable(dt):
-        keys = ["name", "schema", "source", "namespace"]
-        return dask.tokenize._normalize_seq_func(
-            (key, getattr(dt, key)) for key in keys
-        )
+        print("in snapshot strategy")
```
dlovell, line 147: is this a leftover debugging `print`?

dlovell:
let's standardize on either `import letsql` or `import letsql as ls`

python/letsql/backends/duckdb/__init__.py
mesejo: I agree, solving it in the fix/to_pyarrow_batches PR

python/letsql/backends/let/__init__.py
mesejo: Agreed

python/letsql/common/utils/caching_utils.py
mesejo: This exists is a leftover, is actually not used

mesejo:
> let's standardize on either `import letsql` or `import letsql as ls`

Opened PR #403 forcing the standard

dlovell:
Something is broken with caching

Running
```
import operator

import toolz

import letsql as ls
from letsql.common.caching import ParquetSnapshot
from letsql.common.utils.defer_utils import deferred_read_parquet
from letsql.expr.relations import into_backend


rate_to_rate_str = toolz.compose(operator.methodcaller("replace", ".", "p"), str)


def rate_to_parquet(flight, rate):
    base_url = f"https://nasa-avionics-data-ml.s3.us-east-2.amazonaws.com/Tail_652_1_parquet"
    filename = f'{flight}.{rate_to_rate_str(rate)}.parquet'
    return f"{base_url}/{filename}"


def asof_join_flight_data(flight, airborne_only=True):
    """Create an expression for a particular flight's data """

    con = ls.connect()
    ts = (
        deferred_read_parquet(con, parquet_path, rate_to_rate_str(rate))
        .mutate(flight=ls.literal(flight))
        for rate, parquet_path in (
            (rate, rate_to_parquet(flight, rate))
            for rate in sorted(rates, reverse=True)
        )
    )
    db_con = ls.duckdb.connect()
    (expr, *others) = (into_backend(t, db_con, name=f"flight-{flight}-{t.op().parent.name}") for t in ts)
    for other in others:
        expr = expr.asof_join(other, on="time").drop(["time_right", "flight_right"])
    if airborne_only:
        expr = expr[lambda t: t.GS != 0]
    return expr


tail = "Tail_652_1"
flight = "652200101120916"
rates = (0.25, 1.0, 2.0, 4.0, 8.0, 16.0)
expr = asof_join_flight_data(flight)
storage = ParquetSnapshot()
cached = expr.cache(storage=storage)
assert not storage.cache.exists(expr)
ls.execute(cached.count())
assert storage.cache.exists(expr)
```

results in
```
/nix/store/xjz78j087rn070fa5yq49y7wg1k8wpg3-letsql-dev-env/lib/python3.10/site-packages/ibis/expr/types/relations.py:685: FutureWarning: Selecting/filtering arbitrary expressions in `Table.__getitem__` is deprecated and will be removed in version 10.0. Please use `Table.select` or `Table.filter` instead.
  warnings.warn(
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[1], line 49
     47 assert not storage.cache.exists(expr)
     48 ls.execute(cached.count())
---> 49 assert storage.cache.exists(expr)

AssertionError: 
```

mesejo:
@dlovell 

This is happening because we are putting a RemoteTable before cross-source caching, any of these assertions work:
```python
assert storage.cache.exists(cached.op().replace(replace_fix(transform_cached_node)).parent)
assert storage.cache.exists(uncached(cached.op()))
```

I'm working on a fix

mesejo:
Now, it is possible to do the following:

```python
storage.exists(expr)
```

dlovell (approved):

PR #398, opened 2024-12-09
fix: to_pyarrow_batches

### LINKED ISSUE #397: `into_backend` does not work for `to_pyarrow_batches`
This will raise because `df.empty` is `True`

```
import letsql as ls
from letsql.expr.relations import (
    into_backend,
)
from letsql.tests.test_into_backend import (
    make_merged,
)


def get_trino_t():
    return (
        ls.trino.connect(database="tpch", schema="sf1")
        .table("orders")
        .cast({"orderdate": "date"})
    )


trino_table = get_trino_t()
db_con = ls.duckdb.connect()
df = (
    get_trino_t().head(10_000)
    .pipe(into_backend, db_con)
    .pipe(make_merged)
    .pipe(ls.to_pyarrow_batches)
    .read_pandas()
)
assert not df.empty
```

mesejo:
closes #397

### DIFF
--- a/python/letsql/expr/relations.py
+++ b/python/letsql/expr/relations.py
@@ -180,28 +180,29 @@ def register_and_transform_remote_tables(expr):
         replicas = SafeTee.tee(batches, count)
         batches_table[arg] = (schema, list(replicas))
 
+    def mark_remote_table(node):
+        schema, batchess = batches_table[node]
+        name = f"{node.name}_{len(batchess)}"
+        result = MarkedRemoteTable(
+            name,
+            schema=node.schema,
+            source=node.source,
+            namespace=node.namespace,
+            remote_expr=node.remote_expr,
+        )
+        reader = pa.RecordBatchReader.from_batches(schema, batchess.pop())
+        node.source.register(reader, table_name=name)
+        created[name] = node.source
+        return result
+
     def replacer(node, _, **kwargs):
         if isinstance(node, Relation):
             updated = {}
             for k, v in list(kwargs.items()):
                 try:
                     if v in batches_table:
-                        schema, batchess = batches_table[v]
-                        name = f"{v.name}_{len(batchess)}"
-                        kwargs[k] = MarkedRemoteTable(
-                            name,
-                            schema=v.schema,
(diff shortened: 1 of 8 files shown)


python/letsql/expr/relations.py
```diff
@@ -236,3 +163,72 @@ def make_unbound_dt(self):
...
+                        name = f"{v.name}_{next(remote.count)}"
+                        kwargs[k] = MarkedRemoteTable(
+                            name,
+                            schema=v.schema,
+                            source=v.source,
+                            namespace=v.namespace,
+                            remote_expr=v.remote_expr,
+                        )
+
+                        batches = get_batches(remote.remote_expr)
```
dlovell, line 205: I think we should call `get_batches` only once

python/letsql/expr/relations.py
```diff
@@ -236,3 +163,72 @@ def make_unbound_dt(self):
...
+        seen_expr[ex] = keep
+        return pa.RecordBatchReader.from_batches(schema, result)
+
+    def replacer(node, _, **kwargs):
+        if isinstance(node, Relation):
+            updated = {}
+            for k, v in list(kwargs.items()):
+                try:
+                    if v in tables:
+                        remote: RemoteTableCounter = tables[v]
```
dlovell, line 195: why do we need `RemoteTableCounter`?

python/letsql/expr/relations.py
mesejo: We need to register one batch per kwarg, each kwarg representing a reference to the RemoteTable. It is similar to searching for occurrences in the SQL string.

python/letsql/expr/relations.py
mesejo: We don't need it, removed it

python/letsql/expr/relations.py
dlovell: I'm saying we should invoke the tee'ing only once: we should not tee a tee.

mesejo: Done

dlovell (approved):

PR #404, opened 2024-12-12
feat: add caching for to_sql

mesejo:


dlovell:
The failure is because `letsql.common.utils.defer_utils.visit_Read` dynamically generates its name and how we're caching precludes that.

dlovell (approved):

PR #412, opened 2024-12-17
chore: execute tests in parallel

mesejo:


dlovell:
@mesejo thanks for this! i see you made some changes to test fixtures, can you give a quick overview or point me to something to read about what constraints / requirements using `pytest-xdist` imposes on fixtures?

mesejo:
So, essentially, `pytest-xdist` launches multiple workers, and each worker executes a set of parallel tests. If a test modifies a shared resource like the database behind the fixture, this causes the assumptions about the state of the database to no longer be met; for example, we could have more tables than expected.

One strategy to cope with such a problem is to make the fixtures parallel-friendly, i.e., they can be used and modified by multiple tests.

Another one is to change the sets of tests each worker executes and assign all the tests that share a fixture to the same worker. For more on that strategy, you can read:

- https://pytest-xdist.readthedocs.io/en/stable/distribution.html
- https://jcleow.github.io/2024/02/12/Understanding-pytest-xdist.html#e2-customising-the-scoping-capabilities-of-pytest-xdist 

This PR may not provide a significant speed-up over running the tests in a single thread.

dlovell (approved):

dlovell:
> One strategy to cope with such a problem is to make the fixtures parallel-friendly, i.e., they can be used and modified by multiple tests.
> 
> Another one is to change the sets of tests each worker executes and assign all the tests that share a fixture to the same worker. For more on that strategy, you can read:
> 
>     * https://pytest-xdist.readthedocs.io/en/stable/distribution.html
> 
>     * https://jcleow.github.io/2024/02/12/Understanding-pytest-xdist.html#e2-customising-the-scoping-capabilities-of-pytest-xdist

@mesejo thanks for the pointers!

PR #431, opened 2025-01-07
 Feature implement train test split

ghoersti:
This PR adds functionality for a train test split to letsql. 

 I have not updated the crate to bump semantic version.

```python
  import letsql as ls 
  table = memtable([(i, "val") for i in range(N)], columns=["key1", "val"])
  train_table, test_table = ls.train_test_split(
      table, unique_key="key1", test_sizes=test_size, random_seed=42
  )
```

There was a name space conflict of some sort so there is some odd imports at the top `api.py` , this will be fixed. 
It needs more tests on errors & maybe a warning about how bin size & its impact on  train test split percentages, but. I wanted to get something out there to receive feedback. 

if `test_size` is a `float` you will get back   train , test sets with a proportion of (1 - test_size) , ( test_size) 
Otherwise `test_sizes` must be an iterable of floats that sum to 1.

```python
# If test sizes is a list of floats , mutually exclusive deferred partitions will be returned
partition_proportions = [0.1, 0.2, 0.3, 0.4]


hold_out, test, validation, training = ls.train_test_splits(
    table,
    unique_key="key1",
    test_sizes=partition_proportions,
    num_buckets=N,
    random_seed=42,
)
```

### DIFF
--- a/python/letsql/expr/api.py
+++ b/python/letsql/expr/api.py
@@ -40,9 +40,7 @@ from letsql.expr.relations import (
     CachedNode,
     register_and_transform_remote_tables,
 )
-
-from letsql.expr.ml import train_tests_splits
-
+from letsql.expr.ml import _train_test_splits
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
@@ -1691,4 +1689,4 @@ def to_parquet(
 
 
 def train_test_splits(*args, **kwargs) -> Iterator[tuple[ir.Table, ir.Table]]:
-    return train_tests_splits(*args, **kwargs)
+    return _train_test_splits(*args, **kwargs)
--- a/python/letsql/expr/api.py
+++ b/python/letsql/expr/api.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import datetime
 import functools
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Union, overload, Mapping
+from typing import TYPE_CHECKING, Any, Union, overload, Iterable, Iterator, Mapping
 
 import ibis
 import ibis.expr.builders as bl
@@ -40,6 +40,8 @@ from letsql.expr.relations import (
     CachedNode,
     register_and_transform_remote_tables,
 )
+
+from letsql.expr.ml import train_tests_splits
 
 
 if TYPE_CHECKING:
@@ -108,6 +110,7 @@ __all__ = (
     "table",
     "time",
     "today",
+    "train_test_splits",
     "to_parquet",
     "to_pyarrow",
     "to_pyarrow_batches",
@@ -1685,3 +1688,7 @@ def to_parquet(
         with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
             for batch in batch_reader:
                 writer.write_batch(batch)
+
+
+def train_test_splits(*args, **kwargs) -> Iterator[tuple[ir.Table, ir.Table]]:
+    return train_tests_splits(*args, **kwargs)
--- a/python/letsql/expr/api.py
+++ b/python/letsql/expr/api.py
@@ -40,7 +40,7 @@ from letsql.expr.relations import (
     CachedNode,
     register_and_transform_remote_tables,
 )
-from letsql.expr.ml import _train_test_splits
+from letsql.expr import ml
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
@@ -1689,4 +1689,4 @@ def to_parquet(
 
 
 def train_test_splits(*args, **kwargs) -> Iterator[tuple[ir.Table, ir.Table]]:
-    return _train_test_splits(*args, **kwargs)
+    return ml.train_test_splits(*args, **kwargs)
--- a/python/letsql/expr/api.py
+++ b/python/letsql/expr/api.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import datetime
 import functools
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Union, overload, Iterable, Iterator, Mapping
+from typing import TYPE_CHECKING, Any, Union, overload, Iterable, Mapping
 
 import ibis
 import ibis.expr.builders as bl
@@ -40,7 +40,7 @@ from letsql.expr.relations import (
     CachedNode,
     register_and_transform_remote_tables,
 )
-from letsql.expr import ml
+from letsql.expr.ml import train_test_splits
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
@@ -1686,7 +1686,3 @@ def to_parquet(
         with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
             for batch in batch_reader:
                 writer.write_batch(batch)
-
-
-def train_test_splits(*args, **kwargs) -> Iterator[tuple[ir.Table, ir.Table]]:
-    return ml.train_test_splits(*args, **kwargs)
--- a/python/letsql/expr/ml.py
+++ b/python/letsql/expr/ml.py
@@ -1,0 +1,143 @@
+from random import Random
+from typing import Tuple, Iterable, List, Iterator
+
+# TODO: How should we / should we enforce letsql table ?
+import ibis.expr.types as ir
+from ibis import literal
+
+
+def _calculate_bounds(test_sizes: List[float]) -> List[Tuple[float, float]]:
+    """
+    Calculates the cumulative sum of test_sizes and generates bounds for splitting data.
+
+    Parameters
+    ----------
+    test_sizes : List[float]
+        A list of floats representing the desired proportions for data splits.
+        Each value should be between 0 and 1, and their sum should ideally be less than or equal to 1.
+
+    Returns
+    -------
+    List[Tuple[float, float]]
+        A list of tuples, where each tuple contains two floats representing the
+        lower and upper bounds for a split. These bounds are calculated based on
+        the cumulative sum of the `test_sizes`.
+    """
+
+    num_splits = len(test_sizes)
+    cumulative_sizes = [sum(test_sizes[: i + 1]) for i in range(num_splits)]
+    cumulative_sizes.insert(0, 0.0)
+    bounds = [(cumulative_sizes[i], cumulative_sizes[i + 1]) for i in range(num_splits)]
+    return bounds
+
+
+def train_test_splits(
+    table: ir.Table,
+    unique_key: str | list[str],
+    test_sizes: Iterable[float] | float,
+    num_buckets: int = 10000,
+    random_seed: int | None = None,
+) -> Iterator[ir.Table]:
+    """Generates multiple train/test splits of an Ibis table for different test sizes.
+
+    This function splits an Ibis table into multiple subsets based on a unique key
+    or combination of keys and a list of test sizes. It uses a hashing function to
+    convert the unique key into an integer, then applies a modulo operation to split
+    the data into buckets. Each subset of data is defined by a range of
+    buckets determined by the cumulative sum of the test sizes.
+
+    Parameters
+    ----------
+    table : ir.Table
+        The input Ibis table to be split.
+    unique_key : str | list[str]
+        The column name(s) that uniquely identify each row in the table. This
+        unique_key is used to create a deterministic split of the dataset
+        through a hashing process.
+    test_sizes : Iterable[float] | float
+        An iterable of floats representing the desired proportions for data splits.
+        Each value should be between 0 and 1, and their sum must equal 1. The
+        order of test sizes determines the order of the generated subsets. If float is passed
+        it assumes that the value is for the test size and that a tradition tain test split of (1-test_size, test_size) is returned.
+    num_buckets : int, optional
+        The number of buckets into which the data can be binned after being
+        hashed (default is 10000). It controls how finely the data is divided
+        during the split process. Adjusting num_buckets can affect the
+        granularity and efficiency of the splitting operation, balancing
+        between accuracy and computational efficiency.
+    random_seed : int | None, optional
+        Seed for the random number generator. If provided, ensures
+        reproducibility of the split (default is None).
+
+    Returns
+    -------
+    Iterator[ir.Table]
+        An iterator yielding Ibis table expressions, each representing a mutually exclusive
+        subset of the original table based on the specified test sizes.
+
+    Raises
+    ------
+    ValueError
+        If any value in `test_sizes` is not between 0 and 1.
+        If `test_sizes` does not sum to 1.
+        If `num_buckets` is not an integer greater than 1.
+
+    Examples
+    --------
+    >>> import letsql as ls
+    >>> import ibis
+    >>> table = ibis.memtable({"key": range(100), "value": range(100,200)})
+    >>> unique_key = "key"
+    >>> test_sizes = [0.2, 0.3, 0.5]
+    >>> splits = ls.train_test_splits(table, unique_key, test_sizes, num_buckets=10, random_seed=42)
+    >>> for i, split_table in enumerate(splits):
+    ...     print(f"Split {i+1} size: {split_table.count().execute()}")
+    ...     print(split_table.execute())
+    Split 1 size: 20
+    Split 2 size: 30
+    Split 3 size: 50
+    """
+    # Convert to traditional train test split
+    if isinstance(test_sizes, float):
+        test_sizes = [1 - test_sizes, test_sizes]
+
+    if not all(isinstance(test_size, float) for test_size in test_sizes):
+        raise ValueError("Test size must be float.")
+
+    if not all((0 < test_size < 1) for test_size in test_sizes):
+        raise ValueError("test size should be a float between 0 and 1.")
+
+    if not (isinstance(num_buckets, int)):
+        raise ValueError("num_buckets must be an integer.")
+
+    if not (num_buckets > 1 and isinstance(num_buckets, int)):
+        raise ValueError(
+            "num_buckets = 1 places all data into training set. For any integer x  >=0 , x mod 1 = 0 . "
+        )
+
+    if not sum(test_sizes) == 1:
+        raise ValueError("Test sizes must sum to 1")
+
+    # Get cumulative bounds
+    bounds = _calculate_bounds(test_sizes=test_sizes)
+
+    # Set the random seed if set, & Generate a random 256-bit key
+    random_str = str(Random(random_seed).getrandbits(256))
+
+    if isinstance(unique_key, str):
+        unique_key = [unique_key]
+
+    comb_key = literal(",").join(table[col].cast("str") for col in unique_key)
+
+    table = table.mutate(**{"hash": (comb_key + random_str).hash().abs() % num_buckets})
+    train_test_filters = [
+        (
+            literal(bound[0]).cast("decimal(38, 9)") * num_buckets <= table.hash
+        )  # lower bound condition
+        & (
+            table.hash < literal(bound[1]).cast("decimal(38, 9)") * num_buckets
+        )  # upper bound condition
+        for i, bound in enumerate(bounds)
+    ]
+
+    return (table.filter(_filter).drop(["hash"]) for _filter in train_test_filters)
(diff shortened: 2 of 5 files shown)


ghoersti:
In order to follow the the `CONTRIBUTING.MD` I needed to add these packages to the python env. 

`duckdb pins gcsfs datafusion`  along with having the appraise env vars populated for Postgres.

mesejo:
> In order to follow the the `CONTRIBUTING.MD` I needed to add these packages to the python env.
> 
> `duckdb pins gcsfs datafusion` along with having the appraise env vars populated for Postgres.

I'll will update the `CONTRIBUTING.MD`, thanks for pointing it out.

python/letsql/expr/api.py
```diff
@@ -1655,3 +1659,96 @@ def to_parquet(expr: ir.Expr, path: str | Path, **kwargs: Any):
...
+    random_str = str(grb(256))
+
+    if isinstance(unique_key, str):
+        unique_key = [unique_key]
+
+    # Append random string to the name to avoid collision
+    train_flag = f"train_{random_str}"
+    comb_key = ibis.literal(",").join(table[col].cast("str") for col in unique_key)
+
+    table = table.mutate(
```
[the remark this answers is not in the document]
ghoersti: That just makes a combined key from either the string or the list provided. 
The whole thing is stochastic. 

The way that it tries to get the % is this condition
```python
(comb_key + random_str).hash().abs() % num_buckets
            < int((1 - test_size) * num_buckets)
```

The whole thing is stochastic. 
What happens is that `(comb_key + random_str).hash().abs() ` becomes a positive int ( there are some collision risks here)  and its relying on a uniform hash distribution , for small numbers it isn't.  So the default buckets is `100` 

So it is basically just saying hey turn the hash into an integer between 0-99  if the hash happens to be be 
`hash <  .75 * 99 `  its in the train set.

python/letsql/expr/api.py
```diff
@@ -1655,3 +1659,96 @@ def to_parquet(expr: ir.Expr, path: str | Path, **kwargs: Any):
...
+    """
+    if not (0 < test_size < 1):
+        raise ValueError("test size should be a float between 0 and 1.")
+
+    # Set the random seed for reproducibility
+    if random_seed:
+        rndm_seed(random_seed)
+
+    # Generate a random 256-bit key
+    random_str = str(grb(256))
```
dlovell, line 1735: I think this is better expressed as
`str(random.Random(random_seed).getrandbits(256))`
so that it cannot be impacted by anything outside

python/letsql/expr/api.py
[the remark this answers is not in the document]
ghoersti: @mesejo  I updated comment for clarity

python/letsql/expr/api.py
ghoersti: I think we should sure that the seed can still use system entropy? 


```python
  # Set the random seed, if needed  for reproducibility
  if random_seed:
      random_str = str(random.Random(random_seed).getrandbits(256))
  else:
      #use system entropy to get seed os.urandom() 
      random_str = str(random.Random().getrandbits(256))

```
Users may want to rely on randomness generated  and not rely on providing a seed because the seed provided may not provide enough entropy to be random. 

I think this  option should be there , vs something where we pick a default seed. 


My preference is that we

ghoersti: also it still has a scoping  issue with 
https://github.com/letsql/letsql/blob/497e33364dcf81b645962cc1c52c20f1c1b78dbb/python/letsql/expr/api.py#L460

ghoersti: we can do this though `__import__("random").Random().getrandbits(1)`

ghoersti: or use `importlib`.

dlovell: when `random_seed` is `None`, `random.Random()` and `random.Random(random_seed)` are the same
how about `from random import Random`?

python/letsql/expr/api.py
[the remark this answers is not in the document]
dlovell: If we are going to use buckets, I think we should use a larger number. Perhaps at least 10,000.

dlovell:
@ghoersti How would you feel about extending this so that the user can provide multiple weights to get multiple splits?

python/letsql/expr/api.py
dlovell: on second thought, i think we should avoid putting implementation into `api.py`. how about we put this somewhere else and expose it through `api.py`?

python/letsql/expr/api.py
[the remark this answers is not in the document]
ghoersti: With the default being 10k the data will need to large other wise you run the risk of everything falling in the train set set.  If your data set is less than the num buckets we should probably have a warning.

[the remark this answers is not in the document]
ghoersti: if  default is 10k for `num_buckets` , if someone has a 2k rows. Every thing could easily end up in train set

python/letsql/expr/api.py
ghoersti: >when random_seed is None, random.Random() and random.Random(random_seed)
Ahh yeah true .

I can move it somehwere else.

ghoersti:
> @ghoersti How would you feel about extending this so that the user can provide multiple weights to get multiple splits?

I can do that , but also they could just use a list comprehension and get `[ tuple[ir.Table, ir.Table], tuple[ir.Table, ir.Table]....]`

ghoersti:
I implemented a functions `train_test_split` & `train_test_splits` , I also think it should be clear if you change your bucket size your seed no longer reproduces the same sets.

dlovell:
> I implemented a functions `train_test_split` & `train_test_splits` and added tests. and removed the function from the api and placed it in `ml.py` I can add tests in `api,py` , I did not add tests for the stubs in `api.py` but I did test the implementations. @dlovell @mesejo

I think that `train_test_split` should be a special case of `train_test_splits` and that `train_test_splits` should take an iterable of `float` that should sum to `<=1` so that the exprs that are returned are mutually exclusive but collectively exhaustive of the rows of the input expr

ghoersti:
> > I implemented a functions `train_test_split` & `train_test_splits` and added tests. and removed the function from the api and placed it in `ml.py` I can add tests in `api,py` , I did not add tests for the stubs in `api.py` but I did test the implementations. @dlovell @mesejo
> 
> I think that `train_test_split` should be a special case of `train_test_splits` and that `train_test_splits` should take an iterable of `float` that should sum to `<=1` so that the exprs that are returned are mutually exclusive but collectively exhaustive of the rows of the input expr

Ok this makes sense, I will implement that, should the api still have both exposed ? or just train test splits ?

ghoersti:
@dlovell  would we want to change the name to `train_test_partitions` ? Since now this is just a way to partition deferred?

python/letsql/expr/ml.py
```diff
@@ -0,0 +1,148 @@
...
+    -------
+    List[Tuple[float, float]]
+        A list of tuples, where each tuple contains two floats representing the
+        lower and upper bounds for a split. These bounds are calculated based on
+        the cumulative sum of the `test_sizes`.
+    """
+
+    if not all(0 < size < 1 for size in test_sizes):
+        raise ValueError("All test sizes must be between 0 and 1.")
+    if not sum(test_sizes) == 1:
```
ghoersti, line 29: This does not need a test it is covered in the `train_test_splits`

python/letsql/expr/ml.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if isinstance(unique_key, str):
+        unique_key = [unique_key]
+
+    comb_key = literal(",").join(table[col].cast("str") for col in unique_key)
+
+    table = table.mutate(**{"hash": (comb_key + random_str).hash().abs() % num_buckets})
+    train_test_filters = [
+        (
+            literal(bound[0]).cast("decimal(38, 9)") * num_buckets <= table.hash
```
ghoersti, line 135: `cast("decimal(38, 9)") ` because default when it executed on duck_db  was DECIMAL(18) and a test was breaking due to overflow.

dlovell:
@ghoersti can you implement `visit_Hash` so we can execute via `ls.execute`?

```
File /nix/store/wv660b4m8dydvygpa7cwkrblk041xc7k-letsql-dev-env/lib/python3.10/site-packages/ibis/backends/sql/compilers/base.py:1514, in SQLGlotCompiler.visit_Undefined(self, op, **_)
   1513 def visit_Undefined(self, op, **_):
-> 1514     raise com.OperationNotDefinedError(
   1515         f"Compilation rule for {type(op).__name__!r} operation is not defined"
   1516     )

OperationNotDefinedError: Compilation rule for 'Hash' operation is not defined
```

ghoersti:
> @ghoersti can you implement `visit_Hash` so we can execute via `ls.execute`?
> 
> ```
> File /nix/store/wv660b4m8dydvygpa7cwkrblk041xc7k-letsql-dev-env/lib/python3.10/site-packages/ibis/backends/sql/compilers/base.py:1514, in SQLGlotCompiler.visit_Undefined(self, op, **_)
>    1513 def visit_Undefined(self, op, **_):
> -> 1514     raise com.OperationNotDefinedError(
>    1515         f"Compilation rule for {type(op).__name__!r} operation is not defined"
>    1516     )
> 
> OperationNotDefinedError: Compilation rule for 'Hash' operation is not defined
> ```


> @ghoersti can you implement `visit_Hash` so we can execute via `ls.execute`?
> 
> ```
> File /nix/store/wv660b4m8dydvygpa7cwkrblk041xc7k-letsql-dev-env/lib/python3.10/site-packages/ibis/backends/sql/compilers/base.py:1514, in SQLGlotCompiler.visit_Undefined(self, op, **_)
>    1513 def visit_Undefined(self, op, **_):
> -> 1514     raise com.OperationNotDefinedError(
>    1515         f"Compilation rule for {type(op).__name__!r} operation is not defined"
>    1516     )
> 
> OperationNotDefinedError: Compilation rule for 'Hash' operation is not defined
> ```

@dan oh ok  would go in `compiler.py`  for each of our backends that do not nativley support Hash ?

dlovell:
> @dan oh ok would go in `compiler.py` for each of our backends that do not nativley support Hash ? Probably something using pyarrow.compute.sha256 to hash.

I don't think we have the choice to use `pyarrow.compute.sha256` (though somehow it might still end up being the equivalent). For our `letsql` backend and `datafusion`, I think it will be something that translates to `digest(..., 'sha256')`.

ghoersti:
> > @dan oh ok would go in `compiler.py` for each of our backends that do not nativley support Hash ? Probably something using pyarrow.compute.sha256 to hash.
> 
> I don't think we have the choice to use `pyarrow.compute.sha256` (though somehow it might still end up being the equivalent). For our `letsql` backend and `datafusion`, I think it will be something that translates to `digest(..., 'sha256')`.

I see it looks like `SQLGlotCompiler` already has one: 

```python
>>> [ c for c in dir(SQLGlotCompiler) if 'Hash' in c ] 
['visit_Hash', 'visit_HashBytes']
```

dlovell:
@ghoersti can we refactor the example so we don't recalculate counts?

```
name_to_proportion = {                                                                
    "hold_out": 0.1,                                                                  
    "test": 0.2,                                                                      
    "validation": 0.3,        
    "training": 0.4,                                                                                                                                                        
}                                          
partitions = tuple(ls.train_test_splits(   
    table,                                                                            
    unique_key="key1",                                                                
    test_sizes=tuple(name_to_proportion.values()),                                    
    num_buckets=N,                         
    random_seed=42,                                                                   
))                                                                                                                                                                          
name_to_count = {                                                                                                                                                           
    name: partition.count().execute()
    for name, partition in zip(name_to_proportion, partitions)
}                                          
total = sum(name_to_empirical_proportion.values())
for name, count in name_to_count.items():
    print(f"{name} ratio: {count/total}")
```

dlovell:
> > > @dan oh ok would go in `compiler.py` for each of our backends that do not nativley support Hash ? Probably something using pyarrow.compute.sha256 to hash.
> > 
> > 
> > I don't think we have the choice to use `pyarrow.compute.sha256` (though somehow it might still end up being the equivalent). For our `letsql` backend and `datafusion`, I think it will be something that translates to `digest(..., 'sha256')`.
> 
> I see it looks like `SQLGlotCompiler` already has one:
> 
> ```python
> >>> [ c for c in dir(SQLGlotCompiler) if 'Hash' in c ] 
> ['visit_Hash', 'visit_HashBytes']
> ```

The `duckdb` backend works, but the `datafusion` backend and our backend don't. If we don't implement `visit_Hash` for them here, can you open an issue for them?

mesejo:
> > @dan oh ok would go in `compiler.py` for each of our backends that do not nativley support Hash ? Probably something using pyarrow.compute.sha256 to hash.
> 
> I don't think we have the choice to use `pyarrow.compute.sha256` (though somehow it might still be the equivalent). For our `letsql` backend and `datafusion`, I think it will be something that translates to `digest(..., 'sha256')`.

There is a `visit_HexDigest` in `compiler.py` that can be used as a guide:

```python
    def visit_HexDigest(self, op, *, arg, how):
        return self.f.encode(self.f.digest(arg, how), "hex")
```

So, a possible implementation for hash could be:

```python
    def visit_Hash(self, op, *, arg):
        return self.f.digest(arg, 'sha256')
```

But that returns a binary string, and it is not clear to me that it can be converted into an integer in a performant manner

ghoersti:
> name_to_proportion = {                                                                
>     "hold_out": 0.1,                                                                  
>     "test": 0.2,                                                                      
>     "validation": 0.3,        
>     "training": 0.4,                                                                                                                                                        
> }

```python
In [17]: # If test sizes is a list of floats , mutually exclusive partitions will be returned
    ...: partion_names = {
    ...:     "hold_out": 0.1,
    ...:     "test": 0.2,
    ...:     "validation": 0.3,
    ...:     "training": 0.4,
    ...: 
    ...: }
    ...: 
    ...: partitions = tuple(ls.train_test_splits(
    ...:         table,
    ...:         unique_key="key1",
    ...:         test_sizes=list(partion_info.values()),
    ...:         num_buckets=N,
    ...:         random_seed=42,
    ...:     )
    ...: )
    ...: counts = [p.count().execute() for p in  partitions ]
    ...: total = sum(counts)
    ...: 
    ...: for i, partition_name in enumerate(partion_info.keys()):
    ...:     print(f"{partition_name.upper()} Ratio: {round(counts[i]/ total,2)}")
    ...: 
HOLD_OUT Ratio: 0.1
TEST Ratio: 0.2
VALIDATION Ratio: 0.3
TRAINING Ratio: 0.4
```

Done

python/letsql/expr/api.py
```diff
@@ -1709,3 +1712,7 @@ def to_parquet(
         with pq.ParquetWriter(path, batch_reader.schema, **kwargs) as writer:
             for batch in batch_reader:
                 writer.write_batch(batch)
+
+
+def train_test_splits(*args, **kwargs) -> Iterator[tuple[ir.Table, ir.Table]]:
+    return ml.train_test_splits(*args, **kwargs)
```
dlovell, line 1718: when we do this, we lose the signature
can we use
`from letsql.expr.ml import train_tests_splits`
instead?

dlovell:
All the tests use `memtable`, can we parametrize by `con` and test `duckdb`, `datafusion`, `letsql` and at least one backend we expect to `xfail`?

ghoersti:
> All the tests use `memtable`, can we parametrize by `con` and test `duckdb`, `datafusion`, `letsql` and at least one backend we expect to `xfail`?

I can do this.

dlovell:
@ghoersti in the other tests, we use connections themselves as test fixtures. when done properly, that makes sure we have "clean" connections so that, for instance, if the postgres test fails in the middle, we aren't running other tests from an unknown state.
i don't know that we need to do that here, and iirc, its a pain to parametrize over the cons, so consider this approach instead
```
diff --git a/python/letsql/tests/test_ml.py b/python/letsql/tests/test_ml.py
index e6e420f..c5ca057 100644
--- a/python/letsql/tests/test_ml.py
+++ b/python/letsql/tests/test_ml.py
@@ -183,15 +183,11 @@ def test_train_test_splits_num_buckets_gt_one():
 
 
 @pytest.mark.parametrize(
-    "connection_name,expected_conn_type",
-    [
-        ("letsql", ls.backends.let.Backend),
-        ("duckdb", ls.backends.duckdb.Backend),
-        ("postgres", ls.backends.postgres.Backend),
-    ],
+    "connect_method",
+    (ls.connect, ls.duckdb.connect, ls.postgres.connect_env),
 )
 def test_train_test_splits_intersections_parameterized_pass(
-    create_passing_connections, connection_name, expected_conn_type
+    connect_method,
 ):
     # This is testing the base case where a single float becomes ( 1-test_size , test_size ) proportion
     # Check counts and overlaps in train and test dataset
@@ -200,9 +196,9 @@ def test_train_test_splits_intersections_parameterized_pass(
 
     # create test table for backend
     test_df = pd.DataFrame([(i, "val") for i in range(N)], columns=["key1", "val"])
-    con = create_passing_connections[connection_name]
-    test_table_name = f"{connection_name}_test_df"
-    con.create_table(test_table_name, test_df)
+    con = connect_method()
+    test_table_name = f"{con.name}_test_df"
+    con.create_table(test_table_name, test_df, temp=con.name=="postgres")
 
     table = con.table(test_table_name)
 
@@ -246,19 +242,16 @@ def test_train_test_splits_intersections_parameterized_pass(
         == 0
     )
 
-    assert isinstance(con, expected_conn_type)
```

dlovell:
@ghoersti this is how to mark an individual test in a parametrized list as xfail
```
diff --git a/python/letsql/tests/test_ml.py b/python/letsql/tests/test_ml.py
index e6e420f..204bdbd 100644
--- a/python/letsql/tests/test_ml.py
+++ b/python/letsql/tests/test_ml.py
@@ -183,15 +183,19 @@ def test_train_test_splits_num_buckets_gt_one():
 
 
 @pytest.mark.parametrize(
-    "connection_name,expected_conn_type",
-    [
-        ("letsql", ls.backends.let.Backend),
-        ("duckdb", ls.backends.duckdb.Backend),
-        ("postgres", ls.backends.postgres.Backend),
-    ],
+    "connect_method",
+    (
+        ls.connect,
+        ls.duckdb.connect,
+        ls.postgres.connect_env,
+        pytest.param(
+            ls.datafusion.connect,
+            marks=pytest.mark.xfail(reason="Compilation rule for 'Hash' operation is not define"),
+        ),
+    ),
 )
```
if we do that, we can remove the other test whose sole purpose is to xfail

python/letsql/expr/ml.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Set the random seed if set, & Generate a random 256-bit key
+    random_str = str(Random(random_seed).getrandbits(256))
+
+    if isinstance(unique_key, str):
+        unique_key = [unique_key]
+
+    comb_key = literal(",").join(table[col].cast("str") for col in unique_key)
+
+    table = table.mutate(**{"hash": (comb_key + random_str).hash().abs() % num_buckets})
```
dlovell, line 132: i think we need a random string component of the hash column name to prevent collisions with user columns

python/letsql/expr/api.py
dlovell: @ghoersti why do we redefine `train_test_splits` in `api.py` rather than importing from `ml.py`?

dlovell: this is the impact of redefining with `*args, **kwargs` rather than importing
```
In [1]: import letsql as ls

In [2]: ls.train_test_splits
Out[2]: <function letsql.expr.api.train_test_splits(*args, **kwargs) -> 'Iterator[tuple[ir.Table, ir.Table]]'>

In [3]: ls.expr.ml._train_test_splits
Out[3]: <function letsql.expr.ml._train_test_splits(table: ibis.expr.types.relations.Table, unique_key: str | list[str], test_sizes: Union[Iterable[float], float], num_buckets: int = 10000, random_seed: int | None = None) -> Iterator[ibis.expr.types.relations.Table]>
```

python/letsql/expr/ml.py
ghoersti: Oh ok this makes sense for sure.

ghoersti:
> ls.postgres.connect_env

why deviate for  `ls.postgres.connect_env` 
?

dlovell:
> > ls.postgres.connect_env
> 
> why deviate for `ls.postgres.connect_env` ?

We don't need any of the existing tables / state, we simply want to create a table and using a temporary table isolates the other tests from our changes here

PR #437, opened 2025-01-08
chore: update contributing workflow

mesejo:
Raised in: https://github.com/letsql/letsql/pull/431#issuecomment-2574264849

The CONTRIBUTING.md file is the step where the environment variables are populated. Additionally, the file `requirements-dev.txt` is missing some dependencies.

Therefore, the solution is to update the:
- CONTRIBUTING.md
- pre-commit hook for exporting development dependencies


ghoersti (commented):
LGTM

PR #444, opened 2025-01-11
fix/_find_backend

### LINKED ISSUE #442: `_find_backend` doesn't work on cached exprs
running
```
import letsql as ls
from letsql.common.caching import ParquetCacheStorage


con = ls.connect()
expr = (
    con.read_parquet(ls.options.pins.get_path("astronauts"))
    .cache(storage=ParquetCacheStorage(source=con))
)
(l, r) = (expr._find_backend().name, con.name)
assert l == r, (l, r)
```
results in 
```
AssertionError                            Traceback (most recent call last)
Cell In[1], line 11
      6 expr = (
      7     con.read_parquet(ls.options.pins.get_path("astronauts"))
      8     .cache(storage=ParquetCacheStorage(source=con))
      9 )
     10 (l, r) = (expr._find_backend().name, con.name)
---> 11 assert l == r, (l, r)

AssertionError: ('duckdb', 'let')
```

dlovell:
fixes #442

@mesejo , i modify `letsql.common.collections._find_backend` as well, though its not clear where its used.

### DIFF
--- a/python/letsql/backends/let/hotfix.py
+++ b/python/letsql/backends/let/hotfix.py
@@ -165,7 +165,7 @@ class LETSQLAccessor:
 def _letsql_find_backend(self, *, use_default=True):
     # FIXME: push this into LETSQLAccessor
     try:
-        if tuple(op.source for op in self.op().find(Read)):
+        if tuple(op.source for op in self.op().find((Read, CachedNode))):
             current_backend = _backend_init()
         else:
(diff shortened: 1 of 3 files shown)


dlovell:
@mesejo i think this isn't the final solution, but fixes the issue for `CachedNode`

mesejo (approved):

Issue #443, opened 2025-01-11
`test` dependencies don't include extra backends


dlovell:
tests use extra backends like `datafusion`, `duckdb` etc, but they are not included in `dependency-groups.test`

i think our setup has been working because all our installation methods (`uv` in `ci-test.yml`, `uv2nix` in `nix/letsql.nix`) install all dependency groups

PR #448, opened 2025-01-14
chore: fix dependency groups

mesejo:


mesejo:
> @mesejo does this close #443?

Yes, it should

dlovell (approved):

PR #451, opened 2025-01-14
feat: add hash(string) function

mesejo:


python/letsql/backends/let/datafusion/compiler.py
```diff
@@ -576,6 +576,9 @@ def visit_GroupConcat(self, op, *, arg, sep, where, order_by):
     def visit_ArrayFlatten(self, op, *, arg):
         return self.if_(arg.is_(NULL), NULL, self.f.flatten(arg))
 
+    def visit_Hash(self, op, *, arg):
+        return self.f.hash_int(arg)
```
ghoersti, line 580: I see the rust implementation of `hash_int` , how does `hash_int` end up becoming accessible by our data fusion compiler?  in `src/context.rs`

ghoersti (approved):

python/letsql/backends/let/datafusion/compiler.py
dlovell: @mesejo i think the expectation of `visit_Hash` is that it returns a string so we should rename this to something else

ghoersti:
Hash in snowflake, ibis returns int why would visit _Hash return string?
https://docs.snowflake.com/en/sql-reference/functions/hash

https://ibis-project.org/reference/expression-generic#ibis.expr.types.generic.Value.hash

src/context.rs
```diff
@@ -240,6 +241,7 @@ impl PySessionContext {
         ctx.register_udf(predict_xgb.clone());
         ctx.register_udf(ScalarUDF::from(GreatestFunc::new()));
         ctx.register_udf(ScalarUDF::from(LeastFunc::new()));
+        ctx.register_udf(ScalarUDF::from(HashIntFunc::new()));
```
mesejo, line 244: @ghoersti In here, we add the function to set of functions available in DataFusion, this is actually how all the functions are created in DataFusion via the ScalarUDF

dlovell:
> Hash in snowflake, ibis returns int why would visit _Hash return string? https://docs.snowflake.com/en/sql-reference/functions/hash https://ibis-project.org/reference/expression-generic#ibis.expr.types.generic.Value.hash

Ah, you're right.

python/letsql/backends/let/datafusion/compiler.py
dlovell: @mesejo , nvm, per @ghoersti  all the other `visit_Hash`s ***do*** return int

PR #479, opened 2025-01-28
feat: add ordering arg to register_record_batch_reader

### LINKED ISSUE #470: enable passing ordering to `PyRecordBatchProvider`
so that it can specify its ordering in `create_physical_plan`

mesejo:
This enables parent plans in DataFusion to take advantage of the
sortedness of the RecordBatchReader.

closes #470

### DIFF
--- a/python/letsql/backends/let/tests/test_client.py
+++ b/python/letsql/backends/let/tests/test_client.py
@@ -103,4 +103,8 @@ def test_register_record_batch_reader_sorted(batting_df, keys):
     expr = t.group_by("yearID").agg(max_tiny=t["stint"].max())
 
     physical_plan = ls.get_plans(expr)["physical_plan"]
-    assert not keys or "ordering_mode=Sorted" in physical_plan
+    ordering_string = "ordering_mode=Sorted"
+    if keys:
+        assert ordering_string in physical_plan
+    else:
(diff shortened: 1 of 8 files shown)


python/letsql/backends/let/tests/test_client.py
```diff
@@ -75,3 +77,30 @@ def test_register_table_with_uppercase_multiple_times(ls_con):
...
+    record_batch_reader = t.select("yearID", "stint").to_pyarrow_batches()
+    t = con.register(
+        record_batch_reader,
+        "t2",
+        ordering=[key.resolve(t) for key in keys],
+    )
+    expr = t.group_by("yearID").agg(max_tiny=t["stint"].max())
+
+    physical_plan = ls.get_plans(expr)["physical_plan"]
+    assert not keys or "ordering_mode=Sorted" in physical_plan
```
dlovell, line 106: I think we should make it explicit that we expect `"ordering_mode=Sorted"` to **not** be in `physical_plan` when there is no ordering
```
    ordering_string = "ordering_mode=Sorted"
    if keys:
        assert ordering_string in physical_plan
    else:
        assert ordering_string not in physical_plan
```

dlovell (approved):

PR #482, opened 2025-01-28
fix/uv2nix changes

dlovell:
update for changes to uv for dynamic versioning and concomitant changes to pytest


hussainsultan (approved):
LGTM

PR #499, opened 2025-02-03
chore(examples): add quickgrove example

hussainsultan:


examples/quickgrove_udf.py
```diff
@@ -0,0 +1,35 @@
+from pathlib import Path
+
+import pandas as pd
+
+import letsql as ls
+from letsql.expr.ml import make_quickgrove_udf, rewrite_quickgrove_expr
+from letsql.expr.relations import into_backend
+
+
```
ghoersti, line 9: Just a comment but probably in the future fix the model path name in pins.

ghoersti (approved):
:godmode:

PR #511, opened 2025-02-05
chore: wrap pins with examples

mesejo:
Wrapping provides a better UX, due to the following reasons:
- less code, hide repetitive code
- the user does not need to know the type (csv, parquet) of the file
- the user can know which examples are available via tab completion on the examples package

### DIFF
--- a/examples/remote_caching.py
+++ b/examples/remote_caching.py
@@ -10,7 +10,7 @@ pg = ls.postgres.connect_env()
 name = "batting"
 
 right = (
-    ls.examples.batting.fetch(backend=ddb)
+    ls.examples.batting.fetch(backend=ddb, table_name=name)
     .filter(_.yearID == 2014)
     .pipe(con.register, table_name=f"ddb-{name}")
 )
--- a/examples/remote_caching.py
+++ b/examples/remote_caching.py
@@ -10,7 +10,7 @@ pg = ls.postgres.connect_env()
 name = "batting"
 
 right = (
-    ls.examples.batting.fetch(backend=ddb, table_name=name)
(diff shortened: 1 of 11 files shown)


examples/remote_caching.py
```diff
@@ -8,10 +8,9 @@
 pg = ls.postgres.connect_env()
 
 name = "batting"
-path = ls.config.options.pins.get_path(name)
 
 right = (
-    ddb.read_parquet(path, table_name=name)
+    ls.examples.batting.fetch(backend=ddb)
```
dlovell, line 13: can we use `get_table_from_parquet` in situations like this?
```
ls.examples.core.get_table_from_parquet(name, backend=ddb, table_name=name)
```

python/letsql/examples/core.py
```diff
@@ -0,0 +1,36 @@
...
+    board = ls.options.pins.get_board()
+    dct = {
+        name: pathlib.Path(board.pin_meta(name).file).suffix
+        for name in board.pin_list()
+        if name in whitelist
+    }
+    return dct
+
+
+def get_table_from_parquet(name, backend, table_name):
```
dlovell, line 29: can we have `table_name` default to `None` and use the value of `name` if not given?

examples/remote_caching.py
mesejo: Yes, we could use it, but the user can also do:
```python
ls.examples.batting.fetch(backend=ddb, table_name=name)
```

dlovell:
@mesejo i think we can simplify this a little bit more
https://github.com/letsql/letsql/commit/cf742892caf7ff9188b79a6e228d16dca47ee49c

dlovell (approved):

PR #516, opened 2025-02-07
feat/train-test-partition-labels

dlovell:


python/letsql/expr/ml.py
```diff
@@ -33,34 +37,201 @@
...
+            literal(lower_bound).cast("decimal(38, 9)") * num_buckets,
+            # upper bound is inclusive
+            literal(upper_bound).cast("decimal(38, 9)") * num_buckets - 1,
+        )
+        for (lower_bound, upper_bound) in bounds
+    )
+    return conditions
+
+
+def calc_split_column(
```
ghoersti, line 163: Why have this as well ? When someone could just enumerate over the conditions ?

ghoersti: `calc_split_column`

ghoersti (approved):
I think this all makes sense to me and I see the utility in exposing `calc_split_column` I was just curious why not add a test.

PR #525, opened 2025-02-08
feat: add ibis-yaml compiler

hussainsultan:
- passes tpch-01-22
    - unbound expr only
    - support builtin ScalarUDF

TODO:
- [ ] add into_backend/RemoteTable 
- [ ] add cache 
- [ ] support PyArrow UDF


python/xorq/ibis_yaml/compiler.py
```diff
@@ -0,0 +1,265 @@
...
+            updated_read_info["sql_file"] = sql_filename
+            updated_read_info.pop("sql")
+            updated_reads["reads"][read_name] = updated_read_info
+
+        return updated_reads
+
+    def compile_expr(self, expr: ir.Expr) -> str:
+        expr_hash = self.artifact_store.get_expr_hash(expr)
+
+        backends = find_all_sources(expr)
```
ghoersti, line 211: @hussainsultan  This would break with new profiles implementation

python/xorq/ibis_yaml/udf.py
```diff
@@ -0,0 +1,97 @@
...
+    arg_names = [
+        name
+        for name in dir(op)
+        if not name.startswith("__") and name not in op.__class__.__slots__
+    ]
+
+    return freeze(
+        {
+            "op": "ScalarUDF",
+            "unique_name": op.__func_name__,
```
dlovell, line 45: is this name supposed to be unique in the sense of a hash?

hussainsultan: No, I don't think the name needs to be unique like a hash. I will update for clarity.

python/xorq/ibis_yaml/compiler.py
```diff
@@ -0,0 +1,265 @@
...
+from xorq.vendor.ibis.backends import Profile
+from xorq.vendor.ibis.common.collections import FrozenOrderedDict
+
+
+class CleanDictYAMLDumper(yaml.SafeDumper):
+    def represent_frozenordereddict(self, data):
+        return self.represent_dict(dict(data))
+
+    def ignore_aliases(self, data):
+        return True
```
dlovell, line 30: what is the purpose of `ignore_aliases`?

hussainsultan: YAML by default will create anchors/aliases for objects that repeat. I thought it was less readable so I disabled it. https://stackoverflow.com/questions/58091449/disabling-alias-for-yaml-file-in-python

dlovell (approved):

PR #532, opened 2025-02-13
expr scalar udf

dlovell:


python/letsql/backends/let/hotfix.py
```diff
@@ -212,3 +213,9 @@ def letsql_cache(self, storage=None):
 @property
 def ls(self):
     return LETSQLAccessor(self)
+
+
+# non method hotfix
+ibis.formats.pyarrow._from_pyarrow_types[pa.binary_view()] = (
```
ghoersti, line 219: Is this overriding the binary view type ?

python/letsql/backends/let/datafusion/__init__.py
```diff
@@ -287,7 +289,10 @@ def _register_builtin_udfs(self):
     def _register_udfs(self, expr: ir.Expr) -> None:
         for udf_node in expr.op().find(ops.ScalarUDF):
             if udf_node.__input_type__ == InputType.PYARROW:
-                udf = self._compile_pyarrow_udf(udf_node)
+                if isinstance(udf_node, ExprScalarUDF):
```
ghoersti, line 292: I see it needed this to be able  handle getting the UDF args from a pandas df.

python/letsql/backends/let/hotfix.py
dlovell: no, its teaching ibis how to convert from the pyarrow type, without this we get an error

mesejo (approved):

PR #538, opened 2025-02-15
feat: register ibis datatypes, raise if not registered

mesejo:


python/letsql/common/utils/dask_normalize_expr.py
```diff
@@ -196,6 +202,11 @@ def normalize_module(module):
     )
 
 
+@dask.base.normalize_token.register(dat.DataType)
+def normalize_ibis_datatype(datatype):
+    return dask.tokenize._normalize_seq_func((datatype.nullable, datatype.name.lower()))
```
dlovell, line 207: `DataType` can be complex, like `Struct`
```In [1]: import ibis; ibis.expr.datatypes.core.Struct({"a": ibis.expr.datatypes.int16}).args
Out[1]: ({'a': Int16(nullable=True)}, True)
```
I think here we should use
```
+@dask.base.normalize_token.register(dt.DataType)
+def normalize_ibis_datatype(obj):
+    return dask.tokenize._normalize_seq_func(obj.args)
```

mesejo: makes sense

dlovell (approved):

PR #565, opened 2025-02-25
add scripts demonstrating deferred fit-transform operations

dlovell:


mesejo (approved):

examples/fit-transform-tfidf.py
```diff
@@ -0,0 +1,113 @@
...
+
+import xorq as xo
+import xorq.expr.udf as udf
+import xorq.vendor.ibis.expr.datatypes as dt
+from xorq.common.caching import ParquetCacheStorage
+from xorq.expr.udf import make_pandas_expr_udf, wrap_model
+
+
+@functools.cache
+def hackernews_stories(n=1000):
```
hussainsultan, line 18: Does it make sense to use parquet or even CSV instead of pickle? I think parquet/CSV would be a lot more relatable for data things.

dlovell: the current solution uses a per-item_id json

dlovell:
refactored this into other PRs

PR #570, opened 2025-02-26
Feat/elide secrets

ghoersti:
Adds the elision of secrets to profiles.

- If env vars exist they will be saved in ${VAR_NAME} format
- Raw values will be converted to existing env vars
- Raw values that are secret will be converted to `***elided***`
- Secret Non existent environment variables will be converted to  `***elided***`
- Env vars that do not exist will fail to load eg ${NOT_A_REAL_ENV_VAR} will fail to load if NOT_A_REAL_ENV_VAR does not exist

- Kwargs tuple now is sorted before hash
- Hash no longer uses idx  
```python
from pathlib import Path
import os
from xorq.vendor.ibis.backends import Profile
# First set up some environment variables
os.environ['SNOWFLAKE_USER'] = 'myuser'
os.environ['SNOWFLAKE_PASSWORD'] = 'mypassword'

# Create the profile
profile = Profile(
    con_name='snowflake',
    kwargs_tuple=(
        ('account', 'gzb05818.us-east-1'),
        ('database', 'SNOWFLAKE_SAMPLE_DATA'),
        ('password', 'mypassword'),  # This will be automatically converted to ${SNOWFLAKE_PASSWORD}
        ('role', 'PUBLIC'),
        ('schema', 'PUBLIC'),
        ('user', 'myuser'),  # This will be automatically converted to ${SNOWFLAKE_USER}
        ('warehouse', 'COMPUTE_WH'),
    )
)

profile_path = profile.save(
    alias='snowflake_sample'
)


# Load the profile
loaded_profile = Profile.load('snowflake_sample')

# Get a connection from the profile
connection = loaded_profile.get_con()
```

### DIFF
--- a/python/xorq/vendor/ibis/backends/__init__.py
+++ b/python/xorq/vendor/ibis/backends/__init__.py
@@ -5,10 +5,7 @@ import collections.abc
 import contextlib
 import functools
 import importlib.metadata
-import itertools
-import json
 import keyword
-import os
 import re
 import sys
 import urllib.parse
@@ -16,27 +13,14 @@ import weakref
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, NamedTuple
 
-import dask
-import toolz
-import yaml
-from attr import (
-    field,
-    frozen,
-)
-from attr.validators import (
-    instance_of,
-    optional,
-)
-from envyaml import EnvYAML
-
 import xorq as xo
 import xorq.common.exceptions as exc
 import xorq.vendor.ibis.config
 import xorq.vendor.ibis.expr.operations as ops
 import xorq.vendor.ibis.expr.types as ir
-from xorq.common.utils.inspect_utils import get_arguments
 from xorq.vendor import ibis
 from xorq.vendor.ibis import util
+from xorq.vendor.ibis.backends.profiles import Profile
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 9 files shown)


python/xorq/vendor/ibis/backends/__init__.py
```diff
@@ -1604,9 +1617,47 @@ def as_dict(self):
...
+            if not v:
+                continue
+            if isinstance(v, str):
+                if v.startswith("${") and v.endswith("}"):
+                    env_name = v[2:-1]
+                    if env_name in os.environ:
+                        new_kwargs[k] = f"${{{env_name}}}"
+                        continue
+
+            for env_name, env_value in os.environ.items():
```
mesejo, line 1643: Are the values of os.environ unique?

python/xorq/vendor/ibis/tests/test_profile.py
```diff
@@ -58,7 +60,7 @@ def test_profiles(monkeypatch, tmp_path):
     assert default_profile_dir == pathlib.Path("~/.config/letsql/profiles").expanduser()
     profiles = Profiles()
     assert profiles.profile_dir == default_profile_dir
-    assert not profiles.list()
+    assert not profiles.list()  # why do this ?
```
dlovell, line 63: do we expect to have any profiles pre-populated?

python/xorq/vendor/ibis/backends/__init__.py
```diff
@@ -1568,9 +1571,19 @@ def validate_con_name(self, attr, value):
...
+
+    @property
+    def kwargs_dict(self):
+        return {k: v for k, v in self.kwargs_tuple}
+
     @property
     def hash_name(self):
-        return dask.base.tokenize(self.as_json())
+        pre_hash = json.loads(self.as_json())
+        pre_hash.pop("idx")
```
dlovell, line 1585: why are you `pop`ing idx?

python/xorq/vendor/ibis/backends/__init__.py
dlovell: even if unique, i don't think we should be inferring which environment variable to use
rather, i think the user should be able to configure `xorq.config.Profiles` to configure which environment variables to substitute in for (backend_name, connection_parameter_name) pairs

python/xorq/vendor/ibis/tests/test_profile.py
ghoersti: No I do not think we want any pre-populated, especially in CICD but if running tests locally I would rather just clean up.

python/xorq/vendor/ibis/backends/__init__.py
ghoersti: Because the IDX forces uniqueness. I originally wanted to not have hashes differ only on IDX, but this may also solves daniels comment about duplicate profiles.

ghoersti: I was thinking the relationships would go:
- Profiles:profile 1:n
- profile:connections 1:m
And that we can instantiate as many connections from a profile as we want.
Because if the connection params are the same thenwhy do we care? Why not let the user manage instances of connections ?

python/xorq/vendor/ibis/backends/__init__.py
ghoersti: My assumption was that the if a value exist `bar` that has an env variable `foo` populated like `foo=bar` .
That when we wrote to disk we write foo = ${foo}. We can remove this behavior or make it a toggle.

@mesejo  This is a good call out the assumption was that values would be unique, but we could clearly have collisions on values withe different keys.

>even if unique, i don't think we should be inferring which environment variable to use
>rather, i think the user should be able to configure xorq.config.Profiles to configure which environment variables >to substitute in for (backend_name, connection_parameter_name) pairs


@dlovell  If we do this we are bound to one set of variables per profile backend name. If you are ok with this the getters setters would be on profiles & would be totally on the user to configure outside of the secrets.

ghoersti: I think that for now it would be appropriate to remove inferring and just allow an `env_vars` dict to be passed to save. That will override keys before saving.  

something like 
```python
env_vars_to_overide = dict(user = ${MY_USER}) 
profile.save(env_vars=env_vars_to_overide)
```

Variables will be elided if necessary, but for now users can choose to save whatever they want it would still do the validation for existence. We just would no longer infer the key from the value.

ghoersti: I propose either `.save` allowing or one set of env vars per back end.

PR #573, opened 2025-02-26
chore: replace pydantic AnyURL with attrs

mesejo:


python/xorq/flight/__init__.py
```diff
@@ -22,18 +23,32 @@
...
-        allowed_schemes=(default_scheme,),
-        default_host=default_host,
-        default_port=default_port,
-    )
+allowed_schemes = ("grpc",)
+default_host = "localhost"
+default_port = 5005
+
+
+@define
```
dlovell, line 31: can we make this `@frozen` so that the fields are immutable?

python/xorq/flight/__init__.py
```diff
@@ -45,16 +60,17 @@ def port_in_use(self):
 
     @classmethod
     def from_defaults(cls, **kwargs):
-        (scheme, *_) = cls._constraints.allowed_schemes
-        _kwargs = toolz.merge(
+        (scheme, *_) = allowed_schemes
```
dlovell, line 63: since all fields have a default, we can do this
```
_kwargs = {
    **{attr.name: attr.default for attr in cls.__attrs_attrs__},
    **kwargs,
}
```

dlovell: on second thought, now that we have defaults, we don't need a `from_defaults` method, we can use the constructor itself
but i'd be happy to leave this as is because we'll likely need another PR to randomly choose a port

dlovell (approved):

PR #581, opened 2025-02-27
feat: add run cli command

mesejo:


python/xorq/cli.py
```diff
@@ -100,13 +130,26 @@ def main():
         "--target-dir", default="build", help="Directory for all generated artifacts"
     )
 
-    # Parse the arguments
+    # Create parser for the "run" command
+    run_parser = subparsers.add_parser(
+        "run", help="Run a build from a builds directory"
+    )
+    run_parser.add_argument("builds_dir", help="Path to the builds directory")
```
dlovell, line 137: this argument should have the same name as the `build` command

i think its probably better to use `builds_dir` and change `build` command to use `--builds-dir`

hussainsultan:
generally, i think this cli could be simplified if we were to just take the path to the build directory e.g. `builds/[hash_name]/` instead of providing both hash_name and builds directory as options. Perhaps, just a single option like `--path` that `xorq run` takes. any thoughts?

mesejo:
> generally, i think this cli could be simplified if we were to just take the path to the build directory e.g. `builds/[hash_name]/` instead of providing both hash_name and builds directory as options. Perhaps, just a single option like `--path` that `xorq run` takes. any thoughts?

We could do that. The only thing is that the user would need to write the path always. Right now, if the directory is the default one, we only need to supply the hash. Additionally, in the future, we could make it so that an environment variable supplies the default path, therefore allowing the liberty of having a different default while at the same time only supplying the hash.

dlovell:
@mesejo here's this branch on top of #598

python/xorq/cli.py
```diff
@@ -57,6 +57,36 @@ def build_command(script_path, expression, builds_dir="builds"):
...
+    Returns
+    -------
+
+    """
+    try:
+        build_manager = BuildManager(builds_dir)
+        expr = build_manager.load_expr(hash_id)
+        frame = expr.execute()
+
+        match output_format:
```
dlovell, line 79: if we `execute`, then we have to pull all the data into memory at once

i think the default format should be `parquet`

dlovell: if the user does not explicitly provide an output format, perhaps the "default" should be writing parquet to `/dev/null`

dlovell: (but still parquet so there's less conversion overhead)

mesejo: If we are going to write files, I prefer to make it so that each format writes to disk

python/xorq/cli.py
```diff
@@ -57,15 +58,51 @@ def build_command(script_path, expression, builds_dir="builds"):
...
+        expr = build_manager.load_expr(expr_path.stem)
+
+        match output_format:
+            case "csv":
+                expr.to_csv(output_path)
+            case "json":
+                expr.to_json(output_path)
+            case "parquet":
+                expr.to_parquet(output_path)
+
```
dlovell, line 93: should we be handling the "wildcard" / `_` case?

dlovell (approved):

PR #584, opened 2025-02-28
feat: add build method to Expr

mesejo:


mesejo:
As per convo, we close this PR till we have a better understanding of the right API

PR #596, opened 2025-03-03
feat: wip

mesejo:


dlovell:
FlightUDFX and FlightExpr are also opaque nodes.

Perhaps we should keep a reference to known opaque nodes somewhere. graph_utils.find_all_sources also needs to know them all

hussainsultan (approved):
LGTM

PR #613, opened 2025-03-06
fix(caching): mark remote tables so clean up won't interfere with others

dlovell:


mesejo (approved):

python/xorq/expr/relations.py
```diff
@@ -401,7 +411,7 @@ def register_and_transform_remote_tables(expr):
 
     def mark_remote_table(node):
         schema, batchess = batches_table[node]
-        name = f"{node.name}_{len(batchess)}"
+        name = f"{node.name}_cu{next(_count)}_t{len(batchess)}"
```
mesejo, line 414: Why do we need this count?

dlovell: If it's not there the added test fails

PR #614, opened 2025-03-07
feat: add llm example

mesejo:


examples/flight_udtf_llm_example.py
```diff
@@ -0,0 +1,82 @@
...
+
+do_hackernews_sentiment_udxf = xo.expr.relations.flight_udxf(
+    process_df=get_hackernews_sentiment_batch,
+    maybe_schema_in=schema_in.to_pyarrow(),
+    maybe_schema_out=schema_out.to_pyarrow(),
+    name="HackerNewsSentimentAnalyzer",
+    # make_server=lambda: FlightServer(flight_url=FlightUrl(port=5005)),
+)
+
+path = Path(__file__).parent / "data" / "hn_nano.parquet"
```
hussainsultan, line 65: Not sure if we should be checking in data for examples. If this is required, it should be something like pins.

examples/flight_udtf_llm_example.py
```diff
@@ -0,0 +1,93 @@
...
+schema_out = xo.schema({"text": "!str", "sentiment": "!str"})
+
+do_hackernews_sentiment_udxf = xo.expr.relations.flight_udxf(
+    process_df=get_hackernews_sentiment_batch,
+    maybe_schema_in=schema_in.to_pyarrow(),
+    maybe_schema_out=schema_out.to_pyarrow(),
+    name="HackerNewsSentimentAnalyzer",
+)
+
+hn = xo.examples.hn_posts_nano.fetch(table_name="hackernews")
```
dlovell, line 77: can we re-upload this file so the fields are text rather than binary?

mesejo: Done

mesejo: I misread the original message; now all the binary fields are strings!

PR #630, opened 2025-03-08
feat(ibis): enable serde for connections

dlovell:


dlovell:
i think this works "as is" but could be misleading for local engines: deserialized engines will not have any of the original con's tables registered

perhaps our yaml serialization is the way to go

dlovell:
on second thought: a deferred pipeline shouldn't rely on the local connections having tables in them

having this capability enables us to pass in an `unbound_expr` into `FlightExpr` that has connections in them that are expected to do work

mesejo (approved):

PR #637, opened 2025-03-10
chore(docs): add run section of build and run tutorial

mesejo:


docs/tutorials/build.mdx
```diff
@@ -1,8 +1,15 @@
 ---
-title: 'Building with xorq'
+title: 'Building and Running Expressions'
 ---
 
-xorq provides a command-line interface with a `build` command that allows you to generate artifacts from Ibis expressions defined in Python scripts. This tutorial will walk you through using this functionality.
+The xorq CLI provides two powerful commands, `build` and `run`, that help you separate the definition of data transformations from their execution. This separation enables:
+
+- **Reproducibility**: Define transformations once and run them consistently across different environments
```
hussainsultan, line 7: Reproducibity and Repliability sounds very close. 

Also Deployment and Version Control is the same thing. It seems repeated. Not sure if decoupling applies.

docs/tutorials/build.mdx
```diff
@@ -60,17 +81,43 @@ This command will:
 You should see output similar to:
 ```
 Building expr from pipeline.py
-Written 'expr' to artifacts/2a3b46d8d3d0
+Written 'expr' to artifacts/3350466c8fcd
 ```
 
-## Use Cases
+<Warning>
+  Currently, we do not support the building of expressions in-memory tables
```
hussainsultan, line 88: I think it's more like we are yet to implement support for in-memory tables for serialization since we are still working out the best way to support tables/dataframes on disk. 

Perhaps:

> Currently, the `build` command does not support in-memory tables as it's unclear what the best approach for serializing/deserializing DataFrames for build artifacts would be.

PR #648, opened 2025-03-12
chore: add ml functions as top-level

mesejo:


python/xorq/ml.py
```diff
@@ -0,0 +1,5 @@
+from xorq.expr import ml
```
dlovell, line 1: what is the purpose of import `ml` into `ml`?

PR #654, opened 2025-03-14
docs: add quickstart

hussainsultan:
- update overview page
- add HN basic example to quickstart


docs/getting_started/quickstart.mdx
```diff
@@ -1,3 +1,95 @@
...
+nix run github:xorq-labs/xorq
+```
+
+## Quick Start
+
+### Write a pandas UDF
+
+```python
+import xorq as xo
+import xorq.vendor.ibis.expr.datatypes as dt
```
mesejo, line 25: I think this line can be:

```suggestion
import xorq.expr.datatypes as dt
```

docs/getting_started/quickstart.mdx
```diff
@@ -1,3 +1,95 @@
...
+
+#### Build
+
+xorq makes it easy to serialize the pipeline in a diffable and human-readable
+format, including YAML for expressions, compiled SQL, and deferred reads. Once
+these artifacts are checked into git, we can build validation, lineage, and
+documentation tools in the CI/CD process.
+
+```sh
+❯ xorq build scrap.py -e expr
```
mesejo, line 65: I think the name of the file is "example.py"

docs/getting_started/quickstart.mdx
```diff
@@ -1,3 +1,95 @@
...
+# Reference the parquet file
+name = "hn-data-small.parquet"
+expr = xo.deferred_read_parquet(
+    con,
+    xo.options.pins.get_path(name),
+    name,
+).mutate(**{"url_in_title": url_in_title.on_expr})
+
+# Display results
+print(expr.execute().head())
```
mesejo, line 50: This will execute the expression `expr` twice when building and running with the CLI. Perhaps we can have:

```suggestion
if __name__ == "__main__":
    print(expr.execute().head())
```

hussainsultan: that makes sense, but I rather not make the example verbose. I don't think it comes in the way of the example point even if it is run twice.

mesejo (approved):

PR #663, opened 2025-03-15
docs/hn/part2

hussainsultan:


docs/tutorials/hn_transform_tfidf.mdx
```diff
@@ -0,0 +1,405 @@
...
+    )
+    .pipe(m.do_hackernews_fetcher_udxf)
+)
+
+# Process the data as we did in Part 1
+processed_expr = (
+    raw_expr
+    .filter(xo._.text.notnull())
+    .pipe(o.do_hackernews_sentiment_udxf, con=con)
+    .cache(storage=ParquetStorage(con))
```
mesejo, line 103: Why are we caching here, not at the Expr's end?

dlovell: We cache after every meaningful amount of work and hitting the openai API is meaningful work we don't want to repeat

mesejo (approved):

PR #677, opened 2025-03-18
fix: avoid duplicated execution of expr by removing read_all

mesejo:


python/xorq/flight/server.py
```diff
@@ -129,14 +129,12 @@ def _make_flight_info(self, query):
         # Execute query to get schema and metadata
         kwargs = loads(query)
         expr = kwargs.pop("expr")
-        result = self._conn.to_pyarrow_batches(expr, **kwargs).read_all()
+        result = self._conn.to_pyarrow_batches(expr, **kwargs)
```
dlovell, line 132: will `to_pyarrow_batches` register any record batch placeholder tables in the case of an expr that uses a `RemoteTable` / `into_backend`?

mesejo: Added a test with into_backend

dlovell: when i run this
```
import xorq as xo


con0 = xo.connect()
con1 = xo.connect()


t = (
    xo.examples.get_table_from_name("diamonds", con0)
    .into_backend(con1)
)
rbr = t.to_pyarrow_batches()
print(con1.list_tables())
```

i see this
```
['ibis_rbr-placeholder_6jtbjfkatvg57k_cu0_t1']
```
which means we pollute the namespace of backends when we invoke `to_pyarrow_batches`

dlovell: can we get the schema directly from `expr` instead of invoking `to_pyarrow_batches`?

PR #680, opened 2025-03-18
feat: add multi DuckDB with FlightServer

mesejo:


python/xorq/flight/backend.py
```diff
@@ -133,9 +134,11 @@ def register(
                 source.schema, source.to_batches()
             )
 
-        if isinstance(source, pa.RecordBatchReader):
-            self.con.upload_batches(table_name, source)
+        return self.read_record_batches(source, table_name=table_name)
 
+    def read_record_batches(self, source, table_name=None):
```
hussainsultan, line 139: could we please add a test for this

PR #736, opened 2025-03-26
feat(postgres): connect_examples: pass kwargs

dlovell:
enables users to use hit the examples server for caching with
```
pg = xo.postgres.connect_examples(database="caching")
```


mesejo:
So caching have write permissions?

dlovell:
> So caching have write permissions?

no, they can read from it. i've staged the `complex_cached_expr` tables to the `caching` database

mesejo (approved):

dlovell:
note: we have intermittent failures for `ci-test-examples` where evaluation of an `ExprScalarUDF` deferred model doesn't return a value

PR #739, opened 2025-03-27
fix(tests): use connect_env for Postgres test setup

### LINKED ISSUE #737: Make Postgres test connection configurable via environment variables
**Problem**  
The Postgres connection parameters in the test suite (host, port, user, password, database) are currently hardcoded. This can cause issues for contributors who have different local setups (e.g., Postgres running on a different port or with different credentials).

**Suggestion**  
Update the test setup to read connection parameters from environment variables:

- `POSTGRES_HOST` (default: localhost)
- `POSTGRES_PORT` (default: 5432)
- `POSTGRES_USER` (default: postgres)
- `POSTGRES_PASSWORD` (default: postgres)
- `POSTGRES_DATABASE` (default: ibis_testing)

This makes the test suite more flexible and easier to run in different environments.

I’m happy to submit a PR if this change is welcome.

soheil-star01:
Summary
Replaced hardcoded Postgres connection parameters in tests with connect_env() for better configurability.

Fixes #737

### DIFF
--- a/python/xorq/backends/let/tests/test_api.py
+++ b/python/xorq/backends/let/tests/test_api.py
@@ -51,7 +51,14 @@ def test_executed_on_original_backend(parquet_dir, csv_dir, mocker):
 
 
 def test_read_postgres():
-    uri = "postgres://postgres:postgres@localhost:5432/ibis_testing"
+    con = xo.postgres.connect_env()
+    uri = (
+        f"postgres://{con.con.info.user}:"
+        f"{con.con.info.password}@"
+        f"{con.con.info.host}:"
+        f"{con.con.info.port}/"
+        f"{con.con.info.dbname}"
+    )
     t = xo.read_postgres(uri, table_name="batting")
     res = xo.execute(t)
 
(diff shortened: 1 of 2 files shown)


python/xorq/backends/let/tests/test_api.py
```diff
@@ -51,7 +51,14 @@ def test_executed_on_original_backend(parquet_dir, csv_dir, mocker):
...
 def test_read_postgres():
-    uri = "postgres://postgres:postgres@localhost:5432/ibis_testing"
+    con = xo.postgres.connect_env()
+    uri = (
+        f"postgres://{con.con.info.user}:"
+        f"{con.con.info.password}@"
+        f"{con.con.info.host}:"
+        f"{con.con.info.port}/"
+        f"{con.con.info.dbname}"
+    )
```
mesejo, line 61: I think in here we can do: 
```suggestion
import os
uri = (
    f"postgres://{os.environ['POSTGRES_USER']}:"
    f"{os.environ['POSTGRES_PASSWORD']}@"
    f"{os.environ['POSTGRES_HOST']}:"
    f"{os.environ['POSTGRES_PORT']}/"
    f"{os.environ['POSTGRES_DATABASE']}"
)
```

So we don't need to create a con object

mesejo (changes requested):

dlovell:
@mesejo how do the correct environment variables get populated in gitlab ci? will we have a test that the environment variables are populated?

mesejo:
@soheil-star01 Thanks for this PR and reporting the issue!

dlovell:
note: closed in favor of https://github.com/xorq-labs/xorq/pull/776

PR #748, opened 2025-03-28
feat(cli): add default value to expr_name

dlovell:


python/xorq/cli.py
```diff
@@ -8,54 +8,50 @@
...
         )
         sys.exit(1)
 
     expr_hash = build_manager.compile_expr(expr)
     print(
-        f"Written '{expression}' to {build_manager.artifact_store.get_path(expr_hash)}"
+        f"Written '{expr_name}' to {build_manager.artifact_store.get_path(expr_hash)}",
+        file=sys.stderr,
     )
+    print(build_manager.artifact_store.get_path(expr_hash))
```
mesejo, line 54: Why are we printing this?

dlovell: per the commit message
> so scripts can use stdout to compose operations

mesejo: But we are also printing the previous line, should we print both?

mesejo (approved):

PR #759, opened 2025-03-30
refactor: remove tenacity from examples

### LINKED ISSUE #758: remove use of tenacity/retry
the example only pulls two values so we are unlikely to need retry

for any larger run, retry is more likely to cause api quota issues

mesejo:
closes #758

### DIFF
--- a/examples/flight_udtf_llm_example.py
+++ b/examples/flight_udtf_llm_example.py
@@ -6,11 +6,6 @@ from urllib.parse import unquote_plus
 import pandas as pd
 import toolz
 from openai import OpenAI
-from tenacity import (
-    retry,
-    stop_after_attempt,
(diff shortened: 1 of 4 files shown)


examples/flight_udtf_llm_example.py
```diff
@@ -31,7 +26,6 @@ def get_client():
 
 
 def extract_sentiment(text):
-    @retry(wait=wait_random_exponential(min=1, max=60), stop=stop_after_attempt(3))
     def completion_with_backoff(**kwargs):
```
dlovell, line 29: if we don't have `tenacity`/`retry`, this should no longer be called `completion_with_backoff`

dlovell (approved):

PR #760, opened 2025-03-30
fix(cli): don't execute __main__

dlovell:
import_from_path: enable passing module_name
build_command: set module_name if ipynb


dlovell:
merged as part of #761

Issue #752, opened 2025-03-29
`xorq run` on `Read` nodes doesn't properly handle `schema`


dlovell:
given this script
```
cat ./simple_example.py 
import xorq as xo


expr = (
    xo.examples.iris.fetch(backend=xo.connect())
    .filter([xo._.sepal_length > 5])
    .group_by("species")
    .agg(xo._.sepal_width.sum())
)


if __name__ == "__main__":
    res = expr.execute()
    print(res)
    pytest_examples_passed = True
```

i can execute
```
python ./simple_example.py 
      species  Sum(sepal_width)
0   Virginica             146.2
1  Versicolor             131.8
2      Setosa              81.7
```

i can `xorq build`
```
xorq build ./simple_example.py 
Building expr from ./simple_example.py
Written 'expr' to builds/73d2c228b10b
```

but i can't `xorq run`
```
xorq run ./builds/73d2c228b10b/
Error: argument 'schema': Expected instance of pyarrow.lib.Schema, got xorq.vendor.ibis.common.collections.FrozenOrderedDict
```

PR #766, opened 2025-03-30
fix(ibis_yaml): read with schema

dlovell:
depends on #765


dlovell:
> Is this related to #752?

yes, closes #752

mesejo (approved):

Issue #117, opened 2024-07-16
Organize the rust to python bindings (internal.py)


mesejo:
Currently, the bindings do not provide tab completion and are missing type information. Add it

PR #792, opened 2025-04-02
chore(rust): migrate to xorq-datafusion

### LINKED ISSUE #786: Add xorq-datafusion as a dependency
These include:
- add xorq-datafusion as a dependency
- remove all Rust code and related files(Cargo.lock, Cargo.toml, rust.toolchain, etc)
- modify GithubActions
- modify pre-commit
- update docs, README, CONTRIBUTING,

mesejo:
With this change, we migrate the Rust functionality to its repository. This has the following benefits:
- Faster CI times, no need to build the Rust side every time
- Less cache space used (no need to cache Rust artifacts)
- Faster publishing times, most of the time was dedicated to building the wheels for Windows
- More contributor friendly (no need to have Rust installed or knowledge of it)
- Automated updates via renovatebot (currently broken because Rust is missing in the environment, [example](https://github.com/xorq-labs/xorq/pull/791#issuecomment-2769911324))

Since removing all Rust code, we need a new [build backend](https://peps.python.org/pep-0517/#source-trees). The most popular option seems to be [hatchling](https://hatch.pypa.io/1.12/config/build/#build-system), 

```
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

Mainly because:

> Hatchling is a standards-compliant [1](https://hatch.pypa.io/1.13/config/build/#fn:1) build backend and is a dependency of Hatch itself. 

More on why hatch can be found [here](https://hatch.pypa.io/1.12/why/#build-backend)

To ensure that `uv build` does not produce empty wheels, we need to add the following configuration
to `pyproject.toml`:

```
[tool.hatch.build.targets.wheel]
sources = ["python"]

[tool.hatch.build.targets.sdist]
only-include = ["python"]
```

For more details, see this [issue](https://github.com/astral-sh/uv/issues/10066)

**Alternatives**

A possible alternative is to use [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/), which will keep all the code in a single repo but has the following drawbacks:

- renovatebot will continue to be broken
- The repo will be more cluttered (additional workflows required to handle the release of both libraries)
- loses the functionality of publishing on release, and the release history of both packages is going to be intertwined  

closes #786, #117


mesejo:
Tested:
- gcs: https://github.com/xorq-labs/xorq/actions/runs/14752765206
- s3: https://github.com/xorq-labs/xorq/actions/runs/14752318556

Additionally, the proposed workflow for working with xorq-datafusion was tested for the addition of this commit: https://github.com/xorq-labs/xorq-datafusion/commit/0898bbdf464ef647787c5cdaa6202925f02f0e8f

dlovell (approved):

PR #793, opened 2025-04-02
chore: update GitHub release workflows

### LINKED ISSUE #788: Modify the release Github workflows
We have two Github workflows: pre-release and release to publish to PyPI (and TestPyPI). We must modify them to use only uv and make them based on trusted publishing.

mesejo:
closes #788


mesejo:
depens on #792

PR #806, opened 2025-04-05
fix(cli): enable `CachedNode`

dlovell:


python/xorq/tests/test_cli.py
```diff
@@ -235,6 +236,7 @@ def test_examples(
     monkeypatch.setattr(sys, "argv", build_args)
     main_no_exit()
     captured = capsys.readouterr()
+    print(captured.err, file=sys.stderr)
```
mesejo, line 239: What is the purpose of this print?

mesejo (approved):

python/xorq/tests/test_cli.py
dlovell: we are capturing information that we would normally have access to, which makes debugging failed test invocations different and more difficult. the print restores access to some of that information in the normal manner

Issue #850, opened 2025-04-19
`xorq run` can't handle `curry`ied functions coming from main


dlovell:
due to underlying issue in `toolz`: https://github.com/pytoolz/toolz/issues/606

dlovell:
worked around in https://github.com/xorq-labs/xorq/pull/856

PR #851, opened 2025-04-19
fix(cli): retain expr script

dlovell:
possible fix for #850


dlovell:
this will likely be closed in favor of https://github.com/xorq-labs/xorq/pull/856

PR #852, opened 2025-04-20
fix: url for reading csv from s3

mesejo:


python/xorq/common/utils/aws_utils.py
```diff
@@ -18,8 +18,8 @@ def make_s3_credentials_defaults():
 def make_s3_connection():
     connection = {
         **make_s3_credentials_defaults(),
-        "aws.session_token": aws_config.get("AWS_SESSION_TOKEN", ""),
-        "aws.allow_http": aws_config.get("AWS_ALLOW_HTTP", "false"),
+        "aws.session_token": aws_config.get("AWS_SESSION_TOKEN") or "",
+        "aws.allow_http": aws_config.get("AWS_ALLOW_HTTP") or "false",
```
dlovell, line 22: @mesejo is this change for style or for function?

mesejo: When I do `aws_config.get("AWS_ALLOW_HTTP", "false")` returns the empty string. Should I modify `.get`?

dlovell: either we should modify the template file or we should set environment variables to the values we want (perhaps both)

i've created #862 to remove `.get` and do some other cleanup of the `EnvConfigable` api

dlovell (approved):

PR #856, opened 2025-04-20
fix(cli): enable __main__ `curry`/`functools.lru_cache`

dlovell:
depends on #849


dlovell:
note: this PR uses https://github.com/xorq-labs/toolz/tree/xorq-toolz

dlovell:
closes #850 by providing `toolz_utils.curry` to use in place of `toolz.curry` when expr must be `xorq run`-able

mesejo (approved):

PR #860, opened 2025-04-21
fix(env_utils): remove extraneous default from __getitem__

dlovell:


dlovell:
closing in favor of https://github.com/xorq-labs/xorq/pull/862

PR #900, opened 2025-04-29
chore: run tests mark with s3 in workflow

mesejo:


mesejo:
Tested:
- s3: https://github.com/xorq-labs/xorq/actions/runs/14740441996/job/41376895777

PR #910, opened 2025-05-05
Feat/pyiceberg daniel

mesejo:


hussainsultan (approved):

python/xorq/backends/pyiceberg/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+@translate.register(ops.Field)
+def column(op, **_):
+    return op.name
+
+
+@translate.register(ops.DatabaseTable)
+def table(op, **kwargs):
+    if "catalog" not in kwargs:
```
dlovell, line 39: can we do this instead?
```
def table(op, catalog=None, namespace=None):
    if catalog is None:
        raise ValueError("catalog must not be None")
    if namespace is None:
        raise ValueError("namespace must not be None")
    ...
```

PR #934, opened 2025-05-13
Feat/add glob interpretation

mesejo:


mesejo:
depends #927

dlovell:
@mesejo its not clear to me what the valid glob patterns are, can we add some tests with valid and invalid glob patterns?

PR #943, opened 2025-05-15
feat(nix): add python3.13

dlovell:


dlovell:
merged as part of #946

PR #950, opened 2025-05-19
 chore(flight): add TLSKwargsTuple class 

mesejo:


python/xorq/flight/__init__.py
```diff
@@ -130,10 +132,23 @@ def __init__(
...
+            self.certificate_path = str(tls_dir / "server.crt")
+            self.key_path = str(tls_dir / "server.key")
+            create_tls_keypair(self.certificate_path, self.key_path)
+            self.cleanup = True
+
         self.root_certificates = root_certificates
         self.auth = auth
         self.connection = connection
-        self.verify_client = verify_client
+        self.verify_client = False
```
dlovell, line 151: @mesejo why is `verify_client` always false?

dlovell: we want to be able to do something like this
https://github.com/apache/arrow/blob/apache-arrow-20.0.0/python/pyarrow/tests/test_flight.py#L2019-L2035

mesejo: I'm solving it in https://github.com/xorq-labs/xorq/pull/958

python/xorq/common/utils/tls_utils.py
```diff
@@ -159,49 +159,76 @@ def from_common_name(
             self = self.signed_with(sign_with)
         return self
 
+
+def get_client_tlscert(self):
+    return TLSCert.from_common_name(sign_with=self.ca_tlscert, common_name="client")
```
dlovell, line 164: what is the purpose of this function?

mesejo: Is a leftover

dlovell (approved):

PR #1000, opened 2025-06-09
feat: make FlightUDXF use mtls by default

mesejo:


python/xorq/expr/relations.py
```diff
@@ -314,16 +314,21 @@ def from_expr(
         name=None,
         **kwargs,
     ):
+        from xorq.common.utils.tls_utils import TLSKwargs
         from xorq.flight import FlightServer
 
+        def make_tls_server():
```
dlovell: i think this would more properly be called `make_mtls_server`

dlovell (approved):

PR #1019, opened 2025-06-14
feat(cli): add serve command

hussainsultan:


python/xorq/cli.py
```diff
@@ -114,8 +122,92 @@ def run_command(expr_path, output_path=None, output_format="parquet"):
...
+    build_manager = BuildManager(expr_path.parent)
+    if not build_manager.artifact_store.exists(expr_hash, "expr.yaml"):
+        raise ValueError(f"Error: expr.yaml not found in build directory {expr_path}")
+
+    expr = build_manager.load_expr(expr_hash)
+
+    if duckdb_path:
+        db_path = Path(duckdb_path)
+    else:
+        db_path = "xorq_serve.db"  # what should be the default?
```
dlovell: `db_path` must be a `Path`, but the `else` condition is only a `str`
`db_path = Path(duckdb_path or "xorq_serve.db")`

python/xorq/flight/metrics.py
```diff
@@ -0,0 +1,207 @@
...
+        return MetricExportResult.SUCCESS
+
+    def shutdown(self, *args, **kwargs) -> None:
+        return None
+
+    def force_flush(self, *args, **kwargs) -> bool:
+        return True
+
+
+def setup_console_metrics(
```
dlovell, line 69: should this only be executed once?

mesejo: This is executed when the 'serve' command is run. Would that be sufficient?

mesejo: @hussainsultan The argument 'duckdb_path' is not used inside setup_console_metrics. What is the purpose of this argument?

python/xorq/cli.py
```diff
@@ -114,8 +120,72 @@ def run_command(expr_path, output_path=None, output_format="parquet"):
...
+        setup_console_metrics(duckdb_path=str(db_path), prometheus_port=prometheus_port)
+    except ImportError:
+        logger.warning(
+            "Metrics support requires 'opentelemetry-sdk' and console exporter"
+        )
+
+    def _make_con():
+        return xo.duckdb.connect(str(db_path))
+
+    server = FlightServer.from_udxf(expr, connection=_make_con)
```
dlovell: rather than defining `_make_con`, we can
```
from functools import partial

server = FlightServer.from_udxf(expr, connection=partial(xo.duckdb.connect, str(db_path)))
```

dlovell: related: lets standardize the name to `make_connection` or similar in `FlightServer`, `FlightServerDelegate` and whatever else gets passed this value

python/xorq/flight/__init__.py
```diff
@@ -153,6 +153,39 @@ def __init__(
...
+        exchangers = []
+        seen = set()
+
+        # Find all FlightUDXF nodes in the expression
+        udxf_nodes = walk_nodes((FlightUDXF,), expr)
+        for node in udxf_nodes:
+            udxf_cls = getattr(node, "udxf", None)
+            if udxf_cls and udxf_cls not in seen:
+                exchangers.append(udxf_cls)
+                seen.add(udxf_cls)
```
dlovell: why do we need both `exchangers` and `seen`?
under what conditions will `getattr(node, "udxf", None)` return `None`?

python/xorq/expr/relations.py
```diff
@@ -345,11 +346,14 @@ def inner(flight_udxf):
                 rbr_in = instrument_reader(rbr_in, "input: ")
             with flight_udxf.make_server() as server:
                 client = server.client
-                client.do_action(
-                    AddExchangeAction.name,
-                    self.udxf,
-                    options=client._options,
-                )
+
+                if self.ephemeral:
```
dlovell: what is the utility of making this conditional?

python/xorq/flight/__init__.py
```diff
@@ -228,28 +255,17 @@ def __exit__(self, *args):
...
-    port=8815,
-    username=None,
-    password=None,
-    cert_chain=None,
-    private_key=None,
-    tls_root_certs=None,
-):
-    from xorq.flight.backend import Backend
-
+def connect(url, tls_kwargs):
```
dlovell: `tls_kwargs` should be optional

python/xorq/flight/tests/test_backend.py
```diff
@@ -51,3 +58,50 @@ def test_create_table_invalid_inputs(flight_server, obj):
...
+        verify_client=False,
+        exchangers=[dummy_udxf],
+        **tls_kwargs.server_kwargs,
+    )
+
+    def serve():
+        server.serve(block=True)
+
+    server_thread = threading.Thread(target=serve)
+    server_thread.daemon = True
```
dlovell:
```
server_thread = threading.Thread(
    target=functools.partial(server.serve, block=True),
    daemon=True,
)
```

python/xorq/flight/tests/test_backend.py
```diff
@@ -51,3 +58,50 @@ def test_create_table_invalid_inputs(flight_server, obj):
...
+
+    server_thread.start()
+    time.sleep(1)
```
dlovell: what is the purpose of this `sleep`?

python/xorq/flight/tests/test_metrics.py
```diff
@@ -0,0 +1,37 @@
...
+    setup_console_metrics()
+    dummy_do_get(None, None)
+
+    for i in range(20):
+        time.sleep(1)
+        if (
+            "flight_server.requests_total{method=dummy_do_get} 1"
+            in capsys.readouterr().out
+        ):
+            assert True
```
dlovell: why do we need `assert True`?

dlovell:
closed in favor of #1067

dlovell (approved):

PR #1021, opened 2025-06-15
feat: flight backend create

hussainsultan:


python/xorq/flight/backend.py
```diff
@@ -46,12 +46,21 @@ def create_table(
...
+            self.con.upload_batches(name, obj)
+            return self.table(name)
+        if isinstance(obj, Iterable):
+            batches_list = list(obj)
+            if not batches_list:
+                raise ValueError("Cannot create table from empty batch list")
+            schema = batches_list[0].schema
+            reader = pa.RecordBatchReader.from_batches(schema, batches_list)
+            self.con.upload_batches(name, reader)
+            return self.table(name)
```
dlovell: I think we should require that the user do this step: `upload_batches` is streaming and deals with a specific type, this admits any iterable and materializes in the case that its a generator of batches

dlovell (approved):

PR #1052, opened 2025-06-25
chore: fix links in README

mesejo:


hussainsultan:
do you wanna grab this commit also here: 56fbdcb (this add the cli section that got missed)

hussainsultan (approved):

PR #1138, opened 2025-07-17
feat(nix): nix run runs xorq cli

dlovell:


dlovell:
closed in favor of #1153

PR #1151, opened 2025-07-21
chore(docs): update 10_minutes_xorq_tour

mesejo:


hussainsultan (approved):
I think we should also add CLI commands section...

mesejo:
> I think we should also add CLI commands section...

Sure, I will add those in a subsequent PR

PR #1195, opened 2025-08-08
feat(compiler): warn when a Read op targets a local file

### LINKED ISSUE #1137: warn when `deferred_read_*` targets a local filesystem path

mesejo:
closes #1137

### DIFF
--- a/python/xorq/ibis_yaml/translate.py
+++ b/python/xorq/ibis_yaml/translate.py
@@ -540,11 +540,20 @@ def _remotetable_from_yaml(yaml_dict: dict, context: any) -> ibis.Expr:
 
 
 def warn_on_local_path(items: dict) -> None:
+    from urllib.parse import urlparse
+
+    def try_is_local_path(any: str | Path) -> bool:
+        try:
+            parsed = urlparse(any)
+            return not parsed.scheme or parsed.scheme in ("file",)
+        except ValueError | AttributeError:
+            return False
+
     items = dict(items)
     if path := next((items[k] for k in ("path", "source") if k in items), None):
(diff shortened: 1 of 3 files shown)


python/xorq/ibis_yaml/translate.py
```diff
@@ -529,13 +532,28 @@ def _remotetable_from_yaml(yaml_dict: dict, context: any) -> ibis.Expr:
     return remote_table_expr
 
 
+def warn_on_local_path(items: dict) -> None:
+    items = dict(items)
+    if path := next((items[k] for k in ("path", "source") if k in items), None):
+        paths = normalize_filenames(path)
+        try:
+            if any(path.exists() for path in map(Path, paths)):
```
dlovell: if any of the invocations of `Path(path).exists()` raises, the rest will be skipped which should not be the case
but also, `Path(path).exists()` is not a good test of whether a local path is being referenced

python/xorq/ibis_yaml/translate.py
```diff
@@ -536,13 +539,37 @@ def _remotetable_from_yaml(yaml_dict: dict, context: any) -> ibis.Expr:
...
+
+    items = dict(items)
+    if path := next((items[k] for k in ("path", "source") if k in items), None):
+        paths = normalize_filenames(path)
+        try:
+            if any(map(try_is_local_path, paths)):
+                warnings.warn(
+                    "The Read op path is using a local filesystem path, running the build may not work in other environments."
+                )
+        except TypeError:
```
dlovell: what might raise `TypeError`?

PR #1222, opened 2025-08-19
ref: move from xorq to xorq.api

mesejo:


python/xorq/backends/pandas/__init__.py
```diff
@@ -21,7 +21,7 @@ def do_connect(
 
         Examples
         --------
-        >>> import xorq as xo
+        >>> import xorq.api as xo
         >>> ls.pandas.connect({"t": pd.DataFrame({"a": [1, 2, 3]})})  # doctest: +ELLIPSIS
```
dlovell: something to fix for later: we're still using `ls` here

python/xorq/backends/postgres/__init__.py
```diff
@@ -7,7 +7,7 @@
 import sqlglot.expressions as sge
 import toolz
 
-import xorq as xo
+import xorq.api as xo
```
dlovell: we shouldn't use `xorq.api` within our own library code

python/xorq/caching/__init__.py
```diff
@@ -21,7 +21,7 @@
 from opentelemetry import trace
 from public import public
 
-import xorq as xo
+import xorq.api as xo
```
dlovell: its okay to use `xorq.api` in tests and examples but we shouldn't be using it within our own library code

dlovell (changes requested):
there are still multiple places where we `import xorq.api as xo` at the top level within our library code

we should be directly importing the desired functionality where possible and never be importing the full api at the top level

dlovell (approved):

PR #1283, opened 2025-09-11
feat: allow for specifying compiler in to_sql

### LINKED ISSUE #1065: Add `ArrayFilter` op to yaml translator
Compilation rule for 'ArrayFilter' operation is not defined

mesejo:
For testing this change, the `ArrayFilter` operation was added to the YAML translator. 

closes #1065

### DIFF
--- a/python/xorq/expr/api.py
+++ b/python/xorq/expr/api.py
@@ -7,10 +7,12 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Mapping
 
 import pyarrow as pa
+import toolz
 from opentelemetry import trace
 
 import xorq.vendor.ibis.expr.types as ir
 from xorq.backends.let import Backend
+from xorq.common.exceptions import XorqError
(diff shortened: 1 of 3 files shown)


python/xorq/expr/api.py
```diff
@@ -203,18 +204,14 @@ def read_postgres(
...
+def _cached_with_op(op, pretty, compiler):
     expr = op.to_expr()
-    sg_expr = con.compiler.to_sqlglot(expr)
+    sg_expr = compiler.to_sqlglot(expr)
     sql = sg_expr.sql(dialect=DataFusion, pretty=pretty)
     return sql
 
 
-def to_sql(expr: ir.Expr, pretty: bool = True) -> SQLString:
+def to_sql(expr: ir.Expr, compiler=Backend.compiler, pretty: bool = True) -> SQLString:
```
dlovell: i think we should default to the `expr`'s compiler (and possibly promote `normalize_expr`'s `get_compiler` to the `api` module for reuse

PR #1287, opened 2025-09-14
chore(postgres): improve testing

mesejo:


dlovell:
> @mesejo for some reason, i'm having difficulty adding a comment on line 20 of the new postgres conftest, but here's my comment for that
> 
> i believe in (most/many) other tests we are using `parquet_dir` from the top level `conftest.py` and explicitly trying to avoid using `pins`. is that the case and if so do we want to do that here as well?

oh, i see, `iris_path` is a csv. do we only pull down the parquet files from the `ibis-testing-data` in our ci/cd?

PR #1347, opened 2025-10-19
fix(snowflake): create table from expr

dlovell:


dlovell:
@mesejo this PR adds `xorq.expr.relations.prepare_create_table_from_expr` that I needed for a snowflake `into_backend`, but I think might be an issue with all remote backends

mesejo (approved):

PR #1354, opened 2025-10-23
feat: add from_ibis function

mesejo:


python/xorq/common/utils/ibis_utils.py
```diff
@@ -0,0 +1,114 @@
...
+    return FrozenOrderedDict(
+        tuple(
+            (map_ibis(key, kwargs), map_ibis(value, kwargs))
+            for key, value in frozendict.items()
+        )
+    )
+
+
+@map_ibis.register(IbisBaseBackend)
+def map_duckdb_backend(backend, kwargs):
```
dlovell: i think we should change this name to `map_backend`

python/xorq/common/utils/snowflake_keypair_utils.py
```diff
@@ -261,7 +262,9 @@ def maybe_decrypt_private_key(kwargs):
                 case str():
                     pass
                 case _:
-                    raise ValueError("expecting str and only str")
+                    warnings.warn(
```
mesejo, line 265: If the value is already decrypted is going to be bytest instead of a string, hence the warning @dlovell

dlovell (approved):

PR #1373, opened 2025-11-05
feat(let): add strftime function

### LINKED ISSUE #1372: Compilation rule for 'Strftime' operation is not defined
```
xorq.common.exceptions.OperationNotDefinedError: Compilation rule for 'Strftime' operation is not defined
```

mesejo:
closes #1372

### DIFF
--- a/python/xorq/backends/let/datafusion/udfs.py
+++ b/python/xorq/backends/let/datafusion/udfs.py
@@ -133,9 +133,9 @@ def regex_split(s: str, pattern: str) -> list[str]:
 
 
 def temporal_strftime(array: dt.Timestamp(scale=9), pattern: dt.string) -> dt.string:
-    pattern, *_rest = typing.cast(pa.StringArray, pattern).unique().to_pylist()
+    pattern, *rest = typing.cast(pa.StringArray, pattern).unique().to_pylist()
 
-    if len(_rest) > 0:
+    if rest:
         raise com.XorqError(
             "Only a single scalar pattern is supported for DataFusion strftime"
         )


python/xorq/backends/let/datafusion/udfs.py
```diff
@@ -129,3 +129,15 @@ def regex_split(s: str, pattern: str) -> list[str]:
         )
     pattern = patterns[0].as_py()
     return pc.split_pattern_regex(s, pattern)
+
+
+def temporal_strftime(array: dt.Timestamp(scale=9), pattern: str) -> dt.string:
+    patterns = pattern.unique()
```
dlovell: `pattern` has type decoration `str` but `str` does not have a `unique` method

mesejo: The type decoration is used to inferred the PyArrow type, when registering a UDF, true that it may be confusing, added a type cast, to make it clearer that patterns is a pyarrow StringArray

python/xorq/backends/let/datafusion/udfs.py
```diff
@@ -129,3 +130,14 @@ def regex_split(s: str, pattern: str) -> list[str]:
         )
     pattern = patterns[0].as_py()
     return pc.split_pattern_regex(s, pattern)
+
+
+def temporal_strftime(array: dt.Timestamp(scale=9), pattern: dt.string) -> dt.string:
+    pattern, *_rest = typing.cast(pa.StringArray, pattern).unique().to_pylist()
+
+    if len(_rest) > 0:
```
dlovell: let's use "truth value testing" here: `if rest:`
https://docs.python.org/3/library/stdtypes.html#truth-value-testing

PR #1390, opened 2025-11-18
docs: fix installation instructions for windows/macOS/Linux 

Olamideod:


docs/tutorials/getting_started/hn_data_prep.qmd
```diff
@@ -34,7 +34,7 @@ export OPENAI_API_KEY=your_api_key
 First, install Xorq and the required dependencies:
 
 ```bash
-pip install 'xorq[examples]'
```
mesejo, line 37: Could you please let me know why you're removing the quotes? If I run without quotes (on Ubuntu): 

```bash
$ pip install xorq[examples]  
zsh: no matches found: xorq[examples]

```

It works with quotes (either single or double).

Olamideod: So I'll just have to specify the Systems, because it doesn't work with quotes on Windows.

(.venv) C:\Users\oluwa\Xorq-Docs\xorq\xorq\docs>pip install 'xorq[examples]'
```
ERROR: Invalid requirement: "'xorq[examples]'": Expected package name at the start of dependency specifier
    'xorq[examples]'
    ^
```

mesejo: So I would leave the quotes, since most of our users are on Linux and macOS, and it would be consistent with other libraries such as Ibis and FastAPI. To solve the Windows problem, add a note:

```markdown
::: {.callout-note}
On Windows do pip install xorq[examples]
::: 
```

Olamideod: Ok that's perfect.

PR #1392, opened 2025-11-18
chore(docs): add Working with the documentation section

mesejo:


mesejo:
> I saw this bug: `psycopg-c can’t build because PostgreSQL dev tools (pg_config) aren’t installed` while running this command: `uv sync --all-extras --all-groups`
> 
> I had to run this command: `sudo apt install -y libpq-dev postgresql-client build-essential python3-dev ` so I could follow up with the steps.

I've change the dependency to the binary to avoid any need for installing system dependencies, see this commit:
https://github.com/xorq-labs/xorq/commit/8b66407ed1205c4ff9242a19cf0d8391fcda4b7d

PR #1394, opened 2025-11-18
fix(cli): remove unused ls command

mesejo:


python/xorq/tests/test_catalog_cli.py
```diff
@@ -17,7 +17,7 @@ def isolate_catalog(tmp_path, monkeypatch):
 
 
 def test_ls_empty_catalog(capsys):
-    args = parse_args(["ls"])
+    args = parse_args(["catalog", "ls"])
```
hussainsultan, line 20: I wonder if we need to add this since `catalog ls` is already hooked up.

mesejo: The `parse_args` function is the one actually used to parse the args received in the CLI, since we always expect to be "catalog ls", the input `["catalog", "ls"]` is appropriate.

PR #1410, opened 2025-11-24
chore(doc): added vale linter integration guide to the repo

### LINKED ISSUE #1409: add Vale linting integration guide for the docs

Olamideod:
fixes #1409

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ To build or preview the documentation locally, follow the steps below.
 
    ```bash
    cd docs  # ensure you are in the docs directory
-   uv run --no-sync quartodoc build --verbose --config docs/_quarto.yml
+   uv run --no-sync quartodoc build --verbose --config _quarto.yml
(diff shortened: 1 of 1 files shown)


Olamideod:
> @Olamideod Please rebase this branch onto main, so only the relevant different files are added. It will be also easier to review if it was only one single commit

Done

mesejo:
So the style is on top level not inside docs? For me this PR should only add the LINTING.md

mesejo:
> I placed styles/ in the repo root because that’s where .vale.ini lives, and StylesPath is relative to that file. This way, anyone can just run vale docs/ from the root, the linter “just works,” and docs stay as content.

If I change the StylesPath in vale.ini to:
`StylesPath = docs/styles `

It works. I can do it successfully:

```
vale docs/
``` 

We already have other config files (for example, the quarto ones) inside the docs dir, so it is ok to have other files besides just content

Olamideod:
Yeah that's true. I'll update the path.

mesejo (approved):

PR #1455, opened 2025-12-10
feat(caching): parquet ttl cache

dlovell:
this pr adds a ttl cache (`ParquetTTLSnapshotCache`)
in order to do this, we restructure the cache classes and change their name from `*Storage` to `*Cache`

### DIFF
--- a/python/xorq/expr/ml/pipeline_lib.py
+++ b/python/xorq/expr/ml/pipeline_lib.py
@@ -21,6 +21,7 @@ from xorq.backends.let import connect
 from xorq.caching import (
     ParquetCache,
     ParquetSnapshotCache,
+    ParquetTTLSnapshotCache,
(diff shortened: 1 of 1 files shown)


python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -348,9 +348,9 @@ class FittedStep:
         converter=tuple,
     )
     target = field(validator=optional(instance_of(str)), default=None)
-    storage = field(
+    cache = field(
         validator=optional(
-            instance_of((ParquetStorage, SourceStorage, ParquetSnapshotStorage))
+            instance_of((ParquetCache, SourceCache, ParquetSnapshotCache))
```
hussainsultan: `ParquetTTLSnapshotCache` seems to be missing..

PR #1457, opened 2025-12-11
ref(compiler): simplify translate

mesejo:


python/xorq/ibis_yaml/translate.py
```diff
@@ -525,7 +519,7 @@ def _remotetable_to_yaml(op: RemoteTable, context: any) -> dict:
 
 
 @register_from_yaml_handler("RemoteTable")
-def _remotetable_from_yaml(yaml_dict: dict, context: any) -> ibis.Expr:
+def _remotetable_from_yaml(yaml_dict: dict, context: any) -> dict:
```
dlovell: I think change in return type here is erroneous: are we still returning `ibis.Expr`?

dlovell:
@mesejo I :heart: this PR !
besides my one comment on the return type of `_remotetable_from_yaml`, this looks good to go

python/xorq/ibis_yaml/translate.py
mesejo: Indeed, we are returning an `ibis.Expr`, good catch

PR #1464, opened 2025-12-17
feat(xorq): add new embedded backend ops

mesejo:


python/xorq/backends/xorq/datafusion/compiler.py
```diff
@@ -550,8 +552,8 @@ def visit_Aggregate(self, op, *, parent, groups, metrics):
 
     def visit_ArraySlice(self, op, *, arg, start, stop):
         array_length = self.f.array_length(arg)
-        start = self.f.coalesce(start, 0)
-        stop = self.f.coalesce(stop, array_length + 1)
+        start = self.cast(self.f.coalesce(start, 0), dt.int32)
```
dlovell: how did we pick `int32`?

mesejo: Since this is supposed to be less than the length of an array, it seemed like a sensible choice; without the cast, it defaults to float. int64 is also a good option.

PR #1473, opened 2025-12-23
docs: update README and add architecture/intro SVG diagrams

hussainsultan:


dlovell:
@hussainsultan re `intro-dark.svg#gh-dark-mode-only`: for the third column, `[EXECUTE]`, it says "same manifest" at the bottom, suggesting the same manifest can be used for both "local development" and "production scale". Currently, I think the manifest would have to change, in particular with a substitution of sources with knock on effects on hash values.

Conceptually, we could support that substitution by way of command line, but I don't believe we currently have that capability.

README.md
```diff
@@ -7,144 +7,362 @@
...
+```python
+expr = from_ibis(penguins).into_backend(xo.sqlite.connect())
+expr.ls.backends
+```
+```
+(<xorq.backends.sqlite.Backend at 0x7926a815caa0>,
+ <xorq.backends.duckdb.Backend at 0x7926b409faa0>)
+```
+
+## Expressions are tools, Arrow is the pipe
```
dlovell, line 103: `Arrow is the interchange protcol`?

README.md
```diff
@@ -7,144 +7,362 @@
...
+
+| Template | Description |
+|----------|-------------|
+| `penguins` | Minimal example: caching, aggregation, multi-engine |
+| `sklearn` | Classification pipeline with train/predict separation |
+
+## Bring your own
+
+```
+xorq init -t ./my-template or xorq init -t git@github.com:org/template. Templates are just directories with a manifest and expr.py. No plugin system to learn.
```
dlovell: I think this is mixing a code snippet and text.
We don't currently support arbitrary urls.
I believe all we need is `expr.py`

PR #1511, opened 2026-01-13
feat: add poc sklearn functionality in docs/sklearn_interop_design.md

ghoersti:
Added experimental scikit-learn compatibility layer with `ColumnTransformer` support

  This adds a proof-of-concept implementation that lets you use scikit-learn estimators within
  xorq pipelines, including support for ColumnTransformer to apply different transformations to
   different columns. Includes working examples and tests.
   
   There are plenty of things that could be better but the cool thing is that it should unlock alot of scitkit functionality the structor tries to do the known things first but when that fails it defaults to a generic wrapper that makes any scikit pipeline deferred. The draw back is deferred column level lineage and more granular caching. 
   
   Also .to_sklearn() will only work on a pipeline and only if the entire pipeline is sklearnish. 
   
-check here for complex with round trip [ColumnTransformer](https://github.com/xorq-labs/xorq/blob/xorq-sklearn-extend-experimental/examples/sklearn_column_transformer.py)for the 

-check here for [more sklearn examples](https://github.com/xorq-labs/xorq/blob/xorq-sklearn-extend-experimental/examples/sklearn_interop.py)

-higher level design changes check out [this markdown](https://github.com/xorq-labs/xorq/blob/xorq-sklearn-extend-experimental/docs/sklearn_interop_design.md)

### DIFF
--- a/docs/sklearn_interop_design.md
+++ b/docs/sklearn_interop_design.md
@@ -1,0 +1,620 @@
+# sklearn ↔ xorq Interoperability Design
+
+## Philosophy & Goals
+
+### Core Principle: Wrap, Don't Reimplement
+
+The fundamental design philosophy is to **wrap sklearn for deferred execution**, not to reimplement sklearn functionality within xorq. This means:
+
+- sklearn handles all ML complexity internally (fitting, transforming, feature engineering)
+- xorq provides the deferred execution layer and caching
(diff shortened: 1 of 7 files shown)


ghoersti:
It still needs more tests on `fit_lib` and on `pipieline_lib` `dask_normalize_other` and `test_sklearn_compat`

PR #1534, opened 2026-01-27
feat: add ColumnTransformer, FeatureUnion, Pipeline structer registra…

ghoersti:
Registers these sklearn containers with structer_from_instance for schema inference. Also fixes Step hashing for estimators with nested list params (like ColumnTransformer's column specs) and adds dask tokenization for all sklearn estimators.

  Key changes:
  - Schema inference for nested sklearn containers (known vs KV-encoded)
  - passthrough_columns tracking on Structer
  - freeze() now recursively handles tuples
  - BaseEstimator dask tokenizer registration
  
 refactored example:
 [bank_marketing.py](https://github.com/xorq-labs/xorq/blob/george/feat/add-ColumnTransformer-FeatureUnion-Pipeline/examples/bank_marketing.py)

### DIFF
--- a/python/xorq/expr/ml/tests/test_pipeline_lib.py
+++ b/python/xorq/expr/ml/tests/test_pipeline_lib.py
@@ -469,6 +469,148 @@ class TestDeeplyNestedPipelines:
         # Assert predictions match
         assert np.array_equal(predictions["predicted"].values, sklearn_preds)
 
+    def test_hybrid_deeply_nested_pipeline(self):
+        """Test depth-4 nested pipeline with hybrid output (mixed known + KV-encoded).
+
+        Pipeline structure:
+        - ColumnTransformer (HYBRID: 2 KV-encoded + 1 known-schema child)
+          - FeatureUnion (known schema)
+            - Pipeline (SimpleImputer -> StandardScaler)
+            - Pipeline (SimpleImputer -> StandardScaler)
+          - Pipeline (SimpleImputer -> OneHotEncoder) [KV-encoded: education]
+          - Pipeline (SimpleImputer -> OneHotEncoder) [KV-encoded: region]
+        - SelectKBest
+        - RandomForestClassifier
+
+        This tests the hybrid transform path where the ColumnTransformer has
+        both known-schema children (numeric features) and multiple KV-encoded
+        children (categorical features encoded separately).
+        """
+        import numpy as np
+        import pandas as pd
+        from sklearn.compose import ColumnTransformer
+        from sklearn.ensemble import RandomForestClassifier
+        from sklearn.feature_selection import SelectKBest, f_classif
+        from sklearn.impute import SimpleImputer
+        from sklearn.pipeline import FeatureUnion
+        from sklearn.pipeline import Pipeline as SklearnPipeline
+        from sklearn.preprocessing import OneHotEncoder, StandardScaler
+
+        from xorq.expr.ml.pipeline_lib import Pipeline
+
+        # Create sample data
+        np.random.seed(42)
+        n_samples = 100
+
+        data = pd.DataFrame(
+            {
+                "age": np.random.randint(18, 80, n_samples).astype(float),
+                "income": np.random.randint(20000, 150000, n_samples).astype(float),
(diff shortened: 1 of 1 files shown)


python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -150,7 +156,9 @@ class Step:
 
     typ = field(validator=instance_of(type))
     name = field(validator=optional(instance_of(str)), default=None)
-    params_tuple = field(validator=instance_of(tuple), default=(), converter=tuple)
+    params_tuple = field(
+        validator=instance_of(tuple), default=(), converter=_freeze_params
```
dlovell: would `freeze` suffice here?

python/xorq/expr/ml/structer.py
```diff
@@ -208,6 +210,11 @@ class Structer:
         default=None,
         converter=lambda x: tuple(x) if x is not None else None,
     )
+    passthrough_columns = field(
+        validator=optional(deep_iterable(instance_of(str), instance_of(tuple))),
+        default=(),
+        converter=lambda x: tuple(x) if x is not None else (),
```
dlovell: would `xorq.common.utils.func_utils.if_not_none(tuple)` suffice here?

python/xorq/expr/ml/structer.py
```diff
@@ -327,11 +334,24 @@ def register_object(instance, expr, features=None):
...
 def lazy_register_sklearn():
+    from sklearn.compose import ColumnTransformer
     from sklearn.feature_extraction.text import TfidfVectorizer
     from sklearn.feature_selection import SelectKBest
     from sklearn.impute import SimpleImputer
+    from sklearn.pipeline import FeatureUnion
+    from sklearn.pipeline import Pipeline as SklearnPipeline
     from sklearn.preprocessing import OneHotEncoder, StandardScaler
 
+    def _normalize_columns(columns):
```
dlovell, line 406: can we use structural pattern matching here with a `case _:` that raises if we have a case that we haven't handled?

python/xorq/expr/ml/pipeline_lib.py
ghoersti: yes resolved, using freeze directly

python/xorq/expr/ml/structer.py
```diff
@@ -373,3 +398,154 @@ def _(instance, expr, features=None):
...
+        - Passthrough columns are included in the schema when all children are known-schema.
+        """
+        transformers = getattr(instance, "transformers_", None) or instance.transformers
+
+        combined_fields = {}
+        any_kv_encoded = False
+        all_input_cols = []
+        passthrough_cols = []
+
+        for name, transformer, columns in transformers:
```
dlovell: we should use a list / dict comprehension instead of a loop wherever possible

python/xorq/expr/ml/fit_lib.py
```diff
@@ -106,6 +106,103 @@ def kv_encode_output(model, df):
...
+    """
+    Get named transformers dict from a container.
+
+    Works for both ColumnTransformer (named_transformers_) and FeatureUnion (named_transformers).
+    """
+    if hasattr(model, "named_transformers_"):
+        return model.named_transformers_
+    if hasattr(model, "named_transformers"):
+        return model.named_transformers
+    return dict(model.transformer_list)
```
dlovell: we can use structural pattern matching here

```
python -c 'class Bunch:
    def __init__(self, **kwargs):
        for name, value in kwargs.items():
            setattr(self, name, value)


def do_match(obj):
    match obj:
        case Bunch(a=value):
            print(f"a={value}")
        case Bunch(b=value):
            print(f"b={value}")
        case _:
            print(f"obj={obj}")



do_match(Bunch(a=1, b=2))
do_match(Bunch(b=2))
do_match(Bunch(c=3))'
a=1
b=2
obj=<__main__.Bunch object at 0x71e6c8b7ce10>
```

ghoersti:
@dan made a class `ChildInfo` and moved the logic from _analyze_containers into registrations in structer to seperate concerns from `fit_lib.py`.

PR #1541, opened 2026-01-29
refactor: consolidate get_structer_out and clarify KV-encoding proper…

ghoersti:
Traversing the container classes and returning a `Structer` is soley handled by `get_structer_out` converting to the schema is trivial using `get_schema_out`. We made several private methods to help with clarity for `get_structer_out`, we could turn this into a class if wanted or encapsulate the methods. 

| Case | `Structer.struct` | Return Type |
|------|-------------------|-------------|
| `is_kv_encoded=True` | `None` | `Array[Struct{key, value}]` |
| `struct_has_kv_fields=True` | `dt.Struct({...})` | `Struct` with some fields of type `Array[Struct{key, value}]` |
| Known schema | `dt.Struct({...})` | `Struct` with scalar fields (float64, etc.) |

## Changes to structer.py

- Rename `any_kv_encoded` → `struct_has_kv_fields` for clarity
- Remove unused `all_kv_encoded` property
- Rename `convert_hybrid` → `convert_struct_with_kv`
- Rename `get_convert_hybrid` → `get_convert_struct_with_kv`
- Add `has_kv_output` property combining is_kv_encoded OR struct_has_kv_fields
- Add explicit `BaseEstimator()` case in get_structer_out match statement
- Add `_` case raising TypeError for unexpected types
- Extract helper functions for container processing:
  - `_prefix_fields`: Add sklearn-style name__ prefixes
  - `_process_child_structer`: Handle KV vs known-schema children
  - `_process_ct_item`: Process ColumnTransformer items
  - `_process_fu_item`: Process FeatureUnion items
  - `_merge_ct_results`: Accumulate schema fields and kv_cols
  - `_get_remainder_fields`: Find unhandled columns for remainder
  - `_accumulate_pipeline_step`: Thread features through pipeline
  - `_normalize_columns`: Convert column specs to tuples
- Refactor get_structer_out to use toolz.reduce with helper functions

## Changes to fit_lib.py

- Update match statement to use `struct_has_kv_fields`
- Update `get_convert_struct_with_kv` method call

## Test additions

- Add TestGetStructerOutErrorHandling with 4 tests for error handling
- Add TestStructerHasKvOutput with 5 tests for has_kv_output property
- Add TestMakeKvTuple with 5 tests for KVEncoder._make_kv_tuple
- Add TestPrefixFields with 5 tests for _prefix_fields helper
- Add TestProcessChildStructer with 3 tests
- Add TestMergeCtResults with 5 tests
- Add TestGetRemainderFields with 4 tests
- Add TestProcessCtItem with 5 tests
- Add TestProcessFuItem with 3 tests
- Add TestAccumulatePipelineStep with 5 tests
- Add TestNormalizeColumns with 6 tests
- Update existing tests for renamed properties

### DIFF
--- a/python/xorq/expr/ml/structer.py
+++ b/python/xorq/expr/ml/structer.py
@@ -346,21 +346,26 @@ class Structer:
         feature_names = tuple(model.get_feature_names_out())
 
         def accumulate_sklearn_col_slice(acc, field_item):
-            """Accumulate (col_idx, slices) -> (new_col_idx, updated_slices)."""
+            """Accumulate (col_idx, slices) -> (new_col_idx, updated_slices).
+
+            For KV-encoded fields, delta is the count of feature names matching
+            the field prefix (e.g., "cat__color_red", "cat__color_blue" for "cat").
+            For non-KV fields, delta is always 1 (single column).
+            """
             col_idx, slices = acc
             field_name, field_type = field_item
             is_kv = KVEncoder.is_kv_encoded_type(field_type)
             start = col_idx
-            end = (
-                start
-                + sum(
+            delta = (
+                sum(
                     1
                     for fn in feature_names[start:]
                     if fn.startswith(f"{field_name}__")
                 )
                 if is_kv
-                else start + 1
+                else 1
             )
+            end = start + delta
             return (end, {**slices, field_name: (start, end, is_kv)})
 
         _, field_slices = toolz.reduce(
--- a/python/xorq/expr/ml/tests/test_structer.py
+++ b/python/xorq/expr/ml/tests/test_structer.py
@@ -2186,3 +2186,162 @@ class TestNormalizeColumns:
             _normalize_columns(123)
         with pytest.raises(TypeError, match="Unsupported columns type"):
             _normalize_columns({"a": 1})
+
+
+class TestConvertStructWithKv:
+    """Tests for Structer.convert_struct_with_kv column slicing.
+
+    These tests verify the accumulate_sklearn_col_slice logic correctly
+    computes column indices for hybrid output (KV + non-KV fields).
+    """
+
+    def test_kv_then_non_kv_slicing(self):
+        """Test slicing when KV field precedes non-KV field.
+
+        This catches a potential bug where end = start + (start + 1) instead of
+        end = start + 1 for non-KV fields. With incorrect logic:
+        - cat field: start=0, end=0+2=2 (correct)
+        - num field: start=2, end=2+(2+1)=5 (WRONG, should be 3)
+
+        The non-KV field would get wrong values or cause index errors.
+        """
+        import pandas as pd
+        from sklearn.compose import ColumnTransformer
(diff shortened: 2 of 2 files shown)


python/xorq/expr/ml/structer.py
```diff
@@ -265,6 +310,75 @@ def get_convert_array(self):
...
+            result = result.toarray()
+        feature_names = tuple(model.get_feature_names_out())
+
+        def accumulate_sklearn_col_slice(acc, field_item):
+            """Accumulate (col_idx, slices) -> (new_col_idx, updated_slices)."""
+            col_idx, slices = acc
+            field_name, field_type = field_item
+            is_kv = KVEncoder.is_kv_encoded_type(field_type)
+            start = col_idx
+            end = (
```
dlovell: this `start + (sum(...) if is_kv else (start + 1))`
is that what it should be?
even if it is, i think this would be more clearly expressed as
```
delta = ...
end = start + delta
```

python/xorq/ibis_yaml/translate.py
```diff
@@ -1132,3 +1132,28 @@ def _schema_to_yaml(schema: Schema, context: TranslationContext) -> dict:
...
+    from sklearn.base import BaseEstimator
+
+    @translate_to_yaml.register(BaseEstimator)
+    def _sklearn_estimator_to_yaml(
+        obj: BaseEstimator, context: TranslationContext
+    ) -> dict:
+        return freeze(
+            {
+                "op": "SklearnEstimator",
+                "pickled_estimator": serialize_callable(obj),
```
dlovell, line 1149: ideally, this would be a combination of `type(obj)` and the contents of  `obj.get_params()`
we can defer that for another time

python/xorq/expr/ml/structer.py
```diff
@@ -325,11 +439,280 @@ def register_object(instance, expr, features=None):
...
+            return (
+                _prefix_fields({col: input_schema[col] for col in cols}, name),
+                None,
+            )
+        case _:
+            child = get_structer_out(transformer, expr, features=cols)
+            return _process_child_structer(child, name, cols)
+
+
+def _merge_ct_results(acc, result):
```
dlovell, line 540: ideally we would have an `@frozen` attrs class for these operations
but we can leave that for another time

dlovell (changes requested):
I think the implementation of `accumulate_sklearn_col_slice` likely not correct

python/xorq/expr/ml/structer.py
dlovell: if this logic turns out to be incorrect, let's add a test that would have errored

ghoersti: logic is correct but naming is poor , but it does look like it WOULD parse incorrectly but does not. I think delta is clearer will implement

ghoersti:
> I think the implementation of `accumulate_sklearn_col_slice` likely not correct

it was correct but for sure looked like it could have parsed wrong it now is explicit about getting the delta for each preifix pattern or just returns 1 when its not kv .

PR #1558, opened 2026-02-03
George/feat/register compatible scikit

ghoersti:
Summary
- Register sklearn transformers via mixin base classes (`OneToOneFeatureMixin`, `ClassNamePrefixFeaturesOutMixin`, `SelectorMixin`) instead of individual classes
- Support estimators with both transform and predict methods (inductive clustering: KMeans, Birch, etc.)
- Reject transductive clustering algorithms (`DBSCAN`, `AgglomerativeClustering`) at deferred fit 
- Serialize sklearn estimators to YAML using type + get_params() instead of pickle
- added `dask.base.normalize_token.register(type(np.mean))` needed for `FeatureAgglomeration` 

Added new example [sklearn_home_run_prediction.py](https://github.com/xorq-labs/xorq/blob/george/feat/register-compatible-scikit/examples/sklearn_home_run_prediction.py) to showcase xorq handling complex pipelines

### DIFF
--- a/python/xorq/expr/ml/structer.py
+++ b/python/xorq/expr/ml/structer.py
@@ -784,7 +784,7 @@ def lazy_register_sklearn():
         """Get number of output features from a feature selector instance.
 
         Returns the number of features the selector will output, or None if
-        it cannot be determined at compile time.
+        it cannot be determined before execution.
         """
         match instance:
             case object(k=k):
@@ -807,13 +807,13 @@ def lazy_register_sklearn():
                 # SelectFromModel uses max_features or threshold
                 match max_f:
                     case None | object(__call__=_):
-                        # Can't determine at compile time
+                        # Can't determine before execution
                         return None
                     case _:
                         return min(max_f, n_features_in)
             case SelectFpr() | SelectFdr() | SelectFwe() | GenericUnivariateSelect():
                 # SelectFpr, SelectFdr, SelectFwe, GenericUnivariateSelect - alpha/threshold based
-                # Can't determine number of features at compile time
+                # Can't determine number of features before execution
                 return None
             case _:
                 return None
@@ -863,7 +863,7 @@ def lazy_register_sklearn():
         """Get number of output components from a transformer instance.
 
         Returns the number of components/clusters the transformer will output,
-        or None if it cannot be determined at compile time.
+        or None if it cannot be determined before execution.
         """
         # PCA, NMF, TruncatedSVD, FastICA, etc. use n_components
         if hasattr(instance, "n_components"):


python/xorq/expr/ml/fit_lib.py
```diff
@@ -226,15 +226,22 @@ def _inner_other(model, df, other, features, return_type, encoded_cols):
...
+
         kwargs = {
             # still need to add: fit, other, return_type, name_infix
             "expr": fitted_step.expr,
             "target": fitted_step.target,
             "features": fitted_step.features,
             "cache": fitted_step.cache,
         }
-        if fitted_step.is_predict:
+        if mode == "predict":
```
ghoersti, line 244: This was added for algorithms like `Kmeans` that have both a transform and predict.

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -366,11 +366,18 @@ class FittedStep:
...
-        # if we are predict, we must have target
-        if self.target is None and self.is_predict:
-            raise ValueError("Can't infer target")
+        # we must have at least transform or predict
+        if not (self.is_transform or self.is_predict):
+            raise ValueError("Step must have transform or predict method")
+        # if we are predict-only, we must have target (except for clustering)
+        if self.target is None and self.is_predict and not self.is_transform:
+            from sklearn.base import ClusterMixin
+
```
ghoersti, line 375: `ClusterMixin` its ok to not have a target because the algorithms are unsupervised but they do have a predict, so it can be ok to not pass a target and it has predict if it is a `ClusterMixin`

python/xorq/expr/ml/structer.py
```diff
@@ -712,46 +712,203 @@ def get_schema_out(sklearnish, expr, features=None):
...
+        """Get number of output features from a feature selector instance.
+
+        Returns the number of features the selector will output, or None if
+        it cannot be determined at compile time.
+        """
+        # SelectKBest uses k parameter
+        if hasattr(instance, "k"):
+            k = instance.k
+            if k == "all":
+                return n_features_in
```
ghoersti: small upfront ops to know columns

python/xorq/expr/ml/structer.py
```diff
@@ -712,46 +712,203 @@ def get_schema_out(sklearnish, expr, features=None):
...
+        OneHotEncoder,
+        PolynomialFeatures,
+        SplineTransformer,
+        TargetEncoder,
+    )
+
+    def _structer_from_maybe_kv_inputs(expr, features):
+        """Create Structer for transformers that preserve column structure.
 
+        If any input column is KV-encoded, output is KV-encoded (feature names
```
ghoersti, line 743: small func to handle struct and kv inputs

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -990,6 +1010,7 @@ def raise_on_unregistered(instance, step, expr, features, target):
 def lazy_register_sklearn():
     from sklearn.base import (
         ClassifierMixin,
+        ClusterMixin,
```
ghoersti, line 1013: this is simpler  than adding `ClusterMixin` support in structer because of transductive algorithms. I think they might also fail but we already cover `Birch` & `Kmeans` without it so there is no value add to add in `Structer`

python/xorq/expr/ml/structer.py
```diff
@@ -712,46 +712,211 @@ def get_schema_out(sklearnish, expr, features=None):
...
+                # SelectFromModel uses max_features or threshold
+                match max_f:
+                    case None | object(__call__=_):
+                        # Can't determine at compile time
+                        return None
+                    case _:
+                        return min(max_f, n_features_in)
+            case SelectFpr() | SelectFdr() | SelectFwe() | GenericUnivariateSelect():
+                # SelectFpr, SelectFdr, SelectFwe, GenericUnivariateSelect - alpha/threshold based
+                # Can't determine number of features at compile time
```
dlovell: let's find a word other than "compile" to use

ghoersti: @dan changed to `before execution`

PR #1560, opened 2026-02-04
docs: quickstart, landing, and getting-started updates

### LINKED ISSUE #1539: docs: finalize quickstart in collaboration with SME

Olamideod:
fixes #1539

## What this PR does

This PR fixes two major things that were blocking people from finishing the quickstart: the `serve-unbound` step failing with a confusing error, and `xorq init` failing on macOS because of SSL.

### 1. The node hash for `serve-unbound` stopped working

**The issue**: The docs told users to run `serve-unbound` with a specific hash. For many people, that hash no longer matched anything in their built expression, so they hit:
`ValueError: not enough values to unpack (expected at least 1, got 0)`

So the command looked right, but the hash was wrong for the current penguins template.

**The fix**: I updated the quickstart to use the hash that works with the current template so that you can copy-paste the command and it runs. I also added a short note: if you ever see that error again, for example, after a template or Xorq update, you can look up the correct hash in `builds/<BUILD_HASH>/expr.yaml` and use that instead.

### 2. macOS: `xorq init` failing with an SSL error

**The issue**: On macOS, `xorq init` often failed with an SSL certificate error during template download. That happens when Python isn’t using the system’s certificates, so the download can’t be verified.

**The fix**: I documented the fix that works: set your cert bundle with certifi, for example, `export SSL_CERT_FILE=$(python -m certifi)` and then run `xorq init` again. I added this to the quickstart and the installation/troubleshooting docs, along with a note on how to make it permanent if a user wants.

Together, these changes make the quickstart runnable for more people and easier to follow when something changes.


docs/getting_started/quickstart.qmd
```diff
@@ -113,10 +111,16 @@ cd penguins_example # <2>
...
 
-The template generates an `expr.py` file. This file contains a complete ML pipeline with data loading, train/test splitting, model training, and prediction.
-
+::: {.callout-note}
+If `xorq init` fails with `[SSL: CERTIFICATE_VERIFY_FAILED]` or "unable to get local issuer certificate", then Python cannot verify the download. Set the certificate bundle, then retry:
+```bash
+export SSL_CERT_FILE=$(python -m certifi)
+xorq init -t penguins -p penguins_example
+```
+See [Installation issues](troubleshooting/installation_issues.qmd#issue-ssl-certificate-verify-failed-on-macos) for details and a permanent fix.
```
DianaHackmamba, line 120: References should be in this format:

For..., see LINK.

docs/getting_started/quickstart.qmd
```diff
@@ -233,37 +219,37 @@ Run the server with your build hash:
...
 ```cmd
-xorq serve-unbound builds/%BUILD_HASH% --host localhost --port 8001 --to_unbind_hash da8ba93cc97709f4ad9502e16d71e8e3
+xorq serve-unbound builds/%BUILD_HASH% --host localhost --port 8001 --to_unbind_hash 0eb004b30d161217f4cba885038c5c02
 ```
 
 :::
 
-The command saves prediction results to `predictions.parquet`. You can open this file with any Parquet-compatible tool. Compatible tools include pandas, DuckDB, or Polars.
+::: {.callout-note}
+If you get an error like "not enough values to unpack", the template or Xorq version may have changed. Open `builds/<BUILD_HASH>/expr.yaml`, find the `CachedNode` with `name: xorq_cached_node_name_placeholder`, and use its full `snapshot_hash` value instead.
```
DianaHackmamba:
```suggestion
If you get an error like "not enough values to unpack", then the template or Xorq version may have changed. Open `builds/<BUILD_HASH>/expr.yaml`, find the `CachedNode` with `name: xorq_cached_node_name_placeholder`, and use its full `snapshot_hash` value instead.
```

DianaHackmamba (approved):
A couple minor issues, then g2g.

PR #1564, opened 2026-02-05
Chore/fix example deps

paddymul:


python/xorq/common/utils/postgres_utils.py
```diff
@@ -77,8 +77,8 @@ def adbc_ingest(
 
 def make_credential_defaults():
     return {
-        "user": "$POSTGRES_USER",
-        "password": "$POSTGRES_PASSWORD",
+        "user": postgres_config["POSTGRES_USER"],
+        "password": postgres_config["POSTGRES_PASSWORD"],
     }
```
paddymul, line 82: nope.  we don't want these changes

python/xorq/env_templates/.env.postgres.template
```diff
@@ -1,5 +1,5 @@
-POSTGRES_USER=
-POSTGRES_PASSWORD=
-POSTGRES_HOST=
-POSTGRES_PORT=
-POSTGRES_DATABASE=
+POSTGRES_USER=postgres
+POSTGRES_PASSWORD=postgres
+POSTGRES_HOST=localhost
```
paddymul: probably not needed

examples/flight_exchange_example.py
```diff
@@ -30,8 +42,8 @@ def train_batch_df(df):
 
 
 class IterativeSplitTrainExchanger(AbstractExchanger):
-    @classproperty
-    def exchange_f(cls):
+    @property
```
dlovell, line 45: i have a slight preference to use `@classproperty` here because the property is not instance dependent.
(its also not class dependent, so maybe we should have a `@staticproperty`)

python/xorq/tests/test_examples.py
```diff
@@ -17,6 +17,8 @@
     "mcp_flight_server.py",
     "duckdb_flight_example.py",
     "complex_cached_expr.py",
+    "xorq_build_and_run.py",
+    "weather_flight.py"
```
paddymul, line 21: I don't know why I needed to add this, the only change made to this file added a big doc string

python/xorq/common/utils/gcloud_utils.py
```diff
@@ -11,7 +11,7 @@
 from google.cloud import storage
 from toolz import curry
 
-from xorq.caching.storage import Storage
+from xorq.caching.storage import CacheStorage as Storage
```
dlovell: why do we import `as Storage` here?

examples/README.md
```diff
@@ -1,14 +1,145 @@
...
+brew install cmake libomp
+
+source .venv/bin/activate # Activate the venv
+# Install dependencies (from repo root)
+uv sync --extra examples --extra postgres
+
+# run the examples
+cd examples
+xorq build simple_example.py
+xorq run simple_example.py
```
dlovell: should this be
```
build_hash=$(xorq build simple_example.py)
xorq run "$build_hash" --format json --output-path - | head
```
?

examples/README.md
```diff
@@ -1,14 +1,145 @@
...
+
+python xorq_build_and_run.py iris_example.py # build and run in a single step
```
dlovell: should this be `iris_example.py` which differs from above's `simple_example.py`? or should this also be `simple_example.py`?

examples/README.md
```diff
@@ -1,14 +1,145 @@
...
 ```
 
 
+## Expressions vs. Scripts
+
+Most examples define a top-level `expr` variable and should be run with `python xorq_build_and_run.py <file>` (or `xorq build` + `xorq run` separately), which compiles the expression graph into a versioned YAML manifest and executes it. A few examples involve runtime behavior like starting servers or CLI subcommands — these are marked as **script** below and should be run directly with `python <file>`.
```
dlovell: i don't think we should be pointing people to `xorq_build_and_run.py`
why does this say "should be run with"?

examples/README.md
dlovell: if `xorq_build_and_run.py` is meant to make an example / expr immediately tangible, perhaps it should default to `--format json --output-path -`

examples/README.md
paddymul: yep.  I'll change it

examples/README.md
paddymul: agree.

ghoersti:
@paddymul  @dlovell  There are many instances still in the examples where we are doing a `make_pipeline` . 

for example in `pipeline_example.py` we are asserting our pipeline against  a `manual_df`  constructed expr using xorq internals. 

```
    sklearn_df = train_predict_sklearn(sklearn_pipeline, train, test, features, target)
    manual_df = test_predicted_manual.execute().reindex_like(sklearn_df)
    pipeline_df = test_predicted_pipeline.execute().reindex_like(sklearn_df)
```
Is it still useful to have `manual_df` ? Is to just show users how you could create it using the internals ?

python/xorq/common/utils/gcloud_utils.py
paddymul: addressed in fix branch

examples/README.md
paddymul: let me double check that there isn't crazy long output for any expression. want to make sure at least two results are shown.

examples/README.md
paddymul: rename it to example_run.sh it scopes it to just examples, add clarification that this isn't the normal way to run xorq, but it makes sense here

examples/README.md
paddymul: this is where we should explain example_run.sh , how it's not normal ...

paddymul: Done

examples/README.md
paddymul: it defaults to json now, pipes to  `head -n 10` and sends Stderr to `/dev/null` to hide the `BrokenPipe` exception.  I think its a pragmatic compromise

dlovell (approved):

PR #1565, opened 2026-02-06
feat: sklearn scorers

ghoersti:
Summary
   Handles all 56 metrics from `from sklearn.metrics import get_scorer_names`  there are two entry points    `deferred_sklearn_metric` &  `Pipeline.score_expr` . Both are standards from skearn and we support both completely for all scorers. scorers are a special subset of metrics that return float. We have a test to assert high fidelity matching between xorq and sklearn for all 56 scorer metrics. Additional work is needed in metrics.py to support the remaining 30 metrics confusion matrix , roc_curve etc , because of the return type but there are only 4 ish return types. 

Examples :
[sklearn_pipeline_scoring.py](https://github.com/xorq-labs/xorq/blob/george/feat/sklearn-scorers/examples/sklearn_pipeline_scoring.py)
[sklearn_deferred_metrics.py](https://github.com/xorq-labs/xorq/blob/george/feat/sklearn-scorers/examples/sklearn_deferred_metrics.py)
   
  The deferred metrics example shows composability with mutate. 
   
  - Introduces Scorer class (attrs frozen) that normalizes all scorer inputs (str, bare callable, _BaseScorer, or None) into a single representation with
  metric_fn, sign, kwargs, and response_method
  - FittedPipeline.score_expr() now accepts a unified scorer param instead of separate metric_fn + use_proba -- response method routing (predict / predict_proba /
   decision_function) is derived automatically from the scorer
  - Sign handling (neg_mean_squared_error -> sign=-1) is extracted from the scorer and applied in MetricComputation.__call__, so deferred metrics match
  sklearn.metrics.get_scorer conventions exactly
  - Default scorer selection uses model mixin type: ClassifierMixin -> accuracy, RegressorMixin -> r2, ClusterMixin -> adjusted_rand
  - Fixes KVEncoder.decode null-row handling and a DeferredFitOther caching bug where _inner_other wrapping was applied outside the @functools.cache boundary

  Breaking changes

  - FittedPipeline.score_expr(): metric_fn / use_proba replaced by scorer
  - Custom callables that aren't known sklearn scorer functions now require make_scorer() wrapping

  Test plan

  - TestScorerFromSpec: unit tests for every Scorer.from_spec input path (str, neg-str, callable, _BaseScorer, Scorer, None + model type, invalid)
  - TestPipelineScoringMatchSklearn: parametrized over all sklearn classification, regression, and cluster scorers -- asserts xorq.score() == 
  sklearn.get_scorer()(pipe, X, y) within 1e-9
  - TestScoreExpr: default scorer detection, string/callable/make_scorer inputs, expression return type, unknown model ValueError
  - test_sign_auto_detected: verifies neg_mean_squared_error returns -1 * mean_squared_error
  - Existing metric tests updated to new API and still pass

### DIFF
--- a/examples/expr_scalar_udf.py
+++ b/examples/expr_scalar_udf.py
@@ -25,7 +25,7 @@ features = (
 )
 target = "event_occurred"
 model_key = "model"
-prediction_key = ResponseMethod.PREDICT
+prediction_key = str(ResponseMethod.PREDICT)
 prediction_typ = "float32"
 
 
--- a/examples/pipeline_example.py
+++ b/examples/pipeline_example.py
@@ -19,7 +19,7 @@ from xorq.expr.ml.pipeline_lib import (
 
 STRUCTED = "structed"
 ORIGINAL_ROW = "original_row"
-PREDICTED = ResponseMethod.PREDICT
+PREDICTED = str(ResponseMethod.PREDICT)
 
 
 @toolz.curry


python/xorq/expr/ml/tests/test_metrics.py
```diff
@@ -664,3 +669,161 @@ def test_metric_kwargs_with_pipeline():
...
+        target="target",
+        pred_col="predicted",
+        scorer="neg_mean_squared_error",
+    ).execute()
+
+    assert mse_negative == -mse_positive
+    assert mse_positive >= 0  # MSE is always non-negative
+
+
+class TestScorerFromSpec:
```
dlovell: was using a test class an explicit choice?

python/xorq/expr/ml/tests/test_pipeline_lib.py
```diff
@@ -470,6 +461,312 @@ def test_non_kv_deeply_nested_pipeline(self):
...
+    """Categorize all sklearn scorers by their type based on internal module path."""
+    from sklearn.metrics import get_scorer, get_scorer_names
+
+    classification = set()
+    regression = set()
+    cluster = set()
+    proba = set()
+    multilabel = set()
+
+    for name in get_scorer_names():
```
dlovell: can we make this a comprehension?

python/xorq/expr/ml/tests/test_pipeline_lib.py
```diff
@@ -470,6 +461,312 @@ def test_non_kv_deeply_nested_pipeline(self):
...
+            classification.add(name)
+        elif "_regression" in module:
+            regression.add(name)
+        elif "_ranking" in module:
+            classification.add(name)  # ranking metrics are for classifiers
+        elif "_scorer" in module:
+            classification.add(name)  # likelihood ratios
+
+    return {
+        "classification": frozenset(classification),
```
dlovell: why is this a frozenset?

python/xorq/expr/ml/metrics.py
```diff
@@ -18,11 +19,163 @@
...
+                scorer_obj = get_scorer(scorer)
+                return cls._from_scorer_obj(scorer_obj)
+
+            case _BaseScorer():
+                return cls._from_scorer_obj(scorer)
+
+            case Scorer():
+                return scorer  # already resolved
+
+            case _ if callable(scorer):
```
dlovell: is this condition different from `object(__call__=_)`?

python/xorq/expr/ml/metrics.py
```diff
@@ -18,11 +19,163 @@
...
+            kwargs=scorer_obj._kwargs,
+            response_method=cls._resolve_response_method(scorer_obj),
+        )
+
+    @staticmethod
+    def _resolve_response_method(scorer_obj):
+        """Extract the response method string from a scorer object."""
+        raw = scorer_obj._response_method
+        match raw:
+            case tuple() as methods:
```
dlovell: this should be `case (method, *_)`

python/xorq/expr/ml/metrics.py
```diff
@@ -158,14 +319,17 @@ def deferred_sklearn_metric(
     ...     expr_with_proba,
     ...     target="target",
     ...     pred_col="predicted_proba",
-    ...     metric_fn=roc_auc_score
+    ...     scorer=roc_auc_score
     ... )
     """
+    s = Scorer.from_spec(scorer) if not isinstance(scorer, Scorer) else scorer
```
dlovell: let's name this `scorer` instead of `s`

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -940,33 +971,55 @@ def score_expr(self, expr, metric_fn=None, use_proba=False, **kwargs):
...
-        )
+        from xorq.expr.ml.metrics import Scorer, deferred_sklearn_metric
+
+        s = Scorer.from_spec(scorer, model=self.predict_step.model)
+
+        # Route predictions based on response_method
+        match s.response_method:
+            case ResponseMethod.PREDICT_PROBA:
+                expr_with_preds = self.predict_proba(expr)
+                pred_col = "predicted_proba"
```
dlovell: we can use `s.repsonse_method.value` here

python/xorq/expr/ml/structer.py
```diff
@@ -134,9 +134,12 @@ def decode_encoded_column(cls, df, features, encoded_col):
...
+        first_row = col.iloc[0]  # for empty columns that can be kv encoded
+        if first_row is None:
+            return df, features
         # remove encoded_col from features and append the columns it becomes
         new_features = tuple(c for c in features if c != encoded_col) + tuple(
-            item[KVField.KEY] for item in col.iloc[0]
+            item[KVField.KEY] for item in first_row
         )
         result_df = df.drop(columns=[encoded_col]).join(cls.decode(df[encoded_col]))
         return result_df, new_features
```
dlovell: this can be a match
```
match tuple(col):
    case ():
        pass
    case (None, *_):
        pass
    case (first_row, *_):
        pass
    case _:
        raise ValueError
```

dlovell (commented):
see comments

python/xorq/expr/ml/tests/test_metrics.py
ghoersti: It was to just help with grouping but I made them individual

python/xorq/expr/ml/tests/test_pipeline_lib.py
ghoersti: I just wanted to emphasize that they should not change but set works fine , I changed to set

python/xorq/expr/ml/pipeline_lib.py
ghoersti: @dan so I switeched from `predicted` -> `predict` and `predicted_proba` -> `predict_proba`. 
So sklearn does not have a default column name, we are making this decision. It does clean up the internal code I still think its clear that if I fit a model and the column name is predict that it is predicted. but I can revert back if you want 
So that the code now becomes 
```
method = ResponseMethod(s.response_method)
expr_with_preds = getattr(self, method.value)(expr)
pred_col = method.value
```

ghoersti: and avoids the match entirely.

PR #1576, opened 2026-02-11
feat(ml): sklearn-non-scorer-metrics

ghoersti:
Summary

  - Unified metric parameter: Merged the separate scorer and metric_fn
  parameters on deferred_sklearn_metric into a single metric parameter that
  accepts scorer name strings ("accuracy"), _BaseScorer objects, Scorer
  dataclasses, or bare sklearn metric callables (accuracy_score,
  confusion_matrix, etc.)
  - Renamed pred_col to pred: The parameter now accepts both column name
  strings and tuples of column names (for multilabel/clustering), so pred is
  more accurate
  - Non-scalar metric support: Auto-detected return types for metrics that
  return Structs (class_likelihood_ratios, homogeneity_completeness_v_measure),
   Arrays (confusion_matrix, pair_confusion_matrix, silhouette_samples), and
  curve Structs (roc_curve, precision_recall_curve, det_curve)
  - deferred_auc_from_curve: New function that computes AUC from a deferred
  curve expression 
  - 94% sklearn.metrics coverage: 61/65 actual metrics supported (4 excluded as
   unsupported return types)

  New examples

  - sklearn_deferred_metrics.py — updated with 3 metrics per pipeline
  (classifier, regressor, clusterer) including non-float64 types
  - examples/sklearn_deferred_metrics_inputs.py — covers all (target, pred) dispatch
  combinations: (str, str), (tuple, str), (tuple, tuple)
  - sklearn_deferred_auc_from_curve.py — demonstrates deferred_auc_from_curve
  with roc_curve, precision_recall_curve, and det_curve

### DIFF
--- a/examples/sklearn_deferred_metrics_inputs.py
+++ b/examples/sklearn_deferred_metrics_inputs.py
@@ -105,9 +105,8 @@ make_cluster_metric = deferred_sklearn_metric(
 )
 
 # Note: h/c/v needs ground-truth labels, not features — use (str, str) here
-cluster_df_with_labels = cluster_df.copy()
-cluster_df_with_labels["true_label"] = np.repeat([0, 1, 2], [67, 67, 66])
-cluster_labeled_expr = con.register(cluster_df_with_labels, "clusters_labeled")
+cluster_df["true_label"] = np.repeat([0, 1, 2], [67, 67, 66])
+cluster_labeled_expr = con.register(cluster_df, "clusters_labeled")
 
 cluster_metrics = cluster_expr.agg(
     # target=tuple of feature columns, pred=single label column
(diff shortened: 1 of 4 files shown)


examples/sklearn_deferred_metrics_inputs.py
```diff
@@ -0,0 +1,215 @@
...
+cluster_df["cluster"] = kmeans.fit_predict(X_cluster)
+cluster_expr = con.register(cluster_df, "clusters")
+
+make_cluster_metric = deferred_sklearn_metric(
+    target=cluster_features,
+    pred="cluster",
+)
+
+# Note: h/c/v needs ground-truth labels, not features — use (str, str) here
+cluster_df_with_labels = cluster_df.copy()
```
dlovell: if this copy is necessary for correctness then we are doing something wrong: we should not be mutating dfs
if this copy is not necessary for correctness it should be removed

examples/sklearn_deferred_metrics_inputs.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+if __name__ == "__pytest_main__":
+    # --- Section 1: (str, str) — classifier non-scalar metrics ---
+    print("=== Section 1: (str, str) — Classifier non-scalar metrics ===")
+    clf_result = clf_metrics.execute()
+    print(clf_result[["accuracy"]].to_string(index=False))
+
+    # Unpack confusion matrix: nested list -> 2D ndarray
+    cm = np.array(clf_result["confusion_matrix"].iloc[0])
```
dlovell: `.iloc[0]` should happen once at the original `.execute()`

python/xorq/tests/test_examples.py
```diff
@@ -57,5 +57,6 @@ def maybe_marks(name: str):
     ],
 )
 def test_script_execution(script):
+    print(f"Running script: {script}", flush=True)
```
dlovell: this should be removed

python/xorq/expr/ml/metrics.py
```diff
@@ -173,12 +174,125 @@ def _resolve_response_method(scorer_obj):
                 raise ValueError(f"Unexpected _response_method: {raw}")
 
 
+@functools.lru_cache(maxsize=1)
```
dlovell: lets maintain consistency and use `@functools.cache`

python/xorq/expr/ml/metrics.py
```diff
@@ -173,12 +174,125 @@ def _resolve_response_method(scorer_obj):
...
+        }
+    )
+
+    all_metric_fns = frozenset(
+        obj
+        for name in dir(sklearn.metrics)
+        if not name.startswith("_")
+        and name not in _EXCLUDED_NAMES
+        and callable(obj := getattr(sklearn.metrics, name))
+        and inspect.isfunction(obj)
```
dlovell: are both of these lines necessary?

python/xorq/expr/ml/metrics.py
```diff
@@ -277,63 +499,195 @@ def deferred_sklearn_metric(
...
-        metric_kwargs_tuple=merged_kwargs,
-        return_type=return_type,
-        name=name,
-    )
-    return metric.on_expr(expr)
+    from sklearn.metrics._scorer import _BaseScorer
+
+    # Helper to build MetricComputation from a Scorer
+    def _from_scorer(scorer, metric_kwargs):
+        merged_kwargs = {**dict(scorer.kwargs), **dict(metric_kwargs)}
```
dlovell: this can be `dict(scorer.kwargs) | dict(metric_kwargs)`

python/xorq/expr/ml/metrics.py
```diff
@@ -173,12 +174,125 @@ def _resolve_response_method(scorer_obj):
...
 
+@functools.lru_cache(maxsize=1)
+def _build_known_non_scorer_metric_fns():
+    """Set of known non-scorer metric functions.
+
+    Computed as all public functions in ``sklearn.metrics`` minus the
+    scorer functions from ``_build_known_scorer_funcs()``, excluding
+    pairwise/scorer utilities by module and unsupported metrics by name.
+    """
+    import inspect
```
dlovell: `import inspect` should be at the top: we only defer costly imports like numpy, sklearn

python/xorq/expr/ml/metrics.py
dlovell: i think this is equivalent
```
all_metric_fns = frozenset(
    obj
    for name, obj in sklearn.metrics.__dict__.items()
    if not name.startswith("_")
    and name not in _EXCLUDED_NAMES
    and inspect.isfunction(obj)
    and not obj.__module__.startswith(
        ("sklearn.metrics.pairwise", "sklearn.metrics._scorer")
    )
)
```

python/xorq/expr/ml/metrics.py
```diff
@@ -173,12 +174,125 @@ def _resolve_response_method(scorer_obj):
...
+        confusion_matrix,
+        det_curve,
+        homogeneity_completeness_v_measure,
+        pair_confusion_matrix,
+        precision_recall_curve,
+        roc_curve,
+        silhouette_samples,
+    )
+
+    return {
```
dlovell: we should make this frozen so it can't be mutated

examples/sklearn_deferred_metrics_inputs.py
ghoersti: It was just there to have a seperate input df only for clarity. Clustering needs integer labels so its just making it clear that it is a different input data set, but its not necessary I can remove it

python/xorq/expr/ml/metrics.py
ghoersti: `MappingProxyType` gives you an immutable read only view of a dict

PR #1585, opened 2026-02-16
ref: add attr_utils.{convert_sorted_kwargs_tuple,validate_kwargs_tuple}

dlovell:


python/xorq/common/utils/attr_utils.py
```diff
@@ -0,0 +1,10 @@
+from toolz import compose
+
+
+convert_sorted_kwargs_tuple = compose(tuple, sorted, dict.items, dict)
```
ghoersti, line 4: Oh this is right to left.

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -152,16 +157,18 @@ class Step:
...
+        converter=compose(freeze, convert_sorted_kwargs_tuple),
+        factory=tuple,
+    )
 
     def __attrs_post_init__(self):
         from sklearn.base import BaseEstimator
 
         assert BaseEstimator in self.typ.mro()
         if self.name is None:
             object.__setattr__(self, "name", f"{self.typ.__name__.lower()}_{id(self)}")
```
ghoersti, line 171: This is much better name

PR #1587, opened 2026-02-16
fix: cache hash collision

### LINKED ISSUE #1570: two exprs that are the same but for cache type result in the same build hash
this means that they would collide in the catalog despite being different

if i run
```
import warnings

import xorq.api as xo
from xorq.caching import ParquetCache
from xorq.ibis_yaml.compiler import build_expr


warnings.simplefilter("ignore")


t = xo.examples.penguins.fetch()
c0, c1 = (t.cache(*args) for args in ((), (ParquetCache.from_kwargs(),)))
b0, b1 = (build_expr(expr) for expr in (c0, c1))
print(b0, b1)
```

i see
```
builds/0c07c6d13636 builds/0c07c6d13636
```

hussainsultan:
closes #1570

### DIFF
--- a/python/xorq/common/utils/tests/test_dask_normalize.py
+++ b/python/xorq/common/utils/tests/test_dask_normalize.py
@@ -244,6 +244,17 @@ def test_patch_normalize_token():
     assert name_to_cls["to_retain"] in dask.base.normalize_token._lookup
 
 
+def test_parquet_cache_tokenize_stable_across_cloudpickle():
+    import cloudpickle
+
+    con = xo.connect()
+    cache = ParquetCache.from_kwargs(source=con)
+    token_before = dask.base.tokenize(cache)
+    cache2 = cloudpickle.loads(cloudpickle.dumps(cache))
+    token_after = dask.base.tokenize(cache2)
+    assert token_before == token_after
+
+
 def test_different_cache_types_produce_different_hashes():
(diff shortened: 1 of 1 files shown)


ghoersti:
@hussain since the bug in the test was failing in `flight_serve_model` I think adding test `1`  is certainly useful. For tests 2 & 3 these are more defensive tests that should always pass, just protects us if the internals change unexpectedly. 


  1. Make sure roundtrip deterministic 
  ```python
  import cloudpickle, dask.base
  cache = ParquetCache.from_kwargs(source=con)
  token_before = dask.base.tokenize(cache)
  cache2 = cloudpickle.loads(cloudpickle.dumps(cache))
  token_after = dask.base.tokenize(cache2)
  assert token_before == token_after
  ```

  2. Also make sure different backends dont collide, I dont think they will but test doesn't hurt imo 
  ```python
  cache_pg = SourceCache.from_kwargs(source=pg_con)
  cache_duck = SourceCache.from_kwargs(source=duck_con)
  assert dask.base.tokenize(cache_pg) != dask.base.tokenize(cache_duck)
```

  3. Same backend same tokens 
  ```python
  con_a = xo.connect()
  con_b = xo.connect()
  cache_a = ParquetCache.from_kwargs(source=con_a)
  cache_b = ParquetCache.from_kwargs(source=con_b)
  assert dask.base.tokenize(cache_a) == dask.base.tokenize(cache_b)
```

PR #1601, opened 2026-02-21
fix(process_utils): drain both pipes concurrently to prevent deadlock

### LINKED ISSUE #1600: failure to build the env in `uv_tool_run` causes silent hang
tests currently pass in ci/cd because the python version is fixed, but on my system uv will try to use python3.14 and has to build pyarrow but fails to do so
```
  help: `pyarrow` (v21.0.0) was included because `xorq` (v0.3.10) depends on `pyarrow
```

this causes `uv_tool_run` to try to raise and use `Popened.stderr` as the assertion failure message and this hangs

dlovell:
`communicated` previously only called `communicate()` when `stdout_peeker` was set, leaving stderr unread in other cases. Also adds a `wait()` method that drains pipes before waiting, and updates `packager.py` to use it instead of `popen.wait()` directly.

fixes #1600

### DIFF
--- a/python/xorq/common/utils/process_utils.py
+++ b/python/xorq/common/utils/process_utils.py
@@ -33,6 +33,11 @@ try_decode_ascii = toolz.excepts(
 )
 
 
+try_decode_utf8 = toolz.excepts(
+    AttributeError, operator.methodcaller("decode", "utf-8")
+)
+
+
 @frozen(eq=False)
 class Popened:
     args = field(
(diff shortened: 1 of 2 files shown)


python/xorq/common/utils/process_utils.py
```diff
@@ -72,13 +77,18 @@ def peek_stdout(self, size):
...
-            _stdout = buf + _stdout
+        # Read already-peeked bytes from the BytesIO buffer (non-blocking)
+        peeked = self.stdout_peeker.buf.read() if self.stdout_peeker else b""
+        # Always drain both pipes concurrently to avoid deadlock
+        (_stdout, _stderr) = self.popen.communicate()
+        if peeked:
+            _stdout = peeked + (_stdout or b"")
         return (_stdout, _stderr)
 
+    def wait(self):
```
ghoersti, line 88: I think this can be 

```python
 def wait(self):
      self.communicated
      return self.popen.returncode
```

PR #1611, opened 2026-02-23
feat(catalog): add tui flag with Textual DataTable viewer

hussainsultan:
Adds `xorq catalog tui` launching an interactive terminal UI that displays catalog entries in a refreshing DataTable with vim-style hjkl navigation.

### DIFF
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -13,13 +13,13 @@ def click_handler(e):
 def _init_hint(ctx):
     """Return the `xorq catalog ... init` command the user should run."""
     group_ctx = ctx.parent or ctx
-    parts = ["xorq catalog"]
-    if name := group_ctx.params.get("name"):
-        parts.append(f"--name {name}")
-    elif path := group_ctx.params.get("path"):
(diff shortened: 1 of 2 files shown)


python/xorq/catalog/cli.py
```diff
@@ -132,6 +135,21 @@ def cli(ctx, name, path, url, root_repo, init):
...
 
+@cli.command()
+@click.option("--refresh", default=10, type=float, help="Refresh interval in seconds.")
+@click.pass_context
+def tui(ctx, refresh):
+    """Launch terminal UI."""
+    with click_context_catalog(ctx):
+        ctx.obj.make_catalog(init=False)  # validate catalog exists
+    from xorq.catalog.tui import CatalogTUI
+
```
ghoersti, line 146: why not use `toolz.curry`?

hussainsultan: can you suggest what that may look like?

dlovell: @ghoersti `make_catalog` is a `functools.partial` on `Catalog.from_kwargs`

PR #1614, opened 2026-02-24
chore(ci): enable parallel testing and add trino backend to matrix

mesejo:
Core backend parallelism:
- Add pytest-xdist to the test dependency group
- Add parallel: true flag to core backend in ci-test.yml with -n auto --dist=loadfile
- Sort set values in get_scorers_by_type() to fix non-deterministic xdist collection
- Mark postgres-touching tests with pytest.mark.postgres across test_cache,
  test_ls_accessor, test_into_backend, test_deferred_read, test_profile, test_register_read

Flight test isolation:
- Add pytest_collection_modifyitems hook in flight/tests/conftest.py with
  xdist_group(name="flight") to keep all flight tests on a single worker,
  preventing port conflicts from concurrent server startup

Trino backend:
- Add trino backend matrix entry with --extra trino and --extra duckdb
- Mark trino-dependent tests with @pytest.mark.trino
- Fix test_catalog_ctor_fails regex to accept both master and main branch names

### DIFF
--- a/python/xorq/expr/ml/tests/test_pipeline_lib.py
+++ b/python/xorq/expr/ml/tests/test_pipeline_lib.py
@@ -7,6 +7,7 @@ import xorq.api as xo
 from xorq.common.utils.graph_utils import walk_nodes
 from xorq.expr.ml.enums import ResponseMethod
 from xorq.expr.relations import Tag
+from xorq.vendor.ibis.common.collections import FrozenOrderedDict
 
 
 sklearn = pytest.importorskip("sklearn")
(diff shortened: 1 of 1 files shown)


python/xorq/expr/ml/tests/test_pipeline_lib.py
```diff
@@ -505,11 +505,11 @@ def get_scorers_by_type():
     cluster = {name for name, module in non_ml if "cluster" in module}
 
     return {
```
ghoersti: Could be ordered dict? I don't feel strongly about it

PR #1615, opened 2026-02-24
feat(cli): add structured logging and telemetry to run_command

mesejo:
Adds file-based structlog logging (get_logger, not get_print_logger) and richer OTel instrumentation to run_command, consistent with serve_command.

Changes:
- run.start log + span event captures all input params at entry
- Per-phase timing: load_expr and arbitrate_output_format measured separately, emitted as run.expr_loaded and run.output_written span events
- Output metrics on run.output_written: bytes (stat) and rows (pq.read_metadata, parquet footer only — no extra query) when writing to a real file path
- Explicit span status: StatusCode.OK on success, StatusCode.ERROR + span.record_exception on failure; previously all spans appeared successful
- Exceptions also logged via logger.exception for file-log capture

Alternatives considered and rejected:
- get_print_logger: ruled out in favour of silent file-only logging
- expr.count().execute() for row count: replaced by pq.read_metadata (fast, reads parquet footer only) to avoid a second query round-trip
- run_unbound_command: deferred, only run_command addressed here

### DIFF
--- a/python/xorq/tests/test_cli.py
+++ b/python/xorq/tests/test_cli.py
@@ -9,6 +9,7 @@ import pytest
 
 import xorq.api as xo
 from xorq.cli import (
+    OutputFormats,
     build_command,
     run_command,
 )
@@ -187,7 +188,7 @@ def test_run_command_default(tmp_path, fixture_dir):
 
 
 @pytest.mark.slow(level=1)
-@pytest.mark.parametrize("output_format", ["csv", "json", "parquet"])
+@pytest.mark.parametrize("output_format", OutputFormats)
 def test_run_command(tmp_path, fixture_dir, output_format):
     target_dir = tmp_path / "build"
     script_path = fixture_dir / "pipeline.py"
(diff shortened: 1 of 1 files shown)


python/xorq/cli.py
```diff
@@ -199,11 +204,53 @@ def run_command(
...
+        output_format=str(output_format),
+        limit=limit,
+    )
 
-    expr = load_expr(expr_path, cache_dir=cache_dir)
-    if limit is not None:
-        expr = expr.limit(limit)
-    arbitrate_output_format(expr, output_path, output_format)
+    try:
+        t_load = time.monotonic()
```
hussainsultan: we should pull time into a context manager:
```
from contextlib import contextmanager

@contextmanager
def timed(span, logger, event_name, **extra):
    t = time.monotonic()
    yield lambda: time.monotonic() - t
    elapsed = time.monotonic() - t
    payload = {"elapsed_s": elapsed, **extra}
    span.add_event(event_name, payload)
    logger.info(event_name, **payload)
```

and parquet_metrics could be pulled out

```
def _parquet_metrics(path: Path) -> dict:
    try:
        import pyarrow.parquet as pq
        return {"rows": pq.read_metadata(path).num_rows}
    except Exception:
        return {}
```

hussainsultan (approved):

PR #1631, opened 2026-02-26
feat(run): persist structured run logs

mesejo:
Each xorq run invocation now writes a unique run record to disk under ~/.local/share/xorq/runs/<expr_hash>/<expr_hash>-<timestamp>/:

  run.jsonl  — append-only event log (run.start, expr_loaded, done,
               output_written) with per-event timestamps and elapsed_s
  meta.json  — summary written at completion with status, params,
               started_at/completed_at, and the OTel trace_id for
               cross-reference with telemetry

A _NullRunLogger fallback ensures the actual run is never blocked if the store directory cannot be created. Read-side helpers (list_expr_hashes, list_runs, read_meta, read_events) expose the on-disk data for future tooling. Return types use tuple[...] per project immutability conventions.

### DIFF
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -225,7 +225,6 @@ def run_command(
                     "elapsed_s": round(get_elapsed(), 3),
                     "output_format": str(output_format),
                 }
-
                 span.add_event("run.done", execute_metrics)
                 rl.log_event("run.done", execute_metrics)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,8 +148,6 @@ dev = [
     "xgboost>=1.6.1",
     "quickgrove>=0.1.2",
     "matplotlib>=3.8.0",
-    "flake8>=7.3.0",
-    "flake8-class-attributes-order>=0.3.0",
 ]
 test = [
     "pytest-codspeed<4.3.1,>=3.0.0",
--- a/uv.lock
+++ b/uv.lock
@@ -1253,33 +1253,6 @@ wheels = [
 ]
 
 [[package]]
-name = "flake8"
-version = "7.3.0"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "mccabe" },
-    { name = "pycodestyle" },
-    { name = "pyflakes" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/9b/af/fbfe3c4b5a657d79e5c47a2827a362f9e1b763336a52f926126aa6dc7123/flake8-7.3.0.tar.gz", hash = "sha256:fe044858146b9fc69b551a4b490d69cf960fcb78ad1edcb84e7fbb1b4a8e3872", size = 48326, upload-time = "2025-06-20T19:31:35.838Z" }
-wheels = [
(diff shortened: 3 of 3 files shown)


python/xorq/cli.py
```diff
@@ -203,51 +187,67 @@ def run_command(
...
+
+            span.set_status(StatusCode.OK)
+
+            span_ctx = span.get_span_context()
+            otel_trace_id = (
+                format(span_ctx.trace_id, "032x")
+                if span_ctx and span_ctx.is_valid
+                else None
+            )
+            rl.finalize(status="ok", otel_trace_id=otel_trace_id)
```
dlovell: in the event of an exception, does finalize occur? if not, should it?

python/xorq/cli.py
```diff
@@ -203,51 +187,67 @@ def run_command(
...
+        with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
+            rl.log_event("run.start", **dict(run_params))
 
-        if limit is not None:
-            expr = expr.limit(limit)
+            with timed() as get_elapsed:
+                expr = load_expr(expr_path, cache_dir=cache_dir)
+            load_metrics = {"elapsed_s": round(get_elapsed(), 3)}
+            span.add_event("run.expr_loaded", load_metrics)
+            rl.log_event("run.expr_loaded", **load_metrics)
```
dlovell: should we modify `log_event` to take a dictionary like `span.add_event` rather than kwargs?

python/xorq/cli.py
```diff
@@ -203,51 +187,67 @@ def run_command(
...
-        with timed(span, logger, "run.expr_loaded"):
-            expr = load_expr(expr_path, cache_dir=cache_dir)
+        with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
+            rl.log_event("run.start", **dict(run_params))
 
-        if limit is not None:
-            expr = expr.limit(limit)
+            with timed() as get_elapsed:
+                expr = load_expr(expr_path, cache_dir=cache_dir)
+            load_metrics = {"elapsed_s": round(get_elapsed(), 3)}
```
dlovell: we shouldn't expect that the contextmanager's yield value is valid outside of its context

dlovell: for legibility, keep span/log events inside of the context that times them

python/xorq/cli.py
dlovell: perhaps either the ctor or finalize should do the work to get `otel_trace_id`

python/xorq/common/utils/logging_utils.py
```diff
@@ -95,3 +110,255 @@ def get_print_logger():
...
+
+    run_dir: Path = field(validator=instance_of(Path))
+    params_tuple = field(validator=instance_of(tuple), default=())
+
+    @property
+    def run_id(self) -> str:
+        return self.run_dir.name
+
+    _started_at = field(init=False)
+    _fh = field(init=False)
```
dlovell, line 153: all fields should be at the top of the class, followed by `__attrs_post_init__` if it exists

python/xorq/common/utils/logging_utils.py
```diff
@@ -95,3 +110,255 @@ def get_print_logger():
...
+
+    @staticmethod
+    def _get_otel_trace_id(span_ctx: SpanContext) -> str | None:
+
+        otel_trace_id = (
+            format(span_ctx.trace_id, "032x")
+            if span_ctx and span_ctx.is_valid
+            else None
+        )
+        return otel_trace_id
```
dlovell: we should put `classmethod`s and `staticmethod`s at the bottom of the class unless there's a readability argument for having them close to their use

PR #1655, opened 2026-03-01
feat: add LazyBackend

mesejo:


python/xorq/vendor/ibis/backends/profiles.py
```diff
@@ -198,12 +199,24 @@ def hash_name(self):
...
-    def get_con(self, **kwargs):
-        """Create a connection using this profile's parameters."""
+    def _get_connect(self, **kwargs):
         _kwargs = dict(self.kwargs_tuple) | kwargs
         connect = getattr(load_backend(self.con_name), "connect")
-        con = connect(**_kwargs)
-        return con
+        return connect, _kwargs
+
+    def get_con(self, **kwargs):
```
hussainsultan: `get_con` should take lazy as kwarg with default = false.

PR #1660, opened 2026-03-02
fix(ml): avoid eager model execution in score_expr and Scorer.from_spec

ghoersti:
fix(ml): avoid eager model execution in score_expr

  score_expr() and _get_default_scorer() accessed self.predict_step.model, which triggers deferred_model.execute() — eagerly fitting sklearn models during expression construction. This made deferred_cross_val_score spend ~5s per pipeline before .execute() was ever called, fitting all k fold
  models upfront.

  Changes

  - pipeline_lib.py: Pass self.predict_step.instance (unfitted) instead of .model (fitted, requires execution) in score_expr() and _get_default_scorer()
  - metrics.py: Docstring updates — _default_scorer_for_model and Scorer.from_spec only inspect the class hierarchy, so an unfitted instance is sufficient
  - test_cross_validation.py: Strengthened lazy-construction tests to patch Expr.execute at the base class level, catching .execute() on any expression during construction

### DIFF
--- a/python/xorq/expr/ml/tests/test_cross_validation.py
+++ b/python/xorq/expr/ml/tests/test_cross_validation.py
@@ -324,9 +324,7 @@ def test_splitter_produces_correct_folds(
             id="RepeatedKFold-accuracy",
         ),
         param(
-            lambda: RepeatedStratifiedKFold(
-                n_splits=3, n_repeats=2, random_state=42
-            ),
+            lambda: RepeatedStratifiedKFold(n_splits=3, n_repeats=2, random_state=42),
             "f1",
             id="RepeatedStratifiedKFold-f1",
         ),
@@ -336,9 +334,7 @@ def test_splitter_produces_correct_folds(
             id="ShuffleSplit-accuracy",
         ),
         param(
-            lambda: StratifiedShuffleSplit(
-                n_splits=4, test_size=0.25, random_state=42
-            ),
+            lambda: StratifiedShuffleSplit(n_splits=4, test_size=0.25, random_state=42),
             "accuracy",
             id="StratifiedShuffleSplit-accuracy",
         ),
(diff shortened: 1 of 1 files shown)


python/xorq/expr/ml/metrics.py
```diff
@@ -105,9 +108,10 @@ def from_spec(cls, scorer, model=None):
         ----------
         scorer : str, callable, _BaseScorer, Scorer, or None
             The scorer specification to normalize.
-        model : estimator, optional
-            A fitted sklearn estimator, used to determine the default scorer
-            when scorer is None.
+        model : estimator or None, optional
```
dlovell: is there anywhere else in the codebase where we say "or None" and "optional"?

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -982,7 +982,7 @@ def score_expr(self, expr, scorer=None, **kwargs):
         """
         from xorq.expr.ml.metrics import Scorer, deferred_sklearn_metric
 
-        s = Scorer.from_spec(scorer, model=self.predict_step.model)
+        s = Scorer.from_spec(scorer, model=self.predict_step.instance)
```
dlovell: the only time `model` is used is when `scorer` is `None`, so we should modify `Scorer` to have a separate path, perhaps `Scorer.from_model`

python/xorq/expr/ml/tests/test_cross_validation.py
```diff
@@ -429,38 +429,41 @@ def test_regressor_kfold_matches_sklearn(self, regression_data, regressor_pipeli
...
 
     def test_sklearn_splitter_no_execute_during_construction(
         self,
         classification_data,
         classifier_pipeline,
     ):
-        """Building a CrossValScore with an sklearn splitter must NOT call .execute().
+        """Building a CrossValScore with an sklearn splitter must NOT call
+        Expr.execute() anywhere in the expression graph."""
+        from unittest.mock import MagicMock, patch
```
dlovell: let's standardize on using `monkeypatch`

python/xorq/expr/ml/metrics.py
ghoersti:
```
  - cross_validation.py:367 — scoring : str, callable, _BaseScorer, Scorer, or None, optional
  - cross_validation.py:371 — random_seed : int or None, optional
  - cross_validation.py:379 — order_by : str, tuple of str, or None, optional
  - metrics.py:111 — model : estimator or None, optional
```

PR #1668, opened 2026-03-03
fix(parse): handle `string(N)` length qualifier from Postgres `VARCHAR` columns

gmurro:
## Problem

When [boring-semantic-layer](https://github.com/boringdata/boring-semantic-layer) uses a **Postgres** backend with a model that has **joins**, every query crashes with:

```
SignatureValidationError: '!string(50)' can't be coerced to DataType
```

The root cause is in this parser. Postgres represents `VARCHAR(N)` columns as
`ibis.String(length=N)`, which serialises via `str(dtype)` to `"string(50)"`.
The parser already handles `varchar(N)` correctly (strips the length → `dt.string`),
but `string(N)` hit the bare `primitive` rule that has no support for parameters,
causing a `ParseError` and ultimately the `SignatureValidationError` upstream.

Other backends, for example DuckDB, are unaffected because return `String(length=None)` → `"string"`.

## Change

`"string"` is moved out of the parameterless `primitive` block and added to the
`varchar_or_char` rule, which already accepts an optional `(N)` length qualifier
and discards it. Two lines changed:

```python
# before
primitive = (
    ...
    spaceless_string(..., "string", ...).map(...)   # no (N) support
    ...
)
varchar_or_char = spaceless_string("varchar", "char")...

# after
primitive = (
    ...
    spaceless_string(...).map(...)   # "string" removed
    ...
)
varchar_or_char = spaceless_string("varchar", "char", "string")...  # added
```

Behaviour is identical for bare `"string"` and `"!string"` (the optional `(N)`
is simply absent). `"string(50)"` now parses to `dt.string` exactly like
`"varchar(50)"` always did.

## Related

Reported in [boringdata/boring-semantic-layer#183](https://github.com/boringdata/boring-semantic-layer/issues/183)

### DIFF
--- a/python/xorq/vendor/ibis/tests/test_dtype_parse.py
+++ b/python/xorq/vendor/ibis/tests/test_dtype_parse.py
@@ -50,8 +50,7 @@ from xorq.vendor.ibis.expr.datatypes.parse import parse
         ("varchar(100)", dt.string),
         ("!varchar(100)", dt.String(nullable=False)),
         ("char(10)", dt.string),
-        # string(N) — Postgres serialises VARCHAR(N) as string(N) via str(dtype);
-        # the length qualifier carries no semantic meaning and must be stripped
+        # string(N) — length qualifier carries no semantic meaning and is stripped
         ("string(50)", dt.string),
         ("!string(50)", dt.String(nullable=False)),
         ("string(255)", dt.string),


mesejo:
Thanks for the contribution @gmurro. Could you add a test to prevent a future regression?

mesejo (approved):

PR #1669, opened 2026-03-03
fix(ml): better wide expr performance

dlovell:


python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -780,6 +780,9 @@ def fit(self, expr, features=None, target=None, cache=None):
         features = features or tuple(col for col in expr.columns if col != target)
         fitted_steps = ()
         transformed = expr
+        # During fit, other (non-feature) columns are only needed if a predict
```
ghoersti, line 783: @dlovell 
I think that this comment would be more useful as 
```bash
 # During fit, retain all non-feature columns if there's a predict step,
 # since the predictor will need the target for training.
```
it also will carry over other non feature columns other than target.

python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -905,15 +911,19 @@ def invoke_predict_method(self, expr, tag_name, tag_key, *, methodname, name=Non
...
-            .pipe(do_into_backend)
-            .tag(
-                tag_name,
-                **{tag_key: tuple(self.predict_step.tag_kwargs.items())},
-            )
+        predicted = method(transformed, name=name)
+        # Array-typed UDF outputs (e.g. predict_proba returning Array[float64]) cause
+        # issues in DataFusion when referenced multiple times in metric computations;
+        # materialization avoids stream exhaustion on second reference.
+        deferred_method = getattr(self.predict_step, f"deferred_{methodname}", None)
```
ghoersti, line 918: @dan Do you think something like this would be more useful?   instead of `getattr` on f string .
```python
  match methodname:
      case ResponseMethod.PREDICT_PROBA:
          deferred_method = self.predict_step.deferred_predict_proba
      case ResponseMethod.DECISION_FUNCTION:
          deferred_method = self.predict_step.deferred_decision_function
      case "feature_importances":
          deferred_method = self.predict_step.deferred_feature_importances
      case _:
          deferred_method = None
```

PR #1670, opened 2026-03-03
fix(ml): select fusion

dlovell:


python/xorq/expr/ml/pipeline_lib.py
```diff
@@ -544,11 +575,39 @@ def tag_kwargs(self):
...
         col = self.transform_raw(expr)
-        others = self.get_others(expr) if retain_others else ()
-        expr = expr.select(*others, col) if others else col.as_table()
-        return self.structer.maybe_unpack(expr, col.get_name()).tag(
-            **self.tag_kwargs,
-        )
+        needs_unpack = not self.structer.is_kv_encoded
+        parent = expr.op()
+        col_op = col.op()
+        if isinstance(col_op, ibis_ops.Alias):
```
ghoersti, line 583: @dlovell  do you think this would be cleaner ? 
```python
  def transform(self, expr, retain_others=True):
      col = self.transform_raw(expr)
      needs_unpack = not self.structer.is_kv_encoded
      parent = expr.op()
      col_op = col.op()

      # Unwrap alias to get the actual operation
      if isinstance(col_op, ibis_ops.Alias):
          col_op = col_op.arg

      # Build the set of columns to keep
      if needs_unpack:
          # Fuse add-col + unpack into one Project, avoiding two Table.select /
          # from_targets calls on wide tables. Builds struct field columns
          # directly from col_op without an intermediate "to_unpack" column.
          struct_fields = {
              fname: ibis_ops.StructField(col_op, fname) for fname in col.names
          }
          keep = struct_fields
      else:
          # Scalar output: just the transformed column
          keep = {col.get_name(): col_op}

      # Add non-feature columns if requested
      if retain_others:
          others = {
              n: ibis_ops.Field(parent, n)
              for n in expr.columns
              if n not in self.features
          }
          keep = others | keep  # merge with others first, then transformed

      return ibis_ops.Project(parent, keep).to_expr().tag(**self.tag_kwargs)
```

PR #1671, opened 2026-03-03
fix(ibis_yaml): deduplicate YAML output and simplify serialization

mesejo:
- Deduplicate YAML output by introducing node_refs: repeated nodes
  (relations, schemas, dtypes) are emitted once and referenced by ID
  elsewhere, reducing verbosity in complex expressions
- Register JoinReference as a node_ref so join relations are also
  deduplicated in the YAML representation
- Use dtype_refs inside schema definitions to further reduce repetition
  of shared data types
- Simplify primitive serialization throughout translate.py and common.py
- Update test snapshots and add compiler tests covering deduplication

### DIFF
--- a/python/xorq/ibis_yaml/tests/test_relations.py
+++ b/python/xorq/ibis_yaml/tests/test_relations.py
@@ -42,7 +42,6 @@ def test_projection(compiler, t):
 def test_aggregation(compiler, t):
     expr = t.group_by("a").aggregate(avg_c=t.c.mean())
     yaml_dict = compiler.to_yaml(expr)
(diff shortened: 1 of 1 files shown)


python/xorq/ibis_yaml/tests/test_relations.py
```diff
@@ -44,6 +44,9 @@ def test_aggregation(compiler, t):
     yaml_dict = compiler.to_yaml(expr)
     expression = yaml_dict["expression"]
```
dlovell: this line is no longer necessary

PR #1681, opened 2026-03-05
feat(expr): add HashingTag node for hash-contributing metadata tags

hussainsultan:


python/xorq/vendor/ibis/expr/types/relations.py
```diff
@@ -3437,6 +3436,16 @@ def tag(self, tag, **kwargs):
         )
         return op.to_expr()
 
+    def tag(self, tag, **kwargs):
+        from xorq.expr.relations import Tag
```
mesejo, line 3440: I think we need to mark these imports as # noqa: PLC0415

PR #1682, opened 2026-03-05
feat(catalog): add Expr kind (XOR-244)

mesejo:


python/xorq/catalog/catalog.py
```diff
@@ -460,6 +462,25 @@ def expr(self):
 
         return load_expr_from_tgz(self.catalog_path)
 
+    @cached_property
+    def kind(self) -> str:
```
mesejo: In this function, I'm not sure if we should raise an Exception is the type is missing or simply return the default value (or a different default value like "?")

dlovell: we should raise
and `kind` should return an `ExprKind`, not a `str`

python/xorq/catalog/catalog.py
```diff
@@ -460,6 +462,25 @@ def expr(self):
...
+        data = self._read_tgz_yaml(DumpFiles.expr)
+        if not isinstance(data, dict):
+            return default_value
+        return data.get("kind", default_value)
+
+    @cached_property
+    def backends(self) -> tuple:
+        data = self._read_tgz_yaml(DumpFiles.profiles)
+        if not isinstance(data, dict):
+            return ()
```
dlovell: under what conditions would this be a valid file/build?
if its not a valid file/build we should raise

python/xorq/catalog/catalog.py
```diff
@@ -460,6 +462,25 @@ def expr(self):
...
+        return tuple(
+            pdata.get("con_name", "?")
```
dlovell: similarly here: under what conditions would this be a valid connection?

python/xorq/catalog/cli.py
```diff
@@ -234,14 +234,23 @@ def remove_alias(ctx, aliases, sync):
...
+            entries = catalog.catalog_entries
+            if not entries:
+                click.echo("No entries.")
+                return
+            for entry in entries:
+                click.echo(f"{entry.name}\t{entry.kind}")
+        else:
+            names = catalog.list() or ("No entries.",)
+            for name in names:
+                click.echo(name)
```
dlovell: unless enumerating `CatalogEntry` objects is costly, we should aim for parsimony and only vary the `click.echo` invocation

python/xorq/catalog/catalog.py
```diff
@@ -492,6 +516,22 @@ def assert_consistency(self):
     def exists(self):
         return all(self._exists_components.values())
 
+    def _read_tgz_yaml(self, filename):
+        with tarfile.open(self.catalog_path, "r:gz") as tf:
+            f = tf.extractfile(f"{self.name}/{filename}")
+            if f is None:
+                return None
```
dlovell: under what conditions would this path happen?

python/xorq/ibis_yaml/enums.py
```diff
@@ -0,0 +1,31 @@
+try:
+    from enum import StrEnum
+except ImportError:
+    from strenum import StrEnum
+
+
+class DumpFiles(StrEnum):
+    deferred_reads = "deferred_reads.yaml"
+    expr = "expr.yaml"
+    entry = "entry.json"
```
dlovell: what is the tradeoff between adding `DumpFiles.entry` and using `DumpFiles.metadata` for the `ExprKind` information?

mesejo: In my view, `Dumpfiles.metadata` contains information about the system that built the Expr (metadata), the Dumpfiles.entry contains _cached_ information about the Expr, not just `ExprKind` but also `schema_in` and `schema_out`, and any other field we consider important in the future.

 I don't think these two pieces of information belong in the same file.  

I understand adding another file increases complexity, but we could write the metadata to `Dumpfiles.expr`.

dlovell: understood.

let's pick a more descriptive name than `entry`

python/xorq/catalog/catalog.py
```diff
@@ -492,6 +516,24 @@ def assert_consistency(self):
...
+    def _read_tgz_yaml(self, filename):
+        with tarfile.open(self.catalog_path, "r:gz") as tf:
+            f = tf.extractfile(f"{self.name}/{filename}")
+            if f is None:
+                # https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractfile
+                raise ValueError(f"{filename} is not a regular file or a link")
+            return yaml.safe_load(f.read())
+
+    def _read_tgz_json(self, filename):
+        import json  # noqa: PLC0415
```
dlovell: why is this import deferred?

python/xorq/catalog/cli.py
```diff
@@ -236,14 +236,18 @@ def remove_alias(ctx, aliases, sync):
...
         catalog = ctx.obj.make_catalog(init=False)
-        names = catalog.list() or ("No entries.",)
-        for name in names:
-            click.echo(name)
+        entries = catalog.catalog_entries
+        if not entries:
+            click.echo("No entries.")
+            return
+        for entry in entries:
+            click.echo(f"{entry.name}\t{entry.kind}" if kind else entry.name)
```
dlovell: let's use `match entries := catalog.entries:`

python/xorq/ibis_yaml/compiler.py
```diff
@@ -448,11 +430,38 @@ def _make_metadata() -> str:
...
+                "schema_in": {
+                    name: str(dtype) for name, dtype in unbound_node.schema.items()
+                }
+            }
+            if unbound_node
+            else {}
+        )
+
+    def _prepare_entry_file(self, expr):
+        path = self.artifact_store.get_path(DumpFiles.expr_metadata)
```
dlovell: lets make these names consistent with the target file

python/xorq/ibis_yaml/compiler.py
```diff
@@ -562,9 +572,15 @@ def expr_hash(self):
     def artifact_store(self):
         return ArtifactStore(self.expr_path)
 
-    def load_expr(self):
+    def load_expr(self, raise_on_unbound: bool = False):
```
dlovell: why does `raise_on_unbound` default to `False`?

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -674,6 +677,43 @@ def ls(self):
...
+    def kind(self) -> ExprKind:
+        return ExprKind.UnboundExpr if self._unbound_node else ExprKind.Expr
+
+    @cached_property
+    def schema_in(self):
+        return toolz.valmap(str, node.schema) if (node := self._unbound_node) else None
+
+    @cached_property
+    def schema_out(self):
+        return toolz.valmap(str, self.expr.as_table().schema())
```
dlovell: if you're going to have properties called, `schema_in` and `schema_out` and a `to_dict` method then `schema_in` and `schema_out` should be `optional[schema]` and `to_dict` should be where conversion happens

PR #1684, opened 2026-03-05
fix(lint): remove functools.cache from methods

mesejo:
- Replace `@property @functools.cache` combos with `@functools.cached_property` to fix B019 violations and eliminate memory leaks from instance-keyed caches
- Convert standalone `@functools.cache` methods (`copy_sdist`) to `@functools.cached_property` and update call sites accordingly
- Replace class-level `cached_property` aliases (`popened = _uv_build_popened`)  with explicit `@property` delegates (Python 3.13 disallows reusing the same  `cached_property` object under two names)
- Add `# noqa: B019` for `make_deferred_other`, which takes extra arguments and cannot be converted to a `cached_property`

Fix a latent bug in `SdistBuilder.maybe_packager`: the field had `converter=toolz.curried.excepts(Exception, Path)`, which silently converted a `Sdister` object to `None` because `Path(sdister_instance)` raises `TypeError`. This caused the `Sdister` to be garbage-collected immediately after `SdistBuilder.from_script_path` returned, cleaning up its
`TemporaryDirectory` and deleting the sdist file that `SdistBuilder.sdist_path` pointed to. Subsequent access to `sdist_path` in `_uv_tool_run_xorq_build` then failed with `FileNotFoundError`. The fix removes the broken converter, so `maybe_packager` holds the `Sdister` directly, keeping it alive for the lifetime of the `SdistBuilder`.

Proof that `@frozen` (attrs) works with `@cached_property`:

```python
from attrs import frozen
from functools import cached_property

@Frozen
class Circle:
    radius: float

    @cached_property
    def area(self):
        print("computing...")
        return 3.14159 * self.radius ** 2

c = Circle(radius=5.0)
print(c.area)  # computing... → 78.53975
print(c.area)  # cached → 78.53975
```

### DIFF
--- a/python/xorq/expr/ml/sklearn_utils.py
+++ b/python/xorq/expr/ml/sklearn_utils.py
@@ -18,7 +18,7 @@ Example: Pipeline([("prep", ColumnTransformer([("num", ..., cols), ...]))])
 
 from __future__ import annotations
 
-from functools import cache
+from functools import cached_property
 
 from attr import field, frozen
 from sklearn.base import clone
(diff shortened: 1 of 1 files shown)


python/xorq/common/utils/trace_utils.py
```diff
@@ -410,7 +408,7 @@ def get_duration_delta(self, parent_span_id, lossless_leafs=True):
         else:
             return 0
 
-    @functools.cache
+    @functools.cached_property
```
dlovell, line 411: the informal convention we have is that `def get_$x` is a method and `def $x` is a property. let's make this `depths`

PR #1701, opened 2026-03-09
fix: decompile() crash on zero-arg window functions

### LINKED ISSUE #1700: decompile() crashes on zero-arg window functions (RowNumber, Rank, etc.)
## Bug

`decompile()` raises `ValueError: not enough values to unpack (expected at least 1, got 0)` when the expression contains any zero-argument window function (`row_number()`, `rank()`, `dense_rank()`, `percent_rank()`, `cume_dist()`).

## Minimal reproduction

```python
import xorq.vendor.ibis as ibis
from xorq.vendor.ibis.expr.decompile import decompile

t = ibis.table(name='t', schema={'x': 'int64', 'y': 'string'})
expr = t.mutate(row_num=ibis.row_number())
decompile(expr)  # ValueError
```

## Traceback

```
File "xorq/vendor/ibis/expr/decompile.py", line 465, in decompile
    node.map(fn)
File "xorq/vendor/ibis/common/graph.py", line 320, in map
    results[node] = fn(node, results, **kwargs)
File "xorq/vendor/ibis/expr/decompile.py", line 456, in fn
    code = translate(node, *args, **kwargs)
File "xorq/vendor/ibis/expr/decompile.py", line 96, in value
    (_, this), *kwargs = kwargs
ValueError: not enough values to unpack (expected at least 1, got 0)
```

## Cause

In `decompile.py` line 88–96, the `value()` handler assumes that if `args` is empty, `kwargs` must contain at least one entry (the `this` reference). But zero-arg analytics like `RowNumber` have neither positional args nor kwargs after the `None`-filtering at line 91, so the destructuring at line 96 fails:

```python
@translate.register(ops.Value)
def value(op, *args, **kwargs):
    method = _get_method_name(op)
    kwargs = [(k, v) for k, v in kwargs.items() if v is not None]

    if args:
        this, *args = args
    else:
        (_, this), *kwargs = kwargs  # <-- crashes when kwargs is empty
```

## Affected functions

All zero-arg analytic/window functions crash:

| Function | Crashes |
|----------|---------|
| `ibis.row_number()` | Yes |
| `ibis.rank()` | Yes |
| `ibis.dense_rank()` | Yes |
| `ibis.percent_rank()` | Yes |
| `ibis.cume_dist()` | Yes |

## Context

Discovered while building a Code tab in xorq-web that decompiles catalog expressions for display. Any expression built with `xorq_run` that uses window functions (e.g., knapsack optimizers ranking players) fails to render.

paddymul:
## Summary

- Adds failing tests for #1700 — `decompile()` crashes with `ValueError: not enough values to unpack` on expressions containing zero-argument analytic functions (`row_number()`, `rank()`, `dense_rank()`, `percent_rank()`, `cume_dist()`)
- **Tests only — no fix yet.** CI should show 7 failures confirming the bug.
- Fix will follow in a subsequent commit once CI confirms the failures.

## Test plan

- [x] 7 new tests in `python/xorq/vendor/ibis/expr/test_decompile.py`
- [x] All 7 fail with `ValueError` on current main
- [ ] Fix commit will make all 7 pass

Closes #1700

### DIFF
--- a/python/xorq/vendor/ibis/expr/decompile.py
+++ b/python/xorq/vendor/ibis/expr/decompile.py
@@ -92,8 +92,11 @@ def value(op, *args, **kwargs):
 
     if args:
         this, *args = args
-    else:
+    elif kwargs:
         (_, this), *kwargs = kwargs
+    else:
+        # Zero-arg analytic functions (row_number, rank, dense_rank, etc.)
+        return f"ibis.{method}()"
 
     # if there is a single keyword argument prefer to pass that as positional
     if not args and len(kwargs) == 1:
--- a/python/xorq/vendor/ibis/expr/decompile.py
+++ b/python/xorq/vendor/ibis/expr/decompile.py
@@ -90,7 +90,12 @@ def value(op, *args, **kwargs):
     method = _get_method_name(op)
     kwargs = [(k, v) for k, v in kwargs.items() if v is not None]
 
-    if args:
+    if args and kwargs:
+        raise NotImplementedError(
+            f"decompile does not support ops with both positional and keyword "
+            f"arguments: {type(op).__name__} args={args} kwargs={kwargs}"
+        )
+    elif args:
         this, *args = args
     elif kwargs:
         (_, this), *kwargs = kwargs
--- a/python/xorq/vendor/ibis/expr/test_decompile.py
+++ b/python/xorq/vendor/ibis/expr/test_decompile.py
@@ -1,0 +1,58 @@
+"""Tests for the expression decompiler."""
+
+import pytest
+
+import xorq.vendor.ibis as ibis
+from xorq.vendor.ibis.expr.decompile import decompile
+
+
+class TestDecompileZeroArgAnalytics:
+    """decompile() should handle zero-argument window/analytic functions.
+
+    Regression test for https://github.com/xorq-labs/xorq/issues/1700
+    """
+
+    @pytest.fixture
+    def table(self):
+        return ibis.table(name="t", schema={"x": "int64", "y": "string"})
+
+    def test_row_number(self, table):
+        expr = table.mutate(rn=ibis.row_number())
+        code = decompile(expr)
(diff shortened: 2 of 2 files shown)


mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: High
**Location**: `python/xorq/vendor/ibis/expr/decompile.py` — the `if args and kwargs: raise NotImplementedError` branch
**Problem**: Regression. The original `if args` branch already handled ops that have both positional and keyword args — it extracted `this` from `args`, then the remaining `args` and `kwargs` were both used in the final `f"{this}.{method}({parameters})"` return. The new guard raises `NotImplementedError` for any op with both, e.g. a `Quantile(arg, quantile=0.5)` call that previously decompiled to `arg.quantile(0.5)` would now crash. The guard is also logically prior to the reassignment of `kwargs` to a list (line 91), so the error message leaks the pre-filtered list form, not the original dict.
**Fix**: Drop the `args and kwargs` guard. The minimal correct fix is just to add the zero-arg case before the existing branches:

```python
if not args and not kwargs:
    return f"ibis.{method}()"
elif args:
    this, *args = args
else:
    (_, this), *kwargs = kwargs
```

---

**Severity**: Low
**Location**: `python/xorq/vendor/ibis/expr/test_decompile.py:20–57`
**Problem**: Test assertions only check substring presence (`assert "row_number()" in code`). They don't verify the decompiled string is syntactically valid Python or that the full expression structure is correct. A fix that returns `"ibis.row_number()"` embedded in garbage would still pass.
**Fix**: At minimum, add `exec`/`compile` of the decompiled code, or assert the full expected string for one representative case.

---

## Summary

Fixes a crash in `decompile()` for zero-argument analytic functions, but introduces a regression by raising `NotImplementedError` for any op that has both positional and keyword arguments, which previously worked correctly.

</details>

---
*Review type: default | Agent: claude-code | Job: 23*

PR #1705, opened 2026-03-10
fix: add decompile() handlers for custom xorq relation types

### LINKED ISSUE #1702: decompile() has no handler for FlightUDXF (and other custom xorq relation types)
## Summary

`decompile()` in `vendor/ibis/expr/decompile.py` uses singledispatch to translate Ibis operations back to Python source. It has handlers for standard ops (`Project`, `Filter`, `Aggregate`, etc.) but **none** for xorq's custom relation types: `FlightUDXF`, `RemoteTable`, `CachedNode`, and `Read`. Any expression containing these nodes will fail or produce incomplete output when decompiled.

## Root Cause

Two gaps:

1. **No `@translate.register()` handlers** for the custom ops defined in `expr/relations.py`. The singledispatch falls through when it encounters these types.

2. **For `FlightUDXF` specifically**, the `process_df` callable is serialized as a cloudpickle blob in `ibis_yaml/udf.py` (lines 347-364). The original Python source is not preserved at serialization time, so even if a handler existed, there's no source to reconstruct. `inspect.getsource()` at `flight_udxf()` call time could capture it, but this isn't done today.

## Steps to Reproduce

```python
import xorq as xo

def my_transform(df):
    return df.assign(new_col=df["x"] * 2)

udxf = xo.expr.relations.flight_udxf(
    process_df=my_transform,
    maybe_schema_in=input_expr.schema(),
    maybe_schema_out=output_schema,
    name="MyTransform",
)
expr = input_expr.pipe(udxf)

from xorq.vendor.ibis.expr.decompile import decompile
decompile(expr)  # fails or produces incomplete output
```

## Proposed Fix

1. Register `@translate` handlers in `decompile.py` for `FlightUDXF`, `RemoteTable`, `CachedNode`, and `Read`.
2. For `FlightUDXF`: capture `process_df` source via `inspect.getsource()` at construction time and store it on the node, so the decompile handler can emit it.

## Relevant Files

- `python/xorq/expr/relations.py` — custom op definitions (FlightUDXF at lines 342-420)
- `python/xorq/ibis_yaml/udf.py` — YAML serialization, cloudpickle of callables (lines 346-392)
- `python/xorq/vendor/ibis/expr/decompile.py` — singledispatch decompile (missing handlers)

paddymul:
## Summary

- Adds failing tests for #1702 — `decompile()` has no `@translate.register()` handlers for `CachedNode`, `RemoteTable`, `FlightUDXF`, or `Read`
- All four types fall back to the generic `DatabaseTable` handler, which emits a plain `ibis.table(name=..., schema=...)` and silently discards `parent`, `remote_expr`, `input_expr`, `read_kwargs`, etc.
- **Tests only — no fix yet.** CI should show 8 failures confirming the bug.

## Test plan

- [x] 8 new tests in `python/xorq/vendor/ibis/expr/test_decompile.py`
  - 4 dispatch registration checks (handler is not the DatabaseTable fallback)
  - 4 output checks (decompiled code mentions cache/remote/read_parquet/udxf)
- [x] All 8 fail on current main
- [ ] Fix commit will register handlers and make all 8 pass

Closes #1702

### DIFF
--- a/python/xorq/vendor/ibis/expr/test_decompile.py
+++ b/python/xorq/vendor/ibis/expr/test_decompile.py
@@ -1,0 +1,179 @@
+"""Tests for the expression decompiler."""
+
+import pytest
+
+import xorq.vendor.ibis as ibis
+import xorq.vendor.ibis.expr.operations as ops
+from xorq.vendor.ibis.expr.decompile import decompile, translate
+
+
+class TestDecompileZeroArgAnalytics:
+    """decompile() should handle zero-argument window/analytic functions.
+
+    Regression test for https://github.com/xorq-labs/xorq/issues/1700
+    """
+
+    @pytest.fixture
+    def table(self):
+        return ibis.table(name="t", schema={"x": "int64", "y": "string"})
+
+    def test_row_number(self, table):
+        expr = table.mutate(rn=ibis.row_number())
+        code = decompile(expr)
+        assert "row_number()" in code
+
+    def test_dense_rank(self, table):
+        expr = table.mutate(rn=ibis.dense_rank())
+        code = decompile(expr)
+        assert "dense_rank()" in code
+
+    def test_rank(self, table):
+        expr = table.mutate(rn=ibis.rank())
+        code = decompile(expr)
+        assert "rank()" in code
+
+    def test_percent_rank(self, table):
+        expr = table.mutate(rn=ibis.percent_rank())
+        code = decompile(expr)
+        assert "percent_rank()" in code
+
+    def test_cume_dist(self, table):
+        expr = table.mutate(rn=ibis.cume_dist())
+        code = decompile(expr)
+        assert "cume_dist()" in code
+
+    def test_row_number_with_filter(self, table):
+        """Ensure decompile works when row_number appears in a filter."""
+        expr = table.filter(ibis.row_number() < 10)
+        code = decompile(expr)
+        assert "row_number()" in code
+
+    def test_row_number_combined_with_other_ops(self, table):
+        """Ensure decompile works with row_number mixed into larger expressions."""
+        expr = table.mutate(
+            rn=ibis.row_number(),
+            x_doubled=table.x * 2,
+        )
+        code = decompile(expr)
+        assert "row_number()" in code
+        assert "t.x" in code
+
+
+class TestDecompileCustomXorqRelations:
+    """decompile() should have handlers for custom xorq relation types.
+
+    Regression tests for https://github.com/xorq-labs/xorq/issues/1702
+
+    CachedNode, RemoteTable, FlightUDXF, and Read should not fall back to
+    the generic DatabaseTable handler, which discards their extra fields.
+    """
+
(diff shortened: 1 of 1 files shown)


mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: High
**Location**: `test_decompile.py` — entire `TestDecompileCustomXorqRelations` class
**Problem**: No `pytest.mark.xfail` on any of these tests. Commit message explicitly says "add failing tests" — these will unconditionally break CI as written. Red-phase TDD tests must either be marked `xfail` or live on a branch that never merges green.
**Fix**: Add `@pytest.mark.xfail(strict=True, reason="...")` to each test (or the class), or add `pytest.ini` / `pyproject.toml` `xfail_strict` config.

---

**Severity**: Medium
**Location**: `test_decompile.py:121` — `test_decompile_remote_table_mentions_remote`
**Problem**: `RemoteTable` field is typed `remote_expr: Expr` (ibis expression wrapper), but the test passes `remote_expr=t.op()` (a bare `Node`). ibis attrs validation will likely raise a `TypeError` before `decompile()` is ever called, so the test fails for the wrong reason and obscures the real dispatch gap it's meant to expose.
**Fix**: Pass the expression, not the op: `remote_expr=t` (not `t.op()`).

---

**Severity**: Medium
**Location**: `test_decompile.py:60-97` — dispatch handler tests
**Problem**: Tests couple to `functools.singledispatch` internals via `translate.dispatch()`. If the dispatch mechanism is changed (e.g., to `plum` or a registry dict), all four handler tests silently break. Also, `translate` is an internal symbol — importing it in tests creates a fragile coupling point.
**Fix**: Test behavior (does `decompile(node.to_expr())` produce correct output?) rather than dispatch registration. The behavior tests already exist; the dispatch tests add little value.

---

**Severity**: Low
**Location**: `test_decompile.py:144` — `normalize_method=lambda x: x`
**Problem**: Lambda is not picklable. If `Read` construction or `decompile()` triggers serialization (e.g., via ibis node hashing or any caching layer), the test will raise `PicklingError` instead of the intended `AssertionError`.
**Fix**: Use a named module-level function instead of a lambda.

---

**Severity**: Low
**Location**: `test_decompile.py:112, 130, 150, 167` — string assertions
**Problem**: `assert "cache" in code.lower()`, `assert "remote" in code.lower()`, etc. are too loose — any coincidental occurrence of those words (error message, variable name, comment in decompiled output) would make the tests pass falsely.
**Fix**: Assert specific API call patterns, e.g., `assert "cached(" in code` or `assert ".cache(" in code`.

---

## Summary

Adds failing regression tests for `decompile()` missing handlers for `CachedNode`, `RemoteTable`, `FlightUDXF`, and `Read`, plus zero-arg analytic functions, but the tests lack `xfail` markers (will break CI), one test passes a `Node` where an `Expr` is required (wrong failure mode), and the dispatch-level assertions are fragile implementation-coupling that duplicates the behavior tests.

</details>

---
*Review type: default | Agent: claude-code | Job: 22*

PR #1716, opened 2026-03-14
perf(packager): switch sdist format from tgz to zip

dlovell:
Zip archives support random access and in-place append, eliminating the gunzip→tar_append→gzip pipeline that TGZAppender required. This simplifies the code and improves sdist build/read performance.

- Add zip_utils.py with ZipProxy, ZipAppender, calc_zip_content_hexdigest, and tgz_to_zip converter (since uv build only outputs .tar.gz)
- Remove tar_utils.py (no remaining importers in common/utils)
- Update packager.py to use zip utilities throughout Sdister, SdistBuilder, SdistRunner, and helper functions
- Add ZipExtFile support to file_digest in dask_normalize_utils
- Update tests to work with zip-based sdists

### DIFF
--- a/python/xorq/common/utils/zip_utils.py
+++ b/python/xorq/common/utils/zip_utils.py
@@ -31,7 +31,7 @@ def try_path_relative_to(from_, to_):
 
 
 def uv_sdist_member_filter(name):
-    return not any(name.endswith(suffix) for suffix in uv_sdist_omit_suffixes)
+    return not name.endswith(uv_sdist_omit_suffixes)
 
 
 def get_root_dir(zip_path):
@@ -49,14 +49,14 @@ def calc_zip_content_hexdigest(path, member_filter=uv_sdist_member_filter):
     )
 
     with zipfile.ZipFile(path, "r") as zf:
-        dct = {
-            name: file_digest(zf.open(name), hashlib.md5)
+        names = sorted(
+            name
             for name in zf.namelist()
             if not name.endswith("/") and member_filter(name)
-        }
+        )
         md5 = hashlib.md5()
-        for _, value in sorted(dct.items()):
-            md5.update(value.encode("ascii"))
+        for name in names:
+            md5.update(file_digest(zf.open(name), hashlib.md5).encode("ascii"))
         return md5.hexdigest()
 
 


python/xorq/common/utils/zip_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        file_digest,
+    )
+
+    with zipfile.ZipFile(path, "r") as zf:
+        dct = {
+            name: file_digest(zf.open(name), hashlib.md5)
+            for name in zf.namelist()
+            if not name.endswith("/") and member_filter(name)
+        }
+        print(tuple(sorted(dct.items())))
```
mesejo: Is this print intentional?

python/xorq/common/utils/zip_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        md5 = hashlib.md5()
+        for _key, value in sorted(dct.items()):
+            md5.update(value.encode("ascii"))
+        return md5.hexdigest()
```
mesejo: Perhaps this could be rewritten as:
```python
        names = sorted(
            name
            for name in zf.namelist()
            if not name.endswith("/") and member_filter(name)
        )
        md5 = hashlib.md5()
        for name in names:
            md5.update(file_digest(zf.open(name), hashlib.md5).encode("ascii"))
```

python/xorq/common/utils/zip_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            shutil.copyfileobj(from_fh, to_fh)
+
+
+@toolz.curried.excepts(ValueError)
+def try_path_relative_to(from_, to_):
+    return Path(from_).relative_to(to_)
+
+
+def uv_sdist_member_filter(name):
+    return not any(name.endswith(suffix) for suffix in uv_sdist_omit_suffixes)
```
mesejo: This could be rewritten to:

```python
return not name.endswith(uv_sdist_omit_suffixes)
```

https://docs.python.org/3/library/stdtypes.html#str.endswith

python/xorq/common/utils/zip_utils.py
dlovell: nope, good catch

mesejo (approved):

PR #1717, opened 2026-03-14
perf(catalog): switch catalog and download_utils from tgz to zip

dlovell:
## Summary
- Extends the tgz-to-zip conversion from #1716 to cover the catalog subsystem and GitHub template downloads
- Replaces `catalog/tar_utils.py` with `catalog/zip_utils.py` (zip has O(1) append, random-access reads, simpler stdlib API)
- Switches GitHub archive downloads from `.tar.gz` to `.zip`
- Renames `REQUIRED_TGZ_NAMES` → `REQUIRED_ARCHIVE_NAMES`, `VALID_SUFFIXES`/`PREFERRED_SUFFIX` to `.zip` only

> **Depends on #1716** (`perf/sdister/use-zip`) — the first commit in this branch is from that PR.

## Test plan
- [x] `python -m pytest python/xorq/catalog/tests/ -x -q -m "not slow"` — 28 passed
- [x] `python -m pytest python/xorq/catalog/tests/test_cli.py -x -q -m "not slow"` — 63 passed
- [x] `python -m pytest python/xorq/tests/test_cli_run_alias.py -x -q -m "not slow"` — 10 passed
- [x] `python -m pytest python/xorq/common/utils/tests/test_io_utils.py -x -q` — 19 passed
- [ ] `python -m pytest python/xorq/ibis_yaml/tests/test_packager.py -x -q --snapshot-update` (slow, needs network)

### DIFF
--- a/python/xorq/catalog/tests/conftest.py
+++ b/python/xorq/catalog/tests/conftest.py
@@ -13,15 +13,11 @@ from xorq.catalog.catalog import (
     CATALOG_YAML_NAME,
     METADATA_APPEND,
     PREFERRED_SUFFIX,
-    VALID_SUFFIXES,
     Catalog,
     with_pure_suffix,
 )
 from xorq.catalog.constants import CatalogInfix
 from xorq.catalog.expr_utils import build_expr_context_zip
-
-
-(VALID_SUFFIX0,) = VALID_SUFFIXES
 
 
 def get_split_tree(repo):


python/xorq/catalog/tests/conftest.py
```diff
@@ -18,10 +18,10 @@
     with_pure_suffix,
 )
 from xorq.catalog.constants import CatalogInfix
-from xorq.catalog.expr_utils import build_expr_context_tgz
+from xorq.catalog.expr_utils import build_expr_context_zip
 
 
-VALID_SUFFIX0, VALID_SUFFIX1, *_ = VALID_SUFFIXES
+(VALID_SUFFIX0,) = VALID_SUFFIXES
```
mesejo: Is this variable used?

dlovell: no. should ruff have caught that and refactored it out?

mesejo: I would expect it to, but I ran it locally, and it passed

PR #1721, opened 2026-03-16
perf(catalog): use git annex

dlovell:


paddymul:
Can you get claude to write some docs.  

Can you add an architecture decision record about why git annex vs git lfs.  once again just ask calude to do it, give it two or three sentences of your reasoning, and see what it spits out.

specifically about how testing for this is supposed to work,  A one sentence definition of what minio is.

talk about the set_remote_config too.  I really like that idea, and I was just wondering about accidentally storing credentials in git, with build local run cloud, earlier today.  It's a good design but we should understand the intent and possibilities there,  it could be the source of security vulnerabilities

PR #1727, opened 2026-03-20
feat(expr): add ExprKind.Source to distinguish source from transformed expressions

hussainsultan:
## Summary
- Adds `ExprKind.Source` enum variant to distinguish bare source tables (e.g. `DatabaseTable`, `InMemoryTable`, `Read`, `CachedNode`) from transformed expressions
- Adds `.ls.kind` and `.ls.unwrapped` accessors on `LETSQLAccessor` for convenient source detection and Tag/HashingTag unwrapping
- Updates `ExprMetadata.kind` to return `Source` when the expression root is a source node

## Test plan
- [x] Unit tests for `ExprMetadata.kind` across source, expr, and unbound cases
- [x] Tests for `.ls.kind` and `.ls.unwrapped` accessors
- [x] Updated catalog and compiler tests to verify `ExprKind.Source` classification

### DIFF
--- a/python/xorq/vendor/ibis/expr/types/core.py
+++ b/python/xorq/vendor/ibis/expr/types/core.py
@@ -832,8 +832,7 @@ class LETSQLAccessor:
 
         root = self.op
         while isinstance(root, (Tag, HashingTag)):
-            parent = root.parent
-            root = parent.op() if isinstance(parent, Expr) else parent
+            root = root.parent
         return root
 
     @property
--- a/python/xorq/expr/relations.py
+++ b/python/xorq/expr/relations.py
@@ -116,7 +116,7 @@ def make_native_op(node):
 
 class Tag(ops.Relation):
     schema: Schema
-    parent: Any
+    parent: ops.Relation
     metadata: FrozenOrderedDict = FrozenOrderedDict()
     values = FrozenDict()
 


python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -807,6 +825,17 @@ def is_cached(self):
     def has_cached(self):
         return bool(self.cached_nodes)
 
+    @cached_property
+    def unwrapped(self):
+        """Unwrap Tag and HashingTag layers to get the underlying op."""
+        from xorq.expr.relations import HashingTag, Tag
+
+        root = self.op
+        while hasattr(root, "parent") and isinstance(root, (Tag, HashingTag)):
```
dlovell: could this be `isinstance(root, (Tag, HashingTag)):` without the `hasattr(root, "parent")`?
when is `parent` of a `Tag`/`HashingTag` not an `Expr`?

hussainsultan: yep, done.

dlovell: i propose making the following two changes

```
diff --git a/python/xorq/expr/relations.py b/python/xorq/expr/relations.py
index 7fd729c1..c91a191e 100644
--- a/python/xorq/expr/relations.py
+++ b/python/xorq/expr/relations.py
@@ -116,7 +116,7 @@ def make_native_op(node):

 class Tag(ops.Relation):
     schema: Schema
-    parent: Any
+    parent: ops.Relation
     metadata: FrozenOrderedDict = FrozenOrderedDict()
     values = FrozenDict()

diff --git a/python/xorq/vendor/ibis/expr/types/core.py b/python/xorq/vendor/ibis/expr/types/core.py
index b16b2be5..0c42010d 100644
--- a/python/xorq/vendor/ibis/expr/types/core.py
+++ b/python/xorq/vendor/ibis/expr/types/core.py
@@ -832,8 +832,7 @@ class LETSQLAccessor:

         root = self.op
         while isinstance(root, (Tag, HashingTag)):
-            parent = root.parent
-            root = parent.op() if isinstance(parent, Expr) else parent
+            root = root.parent
         return root

     @property
```

per claude
```
❯ can Tag.parent ever be anything other than ops.Relation?

  Searched for 3 patterns (ctrl+o to expand)

● The type annotation is Any, so let me check what's actually passed in practice.

  Searched for 1 pattern (ctrl+o to expand)

● Now let me check the .tag() method on expressions to see what it passes as parent.

  Searched for 2 patterns (ctrl+o to expand)

● In practice, yes, Tag.parent is always an ops.Relation. Here's why:

  - _make_tag (the only production constructor) passes self.op() as parent, where self is a Table expression — so .op() always returns an ops.Relation node.
  - The test in test_relations.py:77 also passes an ops.Relation (a memtable op).
  - The code in api.py:398-399 traverses node.parent expecting it to be another node (potentially another Tag or some other ops.Relation).

  The parent: Any annotation is just loose typing — it's not enforced, but all call sites pass an ops.Relation. You could safely tighten it to parent: ops.Relation.
```

PR #1729, opened 2026-03-20
feat(catalog): add composed expr kind and catalog run

hussainsultan:
## Summary
- Introduces `CatalogSource` relation (subclass of `RemoteTable`) that preserves catalog entry references and `kind` through serialization
- Adds `ExprKind.Composed` for expressions created by binding a source to unbound transforms
- New `bind.py` module with schema validation and `bind()` free function for composing catalog entries (source-first, variadic — accepts `CatalogEntry` or `Expr`)
- CLI `run` subcommand: compose entries, execute, print results; auto-catalogs by default (`--execute-only` to skip), `-a` for alias, `-c` for inline code
- CLI `build` subcommand: compose entries, produce build artifact
- Restricted `eval()` (not `exec()`) for inline code with only `xo`, `ibis`, and `source` in scope
- Full YAML serialization/deserialization, dask normalization, and graph traversal support for `CatalogSource`

## Test plan
- [x] `test_bind.py`: ExprKind classification, schema validation, bind roundtrip, variadic bind, inline chaining, provenance preservation (35 tests)
- [x] `test_cli.py`: CLI tests for `run` and `build` commands
- [x] Existing `test_catalog.py` updated for new `ExprKind.Source` and `Composed` variants
- [ ] CI passes

### DIFF
--- a/python/xorq/catalog/tests/test_bind.py
+++ b/python/xorq/catalog/tests/test_bind.py
@@ -57,365 +57,388 @@ def catalog_with_bound(catalog_with_entries):
 # --- ExprKind tests ---
 
 
-class TestExprKind:
-    def test_source_kind(self, source_expr):
-        meta = ExprMetadata(source_expr)
-        assert meta.kind == ExprKind.Source
+def test_source_kind(source_expr):
+    meta = ExprMetadata(source_expr)
+    assert meta.kind == ExprKind.Source
 
-    def test_unbound_kind(self, transform_expr):
-        meta = ExprMetadata(transform_expr)
-        assert meta.kind == ExprKind.UnboundExpr
-        assert meta.schema_in is not None
 
-    def test_bound_kind(self, catalog_with_entries):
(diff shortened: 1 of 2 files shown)


python/xorq/catalog/bind.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _schema_errors(source_schema, transform_schema):
+    """Return error strings for columns that are missing or type-mismatched."""
+    return tuple(
+        f"  missing: {col}"
+        if col not in source_schema
+        else f"  type mismatch: {col} (source: {source_schema[col]}, transform: {typ})"
+        for col, typ in transform_schema.items()
+        if col not in source_schema or source_schema[col] != typ
+    )
```
dlovell, line 16: we are allowing `transform_schema` to be a subset of `source_schema`?

hussainsultan: I can see why this may not be the default but seems practical.

PR #1731, opened 2026-03-20
fix(caching): avoid wrapping Expr without backend

mesejo:
When an expression containing a ParquetSnapshotCache is serialized and deserialized, cache.storage.source is reconstructed from the YAML profile as a fresh backend instance with a different Profile.idx.  The identity comparison in maybe_prevent_cross_source_caching treated this new instance as a different backend and wrapped the parent expression with into_backend, which changed the SQL and produced a different cache key at execute time than at build time.

Fix by comparing backends via Profile.almost_equals(), which ignores the session-scoped idx counter and compares only connection parameters.  Add _backends_equivalent() as a named helper to make the intent explicit.

Add two regression tests in test_compiler.py:
- test_memtable_cache_key_stable_across_roundtrip: asserts calc_key on the sanitized expr equals calc_key on the loaded expr
- test_memtable_creates_same_key: end-to-end check that sanitized.execute() and loaded.execute() write to the same cache file

### DIFF
--- a/python/xorq/caching/__init__.py
+++ b/python/xorq/caching/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import functools
 
 from attr import (
@@ -209,13 +210,7 @@ class GCSCache(Cache):
 
 
 def maybe_prevent_cross_source_caching(expr, storage):
-    from xorq.expr.relations import (  # noqa: PLC0415
-        into_backend,
-    )
-
-    try:
+    with contextlib.suppress(XorqError):
         if storage.storage.source != expr._find_backend(use_default=False):
-            expr = into_backend(expr, storage.storage.source)
-    except XorqError:
-        pass
+            return expr.into_backend(storage.storage.source)
     return expr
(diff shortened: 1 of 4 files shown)


python/xorq/caching/__init__.py
```diff
@@ -207,11 +207,24 @@ def from_kwargs(cls, bucket_name, source):
...
 
 
+def _backends_equivalent(a, b):
+    """True if a and b are the same backend or have the same connection profile."""
+    if a is b:
+        return True
+    if (
+        hasattr(a, "_profile")
+        and hasattr(b, "_profile")
+        and a._profile.almost_equals(b._profile)
```
dlovell: this makes it unclear if we expect this function to work when passed something other than a connection

python/xorq/vendor/ibis/expr/types/relations.py
```diff
@@ -3416,7 +3416,7 @@ def cache(self, cache=None) -> Table:
             name=CACHED_NODE_NAME_PLACEHOLDER,
             schema=expr.schema(),
             parent=expr,
-            source=current_backend,
+            source=cache.storage.source,
```
dlovell, line 3417: what does this change impact?

mesejo: It has no impact, `current_backend` should always be the same as `cache.storage.source`

PR #1735, opened 2026-03-22
ref(hash): use __func_name__ for UDF SQL names and extract canonicali…

dlovell:
…ze_expr

UDF SQL names used a session-global counter suffix (e.g. add_one_0) via type(op).__name__, making build hashes non-deterministic across sessions. Switch __sql_name__ and all backend UDF registration to use __func_name__ (the user-given name), which is stable. This aligns SQL generation with how DataFusion already registers UDFs.

Extract canonicalize_expr as a shared primitive combining _sanitize_generated_names and normalize_profiles, so both the YAML serialization path (ExprDumper) and the hashing path (get_expr_hash) operate on the same canonical expression form.

### DIFF
--- a/python/xorq/common/utils/name_utils.py
+++ b/python/xorq/common/utils/name_utils.py
@@ -7,7 +7,7 @@ def make_name(prefix, to_tokenize):
     from xorq.ibis_yaml.config import config  # noqa: PLC0415
 
     tokenized = dask.base.tokenize(to_tokenize)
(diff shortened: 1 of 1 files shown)


python/xorq/vendor/ibis/backends/sql/compilers/base.py
mesejo, line 1205: I think we can call op.__func_name__ directly here

PR #1747, opened 2026-03-25
feat: add xorq param 

mesejo:


python/xorq/common/utils/graph_utils.py
```diff
@@ -294,6 +295,19 @@ def replacer(node, kwargs):
     return replace_nodes(replacer, expr).to_expr()
 
 
+def validate_params(expr):
```
dlovell, line 329: this should collect all issues and report them all at once

python/xorq/common/utils/name_utils.py
```diff
@@ -1,9 +1,14 @@
 import re
 
+import dask
```
dlovell: why was dask promoted to a top level import?

python/xorq/expr/api.py
```diff
@@ -407,7 +412,15 @@ def replacer(node, kwargs):
 
 
 @tracer.start_as_current_span("_transform_expr")
-def _transform_expr(expr, **kwargs):
+def _transform_expr(expr, params=None, **kwargs):
+    """Transform an expression for execution, binding any named scalar parameters."""
+    name_values = {
+        p.op().label: v
+        for p, v in (params or {}).items()
+        if hasattr(p, "op") and isinstance(p.op(), NamedScalarParameter)
```
dlovell: does this allow inapplicable and/or invalid values in `params`?

PR #1753, opened 2026-03-26
feat(catalog): compose and run commands

hussainsultan:
## Summary
- Add `xorq catalog compose` command to assemble, build, and persist composed expressions to catalog (with `--dry-run`, `--alias`, `--code`)
- Add `xorq catalog run` command that composes and executes in one shot — accepts multiple entries, inline code, all output formats, `--limit`, and Arrow IPC stdin via shared `read_pyarrow_stream`/`maybe_open` machinery

## Test plan
- [x] `python -m pytest python/xorq/catalog/tests/test_bind.py -v` — ExprComposer, bind, source, provenance tests
- [x] `python -m pytest python/xorq/catalog/tests/test_cli.py -v` — run (single, multi-entry, piped arrow, code, limit, formats), compose (alias, dry-run, code), roundtrip tests

### DIFF
--- a/python/xorq/catalog/tests/test_catalog.py
+++ b/python/xorq/catalog/tests/test_catalog.py
@@ -38,7 +38,7 @@ def test_catalog_add(catalog, data_dict):
 
     # test not exists condition
     path = next(iter(data_dict.values()))
-    with pytest.raises(AssertionError):
+    with pytest.raises(ValueError, match="already exists"):
         catalog.add(path)
 
 


python/xorq/catalog/cli.py
```diff
@@ -415,3 +419,205 @@ def check(ctx):
...
+                        except (pa.ArrowInvalid, pa.ArrowException) as err:
+                            raise click.ClickException(
+                                f"Entry {entry!r} is an unbound expression — "
+                                f"pipe Arrow data into it or pass a source entry: "
+                                f"'xorq catalog run SOURCE {entry} -o - -f csv'."
+                            ) from err
+                        span.set_attribute("piped_stdin", True)
+                        expr = replace_unbound(catalog_entry.expr, input_expr.op())
+                    case _:
+                        expr = _make_source_expr(catalog_entry)
```
ghoersti: @hussainsultan should we be checking here that the catalog entry is a source ? 
since 
```python
    case _:
                        expr = _make_source_expr(catalog_entry)
```  

is wildcard ?

python/xorq/catalog/cli.py
```diff
@@ -415,3 +419,198 @@ def check(ctx):
...
+        span.set_attributes({"entries": entries, "has_code": code is not None})
+        with click_context_catalog(ctx):
+            if not entries:
+                raise click.UsageError("At least one entry is required.")
+
+            catalog = ctx.obj.make_catalog(init=False)
+
+            if len(entries) > 1:
+                expr = _compose_expr(catalog, entries, code)
+            else:
```
dlovell, line 613: its not clear why it should matter how many entries there are: what do we gain with a special code path for "no subsequent transforms" when we still have to deal with a "and maybe a special text transform at the end"?

dlovell: reading more closely, it looks like this is (primarily?) to allow doing something like (pseudo-code) `xorq run source | xorq run transform`, with the same `xorq run` prefix regardless of whether its a "full run specification" or using pipes for what we have previously called "run-unbound". is that correct?

hussainsultan: yes, that is correct. it will be `xorq catalog run src | xor catalog run transform` this does not change `xorq run/run-unbound`

PR #1782, opened 2026-04-05
chore: split core coverage job

dlovell:
depends on #1781


mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Medium
**Location**: `python/xorq/ibis_yaml/tests/test_packager.py` — `test_sdist_builder`
**Problem**: `PackagedBuilder.__attrs_post_init__` now validates the sdist via `SdistArchive`, which requires `uv.lock` to be embedded. The test passes `zip_path` from `prep_template_tmpdir` (a downloaded template zip). If any downloaded template was built before `uv.lock` was force-included in sdists (`pyproject.toml` change in this PR), it will fail immediately with `FileNotFoundError` — not from the build step but from construction, with no hint that the template is stale.
**Fix**: Either assert `uv.lock` is present in the template zip at the top of the test with a clear skip/fail message, or run `SdistPackager` to produce a fresh sdist rather than using the raw template zip directly.

---

**Severity**: Medium
**Location**: `python/xorq/ibis_yaml/packager.py:55` — `resolve_python_version`
**Problem**: `get_acceptable_python_versions(path)[-1]` raises bare `IndexError` if `requires-python` excludes every version in the hard-coded 3.8–3.13 range (e.g., `>=3.14`). The error gives no indication of what specifier caused the failure or what range was tested.
**Fix**: Guard with an explicit check and raise a `ValueError` with the specifier and tested range in the message.

---

**Severity**: Low
**Location**: `python/xorq/common/utils/zip_utils.py:140` — `replace_toplevel`
**Problem**: Temp path is `zip_path.with_suffix(".tmp.zip")`, a fixed sibling of the original. Concurrent calls on the same archive (e.g., parallel test workers) will collide and corrupt the file.
**Fix**: Use `tempfile.NamedTemporaryFile(dir=zip_path.parent, suffix=".tmp.zip", delete=False)` to get a unique temp path in the same directory, then `rename` atomically.

---

**Severity**: Low
**Location**: `python/xorq/common/utils/zip_utils.py:57` — `calc_zip_content_hexdigest`
**Problem**: The outer `hashlib.md5(usedforsecurity=False)` is fixed, but the inner `hashlib.md5` passed to `file_digest` still lacks `usedforsecurity=False`, breaking on FIPS-enforcing systems.
**Fix**: Pass `lambda: hashlib.md5(usedforsecurity=False)` to `file_digest` instead of `hashlib.md5`.

---

**Severity**: Low
**Location**: `python/xorq/ibis_yaml/packager.py:422` — `uv_tool_run` nix branch
**Problem**: `os.environ.get("UV_TOOL_RUN_LD_LIBRARY_PATH", "")` silently falls back to an empty string when the env var is unset in a nix shell, setting `LD_LIBRARY_PATH=""` and potentially masking a misconfiguration. Old code raised `KeyError` immediately.
**Fix**: Use `os.environ["UV_TOOL_RUN_LD_LIBRARY_PATH"]` (keep the KeyError) or add an explicit check with a clear error message when `in_nix_shell()` is true but the var is absent.

---

## Summary

Refactors the sdist pipeline from a fragile `Popened`-wrapper abstraction to standard `subprocess.run`, introduces `SdistArchive` for validated sdist handles, and enforces `uv.lock` + `requirements.txt` co-embedding — all architecturally sound, with the main risk being that `test_sdist_builder` implicitly assumes downloaded template zips already contain `uv.lock`.

</details>

---
*Review type: default | Agent: claude-code | Job: 20*

PR #1783, opened 2026-04-06
feat(metadata): add cache_keys to ExprMetadata (XOR-281)

mesejo:
Store the cache key for the root CachedNode in ExprMetadata instead of only baking in the full filesystem path. parquet_cache_paths embeds a machine-specific path at build time; cache_keys are portable — the path can be resolved at read time from key + local cache directory.

- Add `cache_keys: tuple[str, ...]` field to ExprMetadata
- In from_expr(), compute key only when root node is CachedNode (expr.ls.is_cached), not for deeply nested cached nodes
- Serialize/deserialize via to_dict()/from_dict(); omitted when empty
- Add CatalogEntry.cache_keys property delegating to metadata.cache_keys

### DIFF
--- a/python/xorq/catalog/tui.py
+++ b/python/xorq/catalog/tui.py
@@ -97,7 +97,7 @@ def _format_cached(value: bool | None) -> str:
             return "—"
 
 
-def get_cache_keys_paths(cache_keys: CacheKey) -> tuple[str, ...]:
+def get_cache_keys_paths(cache_keys: tuple[CacheKey, ...]) -> tuple[str, ...]:
     return tuple(
         str(resolve_parquet_cache_path(ck.relative_path, ck.key)) for ck in cache_keys
     )


python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -738,6 +739,20 @@ def _extract_kind(unbound_node, catalog_tag_nodes, is_source):
...
 
+def _validate_cache_keys(instance, attribute, value):
+    """deep_iterable(instance_of(CacheKey)) but with a deferred import to break
+    the cycle: core.py → caching_utils → relations → … → core.py"""
+    from xorq.common.utils.caching_utils import CacheKey  # noqa: PLC0415
+
+    if not isinstance(value, tuple):
+        raise TypeError(f"'cache_keys' must be a tuple, got {type(value).__name__}")
+    for ck in value:
+        if not isinstance(ck, CacheKey):
```
dlovell: all invalid instances should be collected and raised together

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -777,14 +792,27 @@ class ExprMetadata:
         default=None, validator=optional(instance_of(ibis.expr.schema.Schema))
     )
     root_tag: Optional[str] = field(default=None)
-    parquet_cache_paths: tuple[str, ...] = field(factory=tuple)
+    cache_keys: tuple[CacheKey, ...] = field(
+        factory=tuple, validator=_validate_cache_keys
```
dlovell: this should be `deep_iterable`

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -777,14 +792,27 @@ class ExprMetadata:
...
 
+    @staticmethod
+    def _parse_cache_keys(raw):
+        """Convert a list of ``{key, relative_path}`` dicts into a tuple of CacheKey."""
+        from xorq.common.utils.caching_utils import CacheKey  # noqa: PLC0415
+
+        if not raw:
+            return ()
+        return tuple(
+            CacheKey(key=ck["key"], relative_path=ck["relative_path"]) for ck in raw
```
dlovell: does this intentionally permit anything that has at least two items: key and relative_path

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -825,11 +853,16 @@ def from_expr(cls, expr):
...
-            if isinstance(cn.cache, ParquetSnapshotCache)
+        op = expr.op()
+        cache_keys = (
+            (
+                CacheKey(
+                    key=expr.ls.get_key(),
+                    relative_path=str(op.cache.storage.relative_path),
+                ),
+            )
+            if expr.ls.is_cached and isinstance(op.cache, ParquetSnapshotCache)
```
dlovell, line 861: if these are ***only*** parquet keys and ***only*** snapshot keys then the name should make that clear

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -862,7 +895,14 @@ def to_dict(self):
                 ),
                 ("schema_out", toolz.valmap(str, self.schema_out)),
                 ("root_tag", self.root_tag),
-                ("parquet_cache_paths", list(self.parquet_cache_paths) or None),
+                (
+                    "cache_keys",
+                    [
+                        {"key": ck.key, "relative_path": ck.relative_path}
```
dlovell: what is the difference between this and `attr.asdict`?

dlovell:
some of these tests are just checking attrs validators
string containment tests should be path checks
from_kwargs on attrs classes are redundant
see https://github.com/xorq-labs/xorq/tree/dan/feat/xor-281-cache-keys-expr-metadata

PR #1787, opened 2026-04-06
feat(metadata): compute synthetic cache keys for uncached expressions (XOR-284)

mesejo:


python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -922,16 +925,18 @@ def from_expr(cls, expr):
...
-            (
-                CacheKey(
-                    key=expr.ls.get_key(),
-                    relative_path=str(op.cache.storage.relative_path),
-                ),
+        cached_expr = (
+            expr
+            if expr.ls.is_cached and isinstance(expr.op().cache, ParquetSnapshotCache)
+            else expr.as_table().ls.uncached_one.cache(
+                cache=ParquetDummySnapshotCache.from_kwargs()
```
dlovell: if we populate this unconditionally, how does the user know if the expr is actually cached or not?

mesejo: Do  you mean if this there is cache parquet file or if the expr actually have a cachenode?

dlovell (approved):

PR #1793, opened 2026-04-06
fix(ibis_yaml): skip _MISSING sentinel when binding default params

hussainsultan:
## Summary

- Fix `InputTypeError` during `xorq build` when expressions have required named parameters (no default value)
- `_extract_sql_queries` filtered defaults with `is not None`, but the sentinel for required parameters is `_MISSING`, not `None` — so `_MISSING` values leaked into `bind_params` which tried to `dt.infer(_MISSING)` and crashed

## Repro

```
$ xorq build expr.py
Building expr from expr.py
Traceback (most recent call last):
  ...
  File ".../xorq/ibis_yaml/compiler.py", line 364, in _extract_sql_queries
    clean = bind_params(clean, defaults)
  File ".../xorq/expr/api.py", line 769, in bind_params
    if name in named and not dt.infer(value).castable(named[name].dtype):
  File ".../xorq/vendor/ibis/expr/datatypes/value.py", line 35, in infer
    raise InputTypeError(
xorq.common.exceptions.InputTypeError: Unable to infer datatype of value _MISSING
  with type <class 'xorq.expr.operations._MissingSentinel'>
```

## Fix

Add `_MISSING` sentinel check alongside the existing `None` check when collecting defaults to bind.

## Test plan

- [ ] `xorq build expr.py` completes without error for expressions with required named parameters

### DIFF
--- a/python/xorq/ibis_yaml/compiler.py
+++ b/python/xorq/ibis_yaml/compiler.py
@@ -55,6 +55,7 @@ from xorq.common.utils.node_utils import (
 )
 from xorq.config import _backend_init
 from xorq.expr.api import deferred_read_parquet, read_parquet
+from xorq.expr.operations import _MISSING
 from xorq.expr.relations import (
     CachedNode,
     Read,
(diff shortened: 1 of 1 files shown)


python/xorq/ibis_yaml/compiler.py
```diff
@@ -356,12 +356,15 @@ def _extract_sql_queries(expr, kind) -> tuple[tuple[str, str, str], ...]:
     # Bind named params to their defaults so SQL generation doesn't see them
     named = {n.label: n for n in clean.op().find(NamedScalarParameter)}
     if named:
+        from xorq.expr.operations import _MISSING  # noqa: PLC0415
```
mesejo: Why the import is not at the top?

hussainsultan: i think it should be.

PR #1810, opened 2026-04-11
fix(graph): prevent Rich unbounded recursion in Node.__rich_repr__

dlovell:
## Summary
- `Node.__rich_repr__` yielded raw child Node objects, causing `rich.pretty.traverse` to recurse into the entire expression DAG without depth limit
- This hangs structlog when logging exceptions with `exc_info=True` — structlog's `to_repr` calls `rich.pretty.traverse` with no `max_depth` on frame-local variables
- Fix: yield compact `"TypeName(...)"` strings for child Node args; non-Node args (Schema, metadata dicts, TableProxy) still pass through normally

## Test plan
- [ ] Verify `structlog.warning("...", exc_info=True)` no longer hangs when frame locals include ibis expressions
- [ ] Existing builder and ibis vendor tests pass (46 + 59 tests verified locally)
- [ ] CI passes under pytest-xdist without `sys.setrecursionlimit(200)` workaround

### DIFF
--- a/python/xorq/vendor/ibis/common/graph.py
+++ b/python/xorq/vendor/ibis/common/graph.py
@@ -284,8 +284,17 @@ class Node(Hashable):
         return tuple(_flatten_collections(self.__args__))
 
     def __rich_repr__(self):
-        """Support for rich reprerentation of the node."""
-        return zip(self.__argnames__, self.__args__)
+        """Support for rich representation of the node.
+
+        Child Nodes are rendered as compact type summaries to prevent
+        unbounded recursion when Rich traverses deep expression graphs
+        (e.g. during structlog exception logging with show_locals=True).
+        """
+        for name, arg in zip(self.__argnames__, self.__args__):
+            if isinstance(arg, Node):
+                yield name, f"{type(arg).__name__}(...)"
+            else:
+                yield name, arg
 
     def map(self, fn: Callable, filter: Optional[Finder] = None) -> dict[Node, Any]:
         """Apply a function to all nodes in the graph.


python/xorq/vendor/ibis/common/graph.py
```diff
@@ -284,8 +284,17 @@ def __children__(self) -> tuple[Node, ...]:
         return tuple(_flatten_collections(self.__args__))
 
     def __rich_repr__(self):
-        """Support for rich reprerentation of the node."""
-        return zip(self.__argnames__, self.__args__)
+        """Support for rich representation of the node.
+
+        Child Nodes are rendered as compact type summaries to prevent
```
ghoersti, line 289: This seems reasonable to yield strings in the case of nodes in the repr.

mesejo (approved):

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `python/xorq/vendor/ibis/common/graph.py`, line 290–295 (new `__rich_repr__`)
- **Problem**: Fix is incomplete. `__args__` can contain tuples, lists, and dicts holding `Node` instances (confirmed by `_flatten_collections` and `__children__`). When `arg` is a collection, the fix passes it through unchanged. Rich then renders that collection, encounters the `Node` within it, calls `__rich_repr__` on it, and the recursion continues. For graphs where Nodes are stored inside tuple/list args (e.g. `children: tuple[Node, ...]`), the original unbounded recursion still occurs.
- **Fix**: Extend the check to recursively substitute Nodes inside collections:
  ```python
  def __rich_repr__(self):
      def _summarize(val):
          if isinstance(val, Node):
              return f"{type(val).__name__}(...)"
          elif isinstance(val, (list, tuple)):
              summarized = [_summarize(v) for v in val]
              return type(val)(summarized)
          elif isinstance(val, dict):
              return {k: _summarize(v) for k, v in val.items()}
          return val

      for name, arg in zip(self.__argnames__, self.__args__):
          yield name, _summarize(arg)
  ```

---

- **Severity**: Low
- **Location**: No test file added
- **Problem**: No test verifies that `__rich_repr__` terminates on a deep graph (e.g. the `MyNode` chain already in the docstring is a ready-made fixture).
- **Fix**: Add a test calling `rich.repr.auto` or directly consuming `node.__rich_repr__()` on a deep chain to assert no recursion.

## Summary

Prevents Rich recursion by truncating direct `Node` children to type-name strings, but misses `Node` instances nested inside collection-typed args (`tuple[Node, ...]`), leaving the original bug reachable through that path.

</details>

---
*Review type: default | Agent: claude-code | Job: 19*

PR #1817, opened 2026-04-12
docs: add ADR template

dlovell:
## Summary

- Adds `docs/adr/template.md` with the common structure extracted from ADR-0002 through ADR-0005
- Includes HTML comment guidance on when to write an ADR vs. a commit message, how to title decisions (not problems), and when Rationale warrants its own section

## Test plan

- [x] Reviewed against existing ADRs for consistency

### DIFF
--- a/docs/adr/template.md
+++ b/docs/adr/template.md
@@ -1,0 +1,90 @@
+# ADR-NNNN: <title — the decision, not the problem>
+
+<!--
+The title should name the decision or design choice, not the bug or symptom.
+Good:  "Make git-annex optional via a CatalogBackend abstraction"
+Bad:   "Fix catalog dependency issue"
+
+Delete all guidance comments (HTML comments like this one) before merging.
+-->
+
+- **Status:** Proposed | Accepted | Rejected | Deprecated | Superseded by ADR-NNNN
+- **Date:** YYYY-MM-DD <!-- when the decision was last updated -->
+- **Deciders:** <names>
+
+## Context
+
+<!--
+What situation or problem prompted this decision? Include enough background
+that a reader unfamiliar with the current code can understand why a decision
+is needed. If there was a specific incident (CI hang, production bug), link
+to it.
+
+This section should make clear that reasonable people could disagree on the
+right approach. If the "decision" is simply "fix the obvious bug," it
+belongs in a commit message, not an ADR.
+-->
+
+## Decision drivers
+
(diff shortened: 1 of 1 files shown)


ghoersti (approved):

docs/adr/template.md
```diff
@@ -0,0 +1,77 @@
...
+## Decision
+
+<!--
+What did we decide to do, and how does it work? This is the core of the ADR.
+
+Structure as prose with subsections (###) for distinct design choices.
+Use tables for comparisons, code blocks for key interfaces or signatures.
+Explain the reasoning inline — why this approach, not just what it does.
+
+Avoid pasting full implementations; reference file:line instead. Include
```
paddymul, line 35: revision:file:line

file:line is likely to change with the implmentation of the ADR

paddymul: in fact I prefer a link to github at that revision with those lines highlighted.  I find it faster to see the code in question that way than running git operations.

docs/adr/template.md
```diff
@@ -0,0 +1,77 @@
...
+<!--
+The title should name the decision or design choice, not the bug or symptom.
+Good:  "Make git-annex optional via a CatalogBackend abstraction"
+Bad:   "Fix catalog dependency issue"
+-->
+
+- **Status:** Proposed | Accepted | Superseded by ADR-NNNN
+- **Date:** YYYY-MM-DD
+- **Context area:** `path/to/primary/file.py`
+
```
paddymul, line 12: I think it would be helpful for a "Planned" state, for something that has been accepted but not yet implemented.  

It would also be helpful to know the commit that implemented the change.
You can't know a priori the commit that a change will be merged to main on, but you can know the PR where something is implemented.  This can help when the ADR is later referenced.

paddymul:
I like the template generally, I hada. couple of requests.

paddymul (approved):

docs/adr/template.md
dlovell: can you show me an example of someone else having a "Planned" state?
it should not be accepted if it can't be implemented or you don't know all the implications
you can't know the implications until you have both a proposal and an implementation

dlovell: the adr lives in the code, why would the implementation commit not be the same one where the ADR moves from proposed to accepted?
can you show me an example of someone else doing something different?

PR #1819, opened 2026-04-13
docs(readme): refresh README (XOR-245)

mesejo:
Adds install verification, a 3-step quickstart (write → build → catalog), catalog CLI reference table, expression types (source/unbound/composed), sklearn side-by-side comparison, BSL integration example, TUI launch instructions, and architecture layer table. Moves the problem section after the solution content so new users see the value proposition first.


mesejo:
## roborev: Pass

No issues found.

---
*Review type: default | Agent: claude-code | Job: 18*

PR #1823, opened 2026-04-14
fix(caching): resolve ParquetCache "relation already exists"

### LINKED ISSUE #1820: ParquetCache with a Postgres source fails on second execution with "relation already exists"
Reproduction script                                                                                                                                                                                         
                                                                                                                                                                                                              
```python
from __future__ import annotations

import tempfile
import time
from pathlib import Path

import pyarrow as pa
import pyarrow.parquet as pq

import xorq.api as xo
from xorq.caching import ParquetCache


_DATA = pa.table(
    {
        "id": pa.array([1, 2, 3, 4, 5], type=pa.int64()),
        "name": pa.array(["alice", "bob", "carol", "dave", "eve"], type=pa.string()),
        "score": pa.array([9.1, 7.4, 8.8, 6.0, 9.9], type=pa.float64()),
    }
)

_tmpdir = tempfile.mkdtemp(prefix="xorq_demo_")
parquet_path = Path(_tmpdir) / "people.parquet"
pq.write_table(_DATA, parquet_path)
print(f"wrote source parquet to {parquet_path}")

ddb = xo.duckdb.connect()
pg = xo.postgres.connect_env()

cache_dir = Path(_tmpdir) / "parquet-cache"

cache = ParquetCache.from_kwargs(
    source=pg,
    relative_path=cache_dir,
)

expr = (
    ddb.read_parquet(str(parquet_path))
    .into_backend(pg)
    .filter(xo._.score > 7.0)
    .cache(cache=cache)
)

for run in (1, 2):
    t0 = time.perf_counter()
    result = expr.execute()
    elapsed = time.perf_counter() - t0

    cache_path = cache.storage.get_path(expr.ls.get_key())
    hit = "HIT" if run == 2 else "MISS"
    print(f"\n--- run {run} ({hit}) [{elapsed:.3f}s] ---")
    print(f"cache file: {cache_path}  exists={cache_path.exists()}")
    print(result.to_string(index=False))

```
   
##  What happens                                                                                                                                                                                                
                  
  Run 1 (cache miss) — succeeds. The expression is executed, the result is written to a local parquet file, and ParquetStorage.get() calls deferred_read_parquet(path=<cache_file>, con=pg, table_name=<key>).
   Materializing that deferred read via make_dt() invokes pg.read_parquet(path, table_name=key) which issues CREATE TABLE "<key>" ... in Postgres and returns the table.
                                                                                                                                                                                                              
  Run 2 (cache hit) — fails.

  adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: [libpq] Failed to create table:
  ERROR:  relation "letsql_cache-a6bf7c3585e3ef60cdeb095e2be25360" already exists                                                                                                                             
                                                                                                                                                                                                              
 ### Root cause                                                                                                                                                                                                  
https://github.com/xorq-labs/xorq/blob/f386dbf2944715f91dcb23e1cfe6190696472649/python/xorq/caching/storage.py#L127-L133

deferred_read_parquet is called with no mode kwarg, so pg.read_parquet defaults to mode="create". Postgres materializes the cached parquet file as a permanent table named after the cache key. On the second execution key_exists(key) is True (the parquet file is on disk), so set_default calls get(key) again — which tries to CREATE TABLE with the same name → conflict.
                                                                                                                                                                                                              
##  Expected behaviour                                                                                                                                                                                          
                  
The second execute() should return the cached result without error. If Postgres is the cache source, get() should either read from the already-ingested table or use mode="replace" when calling read_parquet.

mesejo:
- Pass mode="replace" in ParquetStorage.get() for ADBC-backed sources (postgres, sqlite, snowflake, databricks) so cache hits do not fail when the materialized table already exists from a prior execution.
- Extract shared adbc_ingest logic into ADBCBase mixin; PgADBC and SQLiteADBC now inherit from it instead of duplicating the method body.

- Add regression tests in test_cache.py covering both postgres and sqlite.

closes #1820

### DIFF
--- a/examples/deferred_read_csv.py
+++ b/examples/deferred_read_csv.py
@@ -28,9 +28,9 @@ pd_expr = xo.deferred_read_csv(con=pd_con, path=csv_path, table_name=csv_name).f
 
 # we can even work with postgres!
 pg = xo.postgres.connect_env()
-pg_expr = xo.deferred_read_csv(con=pg, path=csv_path, table_name=csv_name).filter(
-    _.sepal_length > 6
-)
+pg_expr = xo.deferred_read_csv(
+    con=pg, path=csv_path, table_name=csv_name, mode="create"
+).filter(_.sepal_length > 6)
 
 # NOTE: we can't re-run the expr in postgres
 # UNLESS we set the create_table mode to "replace"
--- a/python/xorq/caching/storage.py
+++ b/python/xorq/caching/storage.py
@@ -31,9 +31,6 @@ from xorq.config import _backend_init, options
 from xorq.vendor import ibis
 
 
-_ADBC_BACKENDS = ("sqlite", "postgres", "snowflake", "databricks")
-
-
 def resolve_parquet_cache_dir(
     relative_path: Path | str,
     base_path: Path | None = None,
@@ -128,12 +125,10 @@ class ParquetStorage(CacheStorage):
         return self.get_path(key).exists()
 
     def get(self, key):
-        kwargs = {"mode": "replace"} if self.source.name in _ADBC_BACKENDS else {}
         op = deferred_read_parquet(
             path=self.get_path(key),
             con=self.source,
             table_name=key,
-            **kwargs,
         ).op()
         return op
 
--- a/python/xorq/caching/storage.py
+++ b/python/xorq/caching/storage.py
@@ -31,6 +31,9 @@ from xorq.config import _backend_init, options
 from xorq.vendor import ibis
 
 
+_ADBC_BACKENDS = ("sqlite", "postgres", "snowflake", "databricks")
+
+
 def resolve_parquet_cache_dir(
     relative_path: Path | str,
     base_path: Path | None = None,
@@ -125,10 +128,12 @@ class ParquetStorage(CacheStorage):
         return self.get_path(key).exists()
 
     def get(self, key):
+        kwargs = {"mode": "replace"} if self.source.name in _ADBC_BACKENDS else {}
         op = deferred_read_parquet(
             path=self.get_path(key),
             con=self.source,
             table_name=key,
+            **kwargs,
         ).op()
         return op
 
--- a/python/xorq/common/utils/adbc_utils.py
+++ b/python/xorq/common/utils/adbc_utils.py
@@ -1,0 +1,29 @@
+from abc import ABC, abstractmethod
+
+
+class ADBCBase(ABC):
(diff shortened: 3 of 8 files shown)


dlovell:
PR description says "Refactor rename_table_pg and _copy_table_pg out of make_table_temporary to module level in postgres_utils."
I don't see that in the code. Is that still applicable?

python/xorq/caching/storage.py
```diff
@@ -125,10 +128,12 @@ def exists(self, key):
         return self.get_path(key).exists()
 
     def get(self, key):
+        kwargs = {"mode": "replace"} if self.source.name in _ADBC_BACKENDS else {}
```
dlovell: why is this logic here instead of in `deferred_read_*`?

python/xorq/common/utils/adbc_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    identical implementation is not duplicated across every backend helper.
+    """
+
+    __slots__ = ()
+
+    @abstractmethod
+    def get_conn(self, **kwargs): ...
+
+    def adbc_ingest(
+        self, table_name, record_batch_reader, mode="create", temporary=False, **kwargs
```
dlovell: users should be able to pass `conn_kwargs` to `self.get_conn`

python/xorq/caching/storage.py
mesejo: So the reasoning was that not all deferred_read_* for ADBC backends should be with mode = replace, if the user reads a table (`foo`) using deferred_read_parquet for SQLITE and the there is a table with that name (`foo`) I expect the Read op to fail.

dlovell:
fundamentally, this PR is not about caching but about `deferred_read_*`

examples/deferred_read_csv.py
```diff
@@ -28,9 +28,9 @@
 
 # we can even work with postgres!
 pg = xo.postgres.connect_env()
-pg_expr = xo.deferred_read_csv(con=pg, path=csv_path, table_name=csv_name).filter(
-    _.sepal_length > 6
-)
+pg_expr = xo.deferred_read_csv(
+    con=pg, path=csv_path, table_name=csv_name, mode="create"
```
dlovell, line 32: will this fail if the user runs the example twice?

dlovell:
should we be defaulting `temporary=True` for some invocations of `deferred_read_*`?

dlovell (approved):

PR #1847, opened 2026-04-20
feat(catalog): add replay --rebuild to regenerate entries under current code

dlovell:
## Summary

Add `catalog replay --rebuild` to regenerate catalog entries under current code, plus supporting infrastructure for builder-aware rebuild.

> Split out of this PR (do not block on; can land independently):
> - Annex `fileprefix` namespacing → #1895
> - Tokenize-tolerant `normalize_read` via `read_path` → #1917
> - `Profile.validate_con_name` informative error → #1914
> - Downstream BSL CI workflow → #1915

### `replay --rebuild` core (`replay.py`, `composer.py`)
- Rebuild mode loads each entry's `lazy_expr`, rewrites `HashingTag` catalog refs via a remap dict keyed by old entry hash, and re-adds the fresh expr to the target so new hashes reflect current `build_metadata`
- `ExprComposer.with_inputs_translated(remap, to_catalog)` performs the recipe translation: source + transforms get rewritten to the target catalog's entries, `code` is preserved byte-for-byte, alias is pinned to the source-side selection
- `AddAlias` / `RemoveEntry` translate entry names via the same remap; `UnknownOp` inspects changed paths and raises only when catalog-managed paths are touched — non-catalog patches (e.g. README, scripts) are applied via `format-patch`/`am`
- Preserves source commit metadata (author, email, timestamps) — initial commits created by `Catalog.from_repo_path(init=True)` are rewritten via `git commit-tree` so `authored_date` matches the source. The post-rewrite `git rebase --onto` and the `UnknownOp` `git am` paths abort cleanly on failure rather than leaving the target catalog mid-rebase / mid-am.
- `_rewrite_noop_commits` refuses on a detached HEAD with an actionable error rather than letting `active_branch.name` raise a confusing `TypeError`.

### Builder-aware rebuild (`builders/__init__.py`, `pipeline_lib.py`)
- `TagHandler` gains an optional `reemit` field; `get_rebuild_dispatch` picks handler-level `reemit`, then domain-object `reemit(tag_node, rebuild_subexpr)` (multi-output, e.g. `FittedPipeline`), then `with_inputs_translated(remap, to_catalog) + .expr` (single-output, e.g. `ExprComposer`)
- `_rebuild_subexpr` collects all outermost rebuildable tags in one pass and splices each — so a builder subtree with parallel catalog compositions (e.g. training + predict input) gets every composition rebuilt
- Registered builder tags with no rebuild protocol raise `RuntimeError` by default; pass `on_unrebuilt_builder="warn"` to `Replayer` (or `_rebuild_subexpr` directly) to downgrade to a warning and pass through unchanged
- `FittedPipeline.reemit` refits on rebuilt training data, refreshing `training_hash` and step kwargs, and re-stamps the outer pipeline tag. Refuses (rather than silently producing stale UDFs) when the training subtree itself contains catalog refs, since model UDFs close over the original training expression and `rebuild_subexpr` does not descend into UDF closures.
- Schema-preservation is enforced per splice

### CLI (`cli.py`)
- `catalog replay <target> --rebuild` flag wired through Click

### API change (`catalog.py`)
- `CatalogAddition` now returns a `CatalogEntry` on the existing-entry branch (previously `None`). Required by `AddEntry.do` to access `new_entry.name` for remap recording on rebuild. The one in-tree caller affected is `cli compose`, which now uses `catalog.contains(name)` before `add()` to detect collisions instead of testing the return value.

## Test plan

- [x] `test_replay_rebuild.py` — 20 tests covering consistent target, composed_from linkage, expr rewrite, hash determinism, fresh-bind match, chained bind, code-only/full-recipe composition, outer builder wrapping, pure builder passthrough, alias preservation, remap translation, commit metadata preservation, unknown-op error, non-catalog unknown-op passthrough, content-collision remap, DAG-shared catalog tag, CLI smoke (rebuild and discriminating-vs-replay)
- [x] `test_replay_rebuild_builders.py` — 10 tests covering FittedPipeline reemit table coverage, training reservation, rebuild dispatch (handler-level, domain-object, single-output, no-protocol raise), handler-level reemit integration, schema-change refusal, missing-protocol passthrough, real-sklearn FittedPipeline roundtrip, catalog-training refusal

### DIFF
--- a/python/xorq/catalog/catalog.py
+++ b/python/xorq/catalog/catalog.py
@@ -840,7 +840,7 @@ class CatalogAddition:
                 raise ValueError(f"Entry '{self.name}' already exists in catalog")
             for catalog_alias in self.catalog_aliases:
                 catalog_alias._add()
-            return None
+            return CatalogEntry(self.name, self.catalog, require_exists=True)
         self.ensure_dirs()
         catalog_entry = self.catalog_entry
         catalog_entry.metadata_path.write_text(yaml12.format_yaml(self.metadata))
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -846,9 +846,10 @@ def compose(ctx, entries, code, alias, cache_dir, dry_run, raw_rename_params):
             entry_name = build_path.name
             aliases = (alias,) if alias else ()
             alias_existed = alias and catalog.catalog_yaml.contains_alias(alias)
-            catalog_entry = catalog.add(build_path, aliases=aliases, exist_ok=True)
+            entry_existed = catalog.contains(entry_name)
+            catalog.add(build_path, aliases=aliases, exist_ok=True)
             label = alias or entry_name
-            if catalog_entry is None:
+            if entry_existed:
                 if alias and not alias_existed:
                     click.echo(
                         f"Entry {entry_name!r} already exists; alias {alias!r} added",
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -559,6 +559,19 @@ def log(ctx, as_json):
 @click.option(
     "--dry-run", is_flag=True, help="Show what would be replayed without executing."
 )
+@click.option(
+    "--rebuild",
+    is_flag=True,
+    help=(
+        "Rebuild each entry under current code. Entries with no catalog "
+        "references are re-added from their stored expression; entries "
+        "containing catalog references (Composed, or ExprBuilder wrapping a "
+        "composition) have the catalog subtree recomposed against their "
+        "(already rebuilt) dependencies in the target catalog — outer "
+        "builder wrappings pass through untouched. Updates build_metadata "
+        "and entry hashes. Fetches annex content for every entry."
+    ),
+)
 @click.pass_context
 def replay(
     ctx,
@@ -570,6 +583,7 @@ def replay(
     preserve_commits,
     force,
     dry_run,
+    rebuild,
 ):
     """Replay catalog operations into a target catalog."""
     from xorq.catalog.catalog import Catalog
@@ -577,12 +591,12 @@ def replay(
 
     with click_context_catalog(ctx):
         source = ctx.obj.make_catalog(init=False)
-        replayer = Replayer(from_catalog=source)
+        replayer = Replayer(from_catalog=source, rebuild=rebuild)
         if dry_run:
             replayer.print_plan()
             return
         annex = _resolve_annex_option(env_file, env_prefix, gcs)
-        target = Catalog.from_repo_path(target_path, init=True, annex=annex)
+        target = Catalog.from_repo_path(target_path, annex=annex)
         replayer.replay(target, preserve_commits=preserve_commits)
         click.echo(f"Replayed {len(replayer.ops)} operations into {target_path}")
         if remote_url:
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -562,15 +562,7 @@ def log(ctx, as_json):
 @click.option(
     "--rebuild",
     is_flag=True,
-    help=(
-        "Rebuild each entry under current code. Entries with no catalog "
-        "references are re-added from their stored expression; entries "
-        "containing catalog references (Composed, or ExprBuilder wrapping a "
-        "composition) have the catalog subtree recomposed against their "
-        "(already rebuilt) dependencies in the target catalog — outer "
-        "builder wrappings pass through untouched. Updates build_metadata "
-        "and entry hashes. Fetches annex content for every entry."
-    ),
+    help="Rebuild each entry under current code (refreshes build_metadata and entry hashes).",
 )
 @click.pass_context
 def replay(
@@ -585,7 +577,15 @@ def replay(
     dry_run,
     rebuild,
 ):
-    """Replay catalog operations into a target catalog."""
+    """Replay catalog operations into a target catalog.
+
+    With ``--rebuild``, each entry is re-added under current code:
+    entries with no catalog references are re-added from their stored
+    expression; entries containing catalog references (Composed, or
+    ExprBuilder wrapping a composition) have the catalog subtree
+    recomposed against their already-rebuilt dependencies in the
+    target catalog. Outer builder wrappings pass through untouched.
+    """
     from xorq.catalog.catalog import Catalog
     from xorq.catalog.replay import Replayer
 
--- a/python/xorq/catalog/composer.py
+++ b/python/xorq/catalog/composer.py
@@ -51,6 +51,40 @@ class ExprComposer:
 
         return current
 
+    def with_inputs_translated(self, remap, to_catalog):
+        """Return a new ExprComposer with catalog inputs translated through *remap*.
+
+        ``source`` and each entry in ``transforms`` is rewritten to the
+        corresponding entry in *to_catalog* (following *remap*). ``code``
+        is preserved byte-for-byte. ``alias`` is pinned explicitly when
+        the original was ``None`` so the rebuilt source tag's alias
+        metadata is independent of aliases-list ordering drift across
+        catalogs.
+        """
+
+        def lookup(old_name):
+            new_name = remap.get(old_name, old_name) if remap else old_name
+            if new_name not in to_catalog.list():
+                raise RuntimeError(
+                    f"rebuild: recipe references {old_name!r} which has not been "
+                    f"rebuilt in target (translated to {new_name!r}). This indicates "
+                    f"a non-topological op ordering or a reference to an entry "
+                    f"outside the source catalog."
+                )
+            return to_catalog.get_catalog_entry(new_name)
+
+        new_source = lookup(self.source.name)
+        pinned_alias = self.alias
+        if pinned_alias is None and new_source.aliases:
+            pinned_alias = new_source.aliases[0].alias
+
+        return ExprComposer(
+            source=new_source,
+            transforms=tuple(lookup(t.name) for t in self.transforms),
+            code=self.code,
+            alias=pinned_alias,
+        )
+
     @classmethod
     def from_expr(cls, expr, catalog):
         """Recover an ExprComposer from a tagged expression.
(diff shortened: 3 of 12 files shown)


mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Medium
**Location**: `python/xorq/catalog/replay.py` — `_rebuild_subexpr`, assert block after `outermost` collection (~line 213 in diff)
**Problem**: The parent-before-descendant ordering invariant is guarded with `assert`, which is silently disabled under `python -O`. If violated in an optimized build, the splice logic produces silently wrong results (ancestor tag rebuilt from a stale node that `replace_nodes` can no longer find by identity, leaving it unrebuilt).
**Fix**: Replace with an explicit `if ... raise RuntimeError(...)` check, or at minimum add a comment that `-O` disables this guard.

---

**Severity**: Medium
**Location**: `python/xorq/catalog/replay.py` — `_rewrite_noop_commits`, `git rebase` call (~line 742 in diff)
**Problem**: `repo.git.rebase("--onto", prev_sha, old_base)` does not pass `--committer-date-is-author-date`. All commits rebased on top of the rewritten noop commits will have committer timestamps reset to "now", diverging from the source history. `preserve_commits` is supposed to preserve full provenance but silently loses committer dates for the non-noop suffix.
**Fix**: Pass `"--committer-date-is-author-date"` to the rebase call: `repo.git.rebase("--committer-date-is-author-date", "--onto", prev_sha, old_base)`.

---

**Severity**: Medium
**Location**: `python/xorq/common/utils/graph_utils.py` — `walk_nodes` docstring
**Problem**: The new docstring asserts "parent-before-descendant order" unconditionally, but the implementation can violate this for DAG-structured expressions (shared nodes). When a tag node `D` is reachable via two paths where one path goes through ancestor tag `B`, deduplication causes `D` to be visited before `B`, putting a descendant before its ancestor in the result. The `assert` in `_rebuild_subexpr` catches this at runtime (in non-optimized builds), but the docstring misleads callers into relying on a guarantee that holds only for trees.
**Fix**: Qualify the docstring: "…in parent-before-descendant order for tree-shaped expressions; shared (DAG) nodes may appear before a non-matching ancestor."

---

**Severity**: Low
**Location**: `python/xorq/expr/ml/pipeline_lib.py` — `FittedPipeline.reemit`
**Problem**: `self.fitted_steps[0].features`, `self.fitted_steps[0].target`, and `self.fitted_steps[0].cache` unconditionally read from the first fitted step. A pipeline with multiple fit steps (uncommon but architecturally possible) would silently use only the first step's fit parameters for the refit.
**Fix**: Assert `len(self.fitted_steps) == 1` before this block, or iterate over all fit steps.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tests/test_replay_rebuild.py` and `python/xorq/catalog/tests/test_replay_rebuild_builders.py`
**Problem**: `_replay_rebuild` helper is copy-pasted identically into both test files.
**Fix**: Move to the catalog test `conftest.py`.

---

## Summary

Adds `catalog replay --rebuild` to re-execute stored entries under current code, with `ExprComposer.with_inputs_translated` for catalog compositions, a `TagHandler.reemit` protocol for registered builders (including `FittedPipeline`), a `RebuildContext` threading remap state through all ops, and `_rewrite_noop_commits` to preserve source commit metadata in the target.

</details>

---
*Review type: default | Agent: claude-code | Job: 16*

mesejo:
## roborev: Fail

<details>

## Review Findings

---

**Severity**: Medium  
**Location**: `python/xorq/catalog/replay.py` — `_rebuild_subexpr`, `get_rebuild_dispatch` in `builders/__init__.py`  
**Problem**: `get_rebuild_dispatch(tag)` is called twice per rebuildable non-catalog tag: once during the outermost-collection loop (`if not (_is_catalog_tag(tag) or get_rebuild_dispatch(tag) is not None)`) and again inside `_rebuild_tag`. Each call executes `handler.from_tag_node(tag_node)`. For `FittedPipeline`, `from_tag_node` reconstructs the fitted model from the tag node — a potentially expensive deserialization — so each rebuildable `FittedPipeline` tag pays that cost twice per rebuild.  
**Fix**: Cache the dispatch callable when it's computed during collection: store `(tag, dispatch)` pairs in `outermost` instead of tags alone, then use the cached callable in `_rebuild_tag`.

---

**Severity**: Medium  
**Location**: `python/xorq/catalog/replay.py:~219` — `_rebuild_subexpr` assertion block  
**Problem**: The post-collection invariant check is O(|outermost|²) with a full `walk_nodes` subtree traversal per pair. For expressions with many builder tags the assertion performs quadratic graph walks on every `_rebuild_subexpr` call, and this runs in the production code path (not gated by `__debug__` or a test flag). The invariant it's checking is already structurally guaranteed by the `claimed`-set algorithm; the assertion is belt-and-suspenders over correct logic.  
**Fix**: Gate behind `if __debug__:` or convert to a dev/test-only mode flag, or drop it and rely on the algorithm invariant + unit test `test_rebuild_handles_dag_shared_catalog_tag`.

---

**Severity**: Medium  
**Location**: `python/xorq/catalog/replay.py` — `_tempfile_patch`  
**Problem**: If `os.write(fd, patch_text.encode())` raises (e.g., disk full, EINTR), `os.close(fd)` is never reached and the file descriptor leaks. The `finally` block only unlinks the path.  
**Fix**:
```python
fd, path = tempfile.mkstemp(suffix=".patch")
try:
    try:
        os.write(fd, patch_text.encode())
    finally:
        os.close(fd)
    yield path
finally:
    os.unlink(path)
```

---

**Severity**: Low  
**Location**: `python/xorq/catalog/replay.py` — `_rewrite_noop_commits`  
**Problem**: `if prev_sha is None: return` is dead code. The caller only invokes this function when `noop_ops` is non-empty, and the loop body always assigns `prev_sha` on every iteration (either `commit.hexsha` or `new_sha`). The guard can never trigger.  
**Fix**: Remove the guard; it obscures control flow without adding safety.

---

**Severity**: Low  
**Location**: `python/xorq/catalog/replay.py` — `AddEntry.do` rebuild branch  
**Problem**: Catching `except Exception` and re-raising as `RuntimeError(f"... {e}")` double-nests error messages when the inner exception is already a `RuntimeError` (e.g., from `_rebuild_subexpr`). The `from e` chain is preserved, but the top-level message reads `"rebuild: failed to rebuild entry 'foo': rebuild: ..."`.  
**Fix**: Either let the inner exceptions propagate directly (they already have context), or only catch and annotate non-`RuntimeError` exceptions.

---

## Summary

Adds `catalog replay --rebuild` to re-execute catalog entries under current code via a composable `RebuildContext`/`_rebuild_subexpr` dispatch system, with `FittedPipeline.reemit` and `ExprComposer.with_inputs_translated` as the two rebuild protocols, backed by extensive new tests.

</details>

---
*Review type: default | Agent: claude-code | Job: 33*

mesejo:
## roborev: Fail

<details>

## Review Findings

---

**Severity**: Medium  
**Location**: `python/xorq/catalog/replay.py` — `_rewrite_noop_commits`  
**Problem**: When the first noop op has `commit_metadata` but the second doesn't, the first commit is rewritten to `new_sha1`, but then `prev_sha` is overwritten with `all_commits[1].hexsha` in the `continue` branch. The final rebase becomes `rebase --onto all_commits[1].hexsha all_commits[1].hexsha` — a no-op — leaving `new_sha1` orphaned and the branch still pointing at the original chain. The rewrite is silently discarded.  
**Fix**: Track `prev_sha` separately from "the last original sha used as a parent fallback". Only update `prev_sha` to the original sha on `continue` if no rewrite has happened yet; once a rewrite has produced a `new_sha`, preserve it through `continue` iterations.

---

**Severity**: Medium  
**Location**: `python/xorq/expr/builders/__init__.py` — `get_rebuild_dispatch`, called twice per tag in `_rebuild_subexpr`  
**Problem**: `get_rebuild_dispatch` is called once in the outermost-detection loop (to test if dispatch exists) and again inside `_rebuild_tag`. Paths 2 and 3 call `handler.from_tag_node(tag_node)` eagerly on both calls, instantiating the domain object twice — for `FittedPipeline` this means deserializing/recovering the pipeline twice per rebuild.  
**Fix**: Cache the dispatch callable in `_rebuild_subexpr` alongside the tag: `outermost.append((tag, get_rebuild_dispatch(tag)))`, then use the cached callable in `_rebuild_tag` instead of re-querying.

---

**Severity**: Low  
**Location**: `python/xorq/catalog/replay.py` — `_rewrite_noop_commits`, `if n < len(all_commits)` branch  
**Problem**: The detached-HEAD guard (`if repo.head.is_detached: raise RuntimeError(...)`) only exists in the `else` branch (all commits are noops). When there are post-noop commits and HEAD is detached, `git rebase` runs but no branch ref is updated; the rebased tip is reachable only from ORIG_HEAD, effectively orphaning the pre-rebase chain.  
**Fix**: Add the detached-HEAD check before calling `repo.git.rebase` in the `if n < len(all_commits)` branch.

---

**Severity**: Low  
**Location**: `python/xorq/catalog/tests/test_replay_rebuild.py` — `test_rebuild_applies_changed_reemit`, `test_rebuild_collision_records_remap_to_existing_entry`  
**Problem**: Both tests drive replay by iterating `op.do(catalog, target, ctx)` directly, bypassing `_assert_target_pristine` and `_rewrite_noop_commits`. The remap interaction with noop-commit rewriting under `replayer.replay()` is therefore not exercised end-to-end for these scenarios.  
**Fix**: Either add a companion assertion after calling `replayer.replay(target)` in a fresh target path (inspect `target.repo.iter_commits` for correct metadata), or add a separate test that goes through the full `replay()` path and inspects `remap` via a post-replay catalog query.

---

## Summary

Adds `catalog replay --rebuild` that re-executes each stored expression under current code, with a three-tier dispatch (handler `reemit`, domain-object `reemit`, `with_inputs_translated`) and schema-preservation guards; the main correctness risk is an orphaned-commit edge case in `_rewrite_noop_commits` when noops have mixed metadata presence.

</details>

---
*Review type: default | Agent: claude-code | Job: 37*

hussainsultan (approved):

PR #1853, opened 2026-04-21
feat(plot): add PlotNode support via UDAF + Tag

hussainsultan:
Plot expressions are created by wrapping a plot function as a UDAF (via make_plot_expr) and tagging the result with PLOT_TAG. This keeps plots in the expression graph as regular table operations, with PNG/SVG bytes stored in binary columns of the cached parquet result.

The tag is a standalone constant ("plot"), not a CatalogTag variant, because plot-ness is orthogonal to catalog composition roles. Detection is via expr.ls.is_plot on the LETSQLAccessor, which checks both the tag and that the schema contains a binary column.

### DIFF
--- a/python/xorq/common/utils/plot_utils.py
+++ b/python/xorq/common/utils/plot_utils.py
@@ -1,0 +1,14 @@
+from xorq.expr import udf
+
+
+PLOT_TAG = "plot"
+
+
+def make_plot_expr(table, plot_fn, return_type, *, name="plot"):
+    plot_udaf = udf.agg.pandas_df(
+        plot_fn,
+        table.schema(),
+        return_type,
+        name=name,
+    )
+    return table.agg(plot_udaf.on_expr(table).name(name)).tag(PLOT_TAG)
--- a/python/xorq/vendor/ibis/expr/types/core.py
+++ b/python/xorq/vendor/ibis/expr/types/core.py
@@ -947,6 +947,16 @@ class LETSQLAccessor:
         return find_all_sources(self.expr)
 
     @property
+    def is_plot(self):
+        from xorq.common.utils.node_utils import find_by_expr_tag
+        from xorq.common.utils.plot_utils import PLOT_TAG
+
+        return any(
+            any(dtype.is_binary() for dtype in node.schema.values())
+            for node in find_by_expr_tag(self.expr, PLOT_TAG)
+        )
+
+    @property
     def is_multiengine(self):
         (_, *rest) = set(self.backends)
         return bool(rest)


python/xorq/common/utils/plot_utils.py
```diff
@@ -0,0 +1,14 @@
+from xorq.expr import udf
+
+
```
ghoersti, line 3: @hussainsultan  Should we follow our enum pattern here ?

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -1062,6 +946,16 @@ def backends(self):
 
         return find_all_sources(self.expr)
 
+    @property
+    def is_plot(self):
```
mesejo, line 950: Should we create a function is_tagged() to check whether the Expr is tagged with a given tag?

python/xorq/vendor/ibis/expr/types/core.py
```diff
@@ -1062,6 +946,16 @@ def backends(self):
...
+        from xorq.common.utils.node_utils import find_by_expr_tag
+        from xorq.common.utils.plot_utils import PLOT_TAG
+
+        return any(
```
mesejo, line 954: Is the objective of this function to check if the last node is a plot or if it contains a plot node?

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: High
**Location**: `python/xorq/vendor/ibis/expr/types/core.py:950-958`
**Problem**: `is_plot` couples tag identity to binary dtype check. If `make_plot_expr` is called with a non-binary `return_type` (e.g., string for SVG), the expression carries `PLOT_TAG` but `is_plot` returns `False`. The tag is the semantic marker; the binary check adds a hidden constraint nowhere enforced or documented.
**Fix**: Drop the inner `any(dtype.is_binary() ...)` check. `bool(list(find_by_expr_tag(self.expr, PLOT_TAG)))` or `any(True for _ in find_by_expr_tag(self.expr, PLOT_TAG))` is sufficient.

---

**Severity**: Medium
**Location**: `python/xorq/common/utils/plot_utils.py:4`
**Problem**: `PLOT_TAG = "plot"` is a bare string. The codebase uses `StrEnum` for all tag constants (`FittedStepTagKey`, `FittedPipelineTagKey`, `CatalogInfix`). Also flagged in PR discussion.
**Fix**: Add a `StrEnum` (e.g., `class PlotTag(StrEnum): PLOT = "plot"`) and use `PlotTag.PLOT` as the tag value.

---

**Severity**: Medium
**Location**: `python/xorq/common/utils/plot_utils.py`, `core.py:948`
**Problem**: No tests for `make_plot_expr` or `is_plot`. Critical property with subtle binary-dtype behavior is entirely untested.
**Fix**: Add unit tests covering: tagged expression returns `True`, non-tagged returns `False`, non-binary return type.

---

**Severity**: Low
**Location**: `python/xorq/vendor/ibis/expr/types/core.py:948`
**Problem**: `is_plot` walks the entire expression graph and returns `True` if *any* node is plot-tagged, not just the root. An expression that wraps a plot node in further operations will still report `is_plot = True`. Likely unintended and inconsistent with how callers would read it.
**Fix**: Clarify intent — if checking the root only, use the root op's schema directly. If containment is intended, document it.

## Summary

Adds `make_plot_expr` (UDAF + tag helper) and `is_plot` property, but `is_plot`'s correctness depends on return type being binary rather than on the tag alone, making it silently broken for non-binary plot types.

</details>

---
*Review type: default | Agent: claude-code | Job: 14*

PR #1854, opened 2026-04-21
fix(worktree): preserve symlinks when copying envrc secrets

dlovell:
## Summary
- Use `cp -Pv` instead of `cp -v` in `dev/setup-worktree` so symlinks like `.envrc.secrets` → `~/.config/xorq/envrcs/.envrc.secrets` are preserved as symlinks rather than dereferenced into regular files
- Check `-e` instead of `-f` so symlinks are detected correctly
- Add `.claude` and `.envrcs/.worktree-symlinks` to `.gitignore.template` to keep worktree noise out of `git status`

## Test plan
- [x] Create a worktree and verify `.envrcs/.envrc.secrets` and `.envrcs/.envrc.user` are symlinks (not regular files)
- [x] Verify `git status` in the worktree is clean (no modifications to tracked files)
- [x] Verify `direnv allow` loads secrets correctly in the worktree

### DIFF
--- a/.gitignore.template
+++ b/.gitignore.template
@@ -1,6 +1,8 @@
 .gitignore
 /target
 
+.claude
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 .pytest_cache/
@@ -88,3 +90,4 @@ ci/ibis_testing.db
 .codspeed
 
 .envrc.secrets
+.envrcs/.worktree-symlinks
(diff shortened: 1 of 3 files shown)


.gitignore.template
```diff
@@ -1,6 +1,8 @@
 .gitignore
 /target
 
+.claude
```
mesejo, line 4: Is this intentional?

dlovell: yes

mesejo (approved):

PR #1875, opened 2026-04-26
fix(profiles): raise informative ValueError for unknown backend

dlovell:
## Summary
- Replace the bare `assert` in `Profile.validate_con_name` with a `ValueError` that names the rejected
backend and lists the installed ones.
- The assert failed silently under `-O` and gave no hint which backend was unavailable.

## Test plan
- [ ] Construct a `Profile` with an unknown `con_name` and confirm the error message lists installed
backends.
- [ ] Existing profile tests still pass.

### DIFF
--- a/python/xorq/tests/test_profile.py
+++ b/python/xorq/tests/test_profile.py
@@ -130,6 +130,11 @@ def test_profile_hash_order_independence():
     assert cloned.hash_name.split("_")[0] == profile1.hash_name.split("_")[0]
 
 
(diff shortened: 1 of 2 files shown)


mesejo:
I've run the critique this pr, and it gave me these two issues:
  1. Refactor to collect names in a single pass (eliminates the generator-exhaustion trap).                                                                                                                   
  2. Add a test for the new ValueError message.                    
I push a commit with those

mesejo (approved):

PR #1877, opened 2026-04-27
ADR-0009: Split data-dependent tokens out of normalize_op

dlovell:
## Summary

- Adds ADR-0009 describing the design for separating data-dependent leaf normalizations from structural expression normalization in `normalize_op`
- Enables cheap table substitution: swap one data dep hash, reuse cached structural token
- Defines a serializable metadata format (JSON) for token recomputation without importing xorq — only `hashlib` needed

Refs: XOR-304

## Test plan

- [ ] Review ADR for completeness and accuracy
- [ ] Validate prototype in `scripts/2026-04-27-pierre-sync.py` against the described mechanism

### DIFF
--- a/docs/adr/0009-split-normalize-op-data-from-structure.md
+++ b/docs/adr/0009-split-normalize-op-data-from-structure.md
@@ -1,0 +1,262 @@
+# ADR-0009: Split data-dependent tokens out of expression normalization
+
+- **Status:** Proposed
+- **Date:** 2026-04-27
+- **Deciders:** Dan, Pierre
+
+## Context
+
+`dask.base.tokenize(expr)` drives xorq's caching — if the token changes, the
+cache misses.  Today `normalize_op` (`dask_normalize_expr.py:713-740`) tangles
+data-dependent values (table contents) with structural values (SQL shape,
+schemas, UDFs) into a single flat tuple:
+
+1. `opaque_node_replacer` replaces opaque nodes (RemoteTable, Read, …) with
+   `UnboundTable(name=dask.base.tokenize(node))`, embedding content hashes as
+   table names in the generated SQL.
+2. `op.find(ir.DatabaseTable)` re-finds the same data nodes and includes their
+   normalizations again in the output tuple.
+
+Because the SQL string itself contains content hashes, there is no way to
+predict the token for a new table without re-running the full pipeline:
+expression compilation, `into_backend`, `op.replace`, SQL generation, and
+recursive normalization of every node.
+
+## Decision drivers
+
+- Substituting one table in a multi-table expression should cost O(size of new
+  table), not O(full normalization pipeline).
+- The structural part of a token (SQL shape, join predicates, filters, schemas)
+  must be stable across data changes so it can be computed once and reused.
+- No string substitution in SQL strings — the unbound expression should
+  naturally produce data-free SQL.
+- No backward compatibility required — a cache break is acceptable.
+- Token recomputation must be possible without importing xorq, given a
+  serialized metadata artifact and the new data dep hash.
+
+## Decision
+
+Replace `normalize_op` with a new implementation that unbinds the expression
+of its leaf data tables, normalizes the unbound (structural) expression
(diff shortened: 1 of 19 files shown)


docs/adr/0009-split-normalize-op-data-from-structure.md
```diff
@@ -0,0 +1,262 @@
+# ADR-0009: Split data-dependent tokens out of expression normalization
```
Barre, line 1: Wouldn't something like:

> Version the build-hash algorithm to separate data and structural tokens

be more accurate with regard to what we want to achieve?

Barre:
Could the same `(data_deps, structural)` split be done as a ~10-line patch to the existing `normalize_op`? Keep the `op.find(ir.DatabaseTable)` collection, just stop folding the data deps into the same tuple as the SQL/UDFs/etc.? That gets the substitution win and the no-xorq metadata format without `replace_nodes`, `UnboundTable`, or the new `normalize_op_split` helper.

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `.devcontainer/Dockerfile`, line 74 (diff)
- **Problem**: `ARG CONTAINER_WORKSPACE` has no default. The `ENV PATH="${CONTAINER_WORKSPACE}/.venv/bin:$PATH"` line evaluates at build time — building directly without `--build-arg CONTAINER_WORKSPACE=...` silently produces `PATH=/.venv/bin:...` instead of an error.
- **Fix**: Add `ARG CONTAINER_WORKSPACE=/workspaces/src` or fail early: `RUN [ -n "$CONTAINER_WORKSPACE" ] || (echo "CONTAINER_WORKSPACE required" && false)`.

---

- **Severity**: Medium
- **Location**: `.devcontainer/docker-compose.yml`, line ~35 (diff); `setup-claude.py`, `copy_user_prefs` line 49
- **Problem**: `${HOME}/.claude.json` bind-mounted into container. If that file doesn't exist on the host, Docker creates it as an empty file. `copy_user_prefs` then calls `json.load(f)` on an empty file → `JSONDecodeError`, crashing setup on first use by a developer who hasn't run Claude locally.
- **Fix**: Wrap `json.load` in `try/except json.JSONDecodeError: pass` in `copy_user_prefs`, `setup_settings`, and `setup_project_settings`.

---

- **Severity**: Low
- **Location**: `dev/devcontainer`, `setup_gh()`, line ~190 (diff)
- **Problem**: `docker cp "$hosts" "$(dc ps -q app)":/home/vscode/.config/gh/hosts.yml` — `dc ps -q app` may return multiple IDs if containers linger from a prior failed run. `image_stale()` guards with `head -1`; `setup_gh` does not.
- **Fix**: `docker cp "$hosts" "$(dc ps -q app | head -1)":/home/vscode/.config/gh/hosts.yml`

---

- **Severity**: Low
- **Location**: `.devcontainer/audit-report.py`, `main()`, line 13
- **Problem**: `sys.argv[1]` accessed unconditionally — `IndexError` if script is called with no arguments directly (e.g., during debugging or CI invocation).
- **Fix**: Add `if len(sys.argv) < 2: print("Usage: audit-report.py <audit_log> [--summary|--new|--all]", file=sys.stderr); sys.exit(1)`.

---

- **Severity**: Low
- **Location**: `docs/adr/0009-split-normalize-op-data-from-structure.md`, "Serializable metadata" section
- **Problem**: `compute_expr_token` replicates `dask.base.tokenize` by calling `hashlib.md5(str((tuple(hashes), structural_hash)).encode())`. This is correct today but depends on dask's internal string-normalization being identity and its tokenize output format staying stable. No version pin or assertion guards the assumption.
- **Fix**: Add a note that this must be re-verified when bumping dask, or add a one-liner test that `compute_expr_token` matches `dask.base.tokenize` for a known input.

## Summary

Adds a Docker-based devcontainer with Claude isolation, permission auditing, worktree support, and an ADR for splitting data-dependent tokens out of expression normalization.

</details>

---
*Review type: default | Agent: claude-code | Job: 12*

PR #1884, opened 2026-04-28
feat(dask_normalize): split expression tokens into data and structural components

Barre:
## Summary

Implements ADR-0010. Splits `normalize_op` into data-dependent and structural
components so expression cache tokens can be cheaply recomputed by substituting
only the changed data-leaf hash.

- `normalize_op` now returns `(data_deps, structural)` instead of a flat token
- New `normalize_op_split` exposes `(leaf_dts, data_deps, structural)` for
  callers that need to inspect or substitute individual data deps
- `_normalize_computed_kwargs_expr` is now data-free — `InMemoryTable` keyed
  by schema only, `Read` strips path identity, `CachedNode` keyed by schema +
  cache class — so ML-pipeline structural hashes are stable across data swaps
- New `expr_metadata` produces a serializable dict for cross-language token
  computation
- New `compute_expr_token` reproduces `dask.base.tokenize(expr)` from hex
  slot hashes + structural hash using only `hashlib` (no xorq/dask import)
- Opaque sub-expressions are placeholdered by their own structural token
  (computed with their own backend's compiler) via `_opaque_structural_name`,
  with per-call `contextvars` memoization capping recursion at O(depth)

## Test plan

- [x] Snapshot tests updated (expected cache break per ADR-0010)
- [x] `test_normalize_op_split_*` — three-tuple shape, structural stability,
  data sensitivity, subclass exclusion, cross-engine Read, InMemoryTable
- [x] `test_compute_expr_token_matches_dask_tokenize` — round-trip fidelity
- [x] `test_cheap_substitution_matches_full_recompute` — slot-swap equivalence
- [x] `test_normalize_computed_kwargs_expr_structural_is_data_free`
- [x] Benchmark tests for full tokenize, cached-structural path, and nested
  cross-engine pipelines


Barre:
```
case                  main full   PR full   PR cached   regression   speedup vs main
------------------------------------------------------------------------------------
  simple_filter_agg       1.34       1.41       0.89        +5.2%          1.5x
  pipeline_50_steps      37.01      41.39       0.89       +11.8%         41.8x

(times in ms)
```

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Medium
**Location**: `dask_normalize_expr.py:975-983` (`compute_expr_token`)
**Problem**: Correctness depends on dask internal hash mechanics — specifically that `_normalize_seq_func` wraps the registered handler's return in exactly one outer tuple and that hex strings are identity-normalized. A dask upgrade that changes either invariant silently breaks `compute_expr_token` without obvious test failures (test fixtures would also need refreshing).
**Fix**: Add a `# tested against dask==<version>` comment pinning the dask version this was validated against, and add a CI check that fails if dask is upgraded without re-running `test_compute_expr_token_matches_dask_tokenize`.

---

**Severity**: Medium
**Location**: `dask_normalize_expr.py:583-595` (`_normalize_computed_kwargs_expr`, TODO block)
**Problem**: `normalize_inmemorytable` still folds training `InMemoryTable` content into the structural hash for `FittedPipeline.predict` expressions. The TODO acknowledges this, but it means the core benefit of this PR (structural hash stable across data swaps) does not apply to the primary ML pipeline use case.
**Fix**: Track as a concrete follow-up issue; document in the ADR Consequences that ML pipelines are explicitly excluded from the optimization until the follow-up lands.

---

**Severity**: Low
**Location**: `dask_normalize_utils.py:120-128` (`patch_normalize_op_caching`) + diff
**Problem**: `patch_normalize_op_caching` wraps `normalize_op` with `functools.cache` and is used by `SnapshotStrategy.normalization_context`. After this PR, `normalize_op` returns `(data_deps, structural)` instead of a single token. Any code path that calls `normalize_op` directly (not through `normalize_expr`) within the patched context will receive a tuple instead of a token. All visible callers go through `normalize_expr` (which is updated), so no breakage today — but the `patch_normalize_op_caching` docstring still refers to memoizing for "tokenize" performance, which is now semantically misleading.
**Fix**: Update the comment in `strategy.py:123` to reflect the new return type.

---

**Severity**: Low
**Location**: `test_benchmark_dask_normalize.py:72-75`
**Problem**: The `cached()` benchmark uses `dask.base.normalize_token(dt)` instead of `_normalize_data_leaf(dt)`. For `InMemoryTable` leaves, these differ — `_normalize_data_leaf` routes through `normalize_inmemorytable`, while `dask.base.normalize_token` would use the default handler. The benchmark therefore measures a path that does not match what production code does for InMemoryTable-backed expressions.
**Fix**: Use `_normalize_data_leaf(dt)` instead of `dask.base.normalize_token(dt)` in the `cached()` function, consistent with `normalize_op_split`'s implementation.

---

**Severity**: Low
**Location**: `docs/adr/0009-split-normalize-op-data-from-structure.md`, Consequences > Negative
**Problem**: Contradiction: Negative Consequences says "Callers should look up slots by table name, not by hardcoded index," but the Decision section says the opposite — "look up by `index` (unambiguous) — name lookup is not safe when multiple slots share a name."
**Fix**: Remove the contradictory sentence from Negative Consequences.

---

## Summary

Splits `normalize_op` into data-dependent and structural components so expression cache tokens can be cheaply recomputed by substituting only the changed data-leaf hash, with a new `compute_expr_token` helper enabling token recomputation from serialized metadata without importing xorq.

</details>

---
*Review type: default | Agent: claude-code | Job: 11*

mesejo:
## roborev: Fail

<details>

## Review Findings

---

**Severity**: Medium  
**Location**: `dask_normalize_expr.py` — `normalize_read` (new code replacing lines 407–409)  
**Problem**: The old guard `if (read_path := read_kwargs.get("read_path")) is not None: tpls = (("read_path", str(read_path)),)` handled catalog-stamped reads where `read_path` is set and `hash_path` is still an absolute path (before YAML-roundtrip rewrite). The new elif `not pathlib.Path(path).is_absolute() and path == read_kwargs.get("read_path")` only fires when `hash_path` has already been rewritten to the relative `read_path`. Any in-process `Read` constructed with `read_path` but an absolute `hash_path` (e.g. during a build step before `ibis_yaml/common.py` rewrites it) now silently falls through to content-hash via `normalize_method(path)` instead of the stable `read_path`-keyed token. The change in `caching/strategy.py:45` (`path = read_kwargs.get("read_path") or read_kwargs["hash_path"]`) shows the broader codebase still prefers `read_path` when present; the tokenizer no longer matches that preference.  
**Fix**: Either keep the old early-exit for `read_path is not None` and add the new `build-relative-path` branch *inside* that arm, or document explicitly that `read_path`-without-rewrite can never reach `normalize_read`.

---

**Severity**: Medium  
**Location**: `dask_normalize_expr.py` — `normalize_op` (lines ~710–720 in new code)  
**Problem**: `normalize_op` previously returned a single token. It now returns `(data_deps, structural)`. `patch_normalize_op_caching` in `dask_normalize_utils.py` wraps `normalize_op` with `functools.cache` and callers use it via `patched_tokenize → dask.base.tokenize → normalize_expr`, so that specific path works. But the function signature is presented as backward-compatible while the return type is a breaking change. Any direct caller of `normalize_op` expecting a hex string (e.g. external plugins, test fixtures not shown) silently gets a tuple.  
**Fix**: Rename the old entry point `normalize_op` → `normalize_op_raw` or add an explicit deprecation wrapper that errors on direct call until all callers migrate.

---

**Severity**: Low  
**Location**: `dask_normalize_expr.py` — `opaque_node_replacer` default case (~line 800)  
**Problem**: All explicit match arms use `api.table(node.schema, name=...).op()`. The new data-leaf default branch uses `ir.UnboundTable(name=node.name, schema=node.schema)` directly. If `node.schema` is an ibis `Schema` object vs. a dict/tuple, the two constructors may accept different types or perform different validation, silently producing a mismatched op.  
**Fix**: Use `api.table(node.schema, name=node.name).op()` to match the pattern used everywhere else.

---

**Severity**: Low  
**Location**: `dask_normalize_expr.py` — `_opaque_structural_name` (~line 695)  
**Problem**: Memo key is `sub_op` alone, not `(sub_op, compiler)`. If the same op is encountered in two different compilation contexts (e.g. two `RemoteTable` boundaries backed by different compilers reaching a shared sub-expression), the first compiler's structural token is returned for all subsequent lookups, silently producing an incorrect structural hash.  
**Fix**: Key the memo as `(sub_op, id(compiler))` or `(sub_op, type(compiler).__name__)`.

---

**Severity**: Low  
**Location**: `test_benchmark_dask_normalize.py:74`  
**Problem**: `cached()` computes `dask.base.tokenize(dask.base.normalize_token(dt))` — applies normalization twice. `normalize_token(dt)` returns a raw tuple; `tokenize(that_tuple)` calls `normalize_token` on the tuple again before hashing. The intended cheap path is `tokenize(dt)` directly. Both produce the same hash only if `normalize_token` is idempotent on its own output, which holds for primitive-typed returns but is an implicit assumption.  
**Fix**: Replace with `dask.base.tokenize(dt)`.

---

## Summary

Splits `normalize_op` into separate data-dependency and structural components to enable O(new-table-size) cache-token recomputation when substituting one table in a multi-table expression, replacing the old single-flat-token approach that embedded content hashes in generated SQL.

</details>

---
*Review type: default | Agent: claude-code | Job: 28*

mesejo:
## roborev: Fail

<details>

## Review Findings

---

**Severity**: Medium
**Location**: `python/xorq/common/utils/tests/test_benchmark_dask_normalize.py:72-77`
**Problem**: The `cached()` benchmark computes slot hashes as `dask.base.tokenize(dask.base.normalize_token(dt))`, but the actual path in `normalize_expr` is `dask.base.tokenize(_normalize_data_leaf(dt))`. For `InMemoryTable` leaves, these diverge: `_normalize_data_leaf` routes through `normalize_inmemorytable` explicitly (no registered handler on InMemoryTable), while `normalize_token` falls back to dask's default tokenizer. The three benchmark cases all use `DatabaseTable` leaves so no test fails, but the benchmark never exercises the InMemoryTable path and would produce wrong tokens for InMemoryTable-based expressions if run there.
**Fix**: Replace `dask.base.normalize_token(dt)` with `_normalize_data_leaf(dt)` in the `cached()` closure (import it at test-module level).

---

**Severity**: Low
**Location**: `python/xorq/common/utils/tests/test_dask_normalize.py` (`test_normalize_op_split_includes_inner_read_under_remote_table`)
**Problem**: `NamedTemporaryFile(delete=False)` is used without cleanup. The temp parquet file persists on disk after the test.
**Fix**: Wrap in `try/finally: os.unlink(path)` or use `tmp_path` pytest fixture instead.

---

**Severity**: Low
**Location**: `docs/adr/0009-split-normalize-op-data-from-structure.md:4`
**Problem**: ADR status is "Proposed" but the change is being implemented in the same PR.
**Fix**: Set status to "Accepted" or "Implemented".

---

**Severity**: Low
**Location**: `python/xorq/common/utils/dask_normalize/dask_normalize_expr.py` (`compute_expr_token`)
**Problem**: The function depends on dask's undocumented internal tokenization format — that hex strings are identity-normalized and `_normalize_seq_func` wraps the handler return in a single outer tuple. Dask is pinned to `==2025.1.0` but the function contains no inline reference to that pin. A future pin bump could silently break token recomputation in external consumers (the `test_compute_expr_token_matches_dask_tokenize` test guards against breakage only if it is re-run after the upgrade).
**Fix**: Add a comment naming the dask version this behavior was validated against, e.g. `# Verified against dask==2025.1.0`.

---

## Summary

Splits `normalize_op`'s token into `(data_deps, structural)` components so the structural hash can be cached and only data-leaf hashes need recomputing when input tables change, enabling O(new-table) cache-key updates instead of O(full pipeline).

</details>

---
*Review type: default | Agent: claude-code | Job: 65*

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Medium
**Location**: `dask_normalize_expr.py` — `compute_expr_token`
**Problem**: The function hardcodes dask's internal `_normalize_seq_func` wrapping formula: `str(((tuple(data_dep_hashes), structural_hash),))`. Any dask version that changes how it wraps registered-handler output will silently produce wrong cache keys — no xorq error, just cache misses/hits on wrong expressions.
**Fix**: The test `test_compute_expr_token_matches_dask_tokenize` guards this at test time, but the docstring should call out the exact dask version constraint (and CI should pin dask accordingly). Consider also adding a version sentinel to `expr_metadata` so deployed systems can detect formula drift.

---

**Severity**: Medium
**Location**: `dask_normalize_expr.py` — new `normalize_read`, lines ~406–450 (post-diff)
**Problem**: The `read_path` priority branch is removed. Build-bundled `Read` nodes (created by `compiler.py:596-607`) have both `read_path` (stable catalog key) and `hash_path` (actual temp path). The old code used `read_path` as a fast stable token without reading the file; the new code falls through to `normalize_method(hash_path)`, which reads the file. If `hash_path` points to a file that doesn't exist at tokenize time (e.g., planning a token before the build outputs are written), the new code raises `NotImplementedError` where the old code succeeded.
**Fix**: Document the new invariant explicitly — that `hash_path` must exist on disk whenever `normalize_read` is called — or add a fallback to `read_path` when `hash_path` doesn't exist.

---

**Severity**: Low
**Location**: `dask_normalize_expr.py` — `_normalize_computed_kwargs_expr`, `read_structural` tuple
**Problem**: `r.normalize_method` (a function object) is included alongside `r.method_name` (a string). `method_name` already uniquely identifies the normalization method; the function object is redundant and its dask tokenization is less portable (may vary if it's a closure or across process restarts).
**Fix**: Remove `r.normalize_method` from the structural tuple; `r.method_name` is sufficient.

---

**Severity**: Low
**Location**: `test_dask_normalize.py` — no test for `normalize_read` with absolute `read_path`
**Problem**: The behavioral change (from `read_path`-as-stable-key to `normalize_method`-on-`hash_path`) is not exercised by any test. The only new `normalize_read`-adjacent test covers `test_normalize_op_split_includes_inner_read_under_remote_table` but not the catalog/build-bundled path.
**Fix**: Add a test that creates a `Read` with both `read_path` and `hash_path` set, verifies the token is stable, and documents the new contract.

## Summary

Splits `normalize_op` into structural and data-dependent components so expression cache keys can be recomputed cheaply by substituting only the changed data slots, with a new `compute_expr_token` API for cross-language token recomputation without importing xorq.

</details>

---
*Review type: default | Agent: claude-code | Job: 66*

dlovell (approved):

PR #1888, opened 2026-04-28
docs: add tutorial for composing unbound transforms with local sources

paddymul:
## Summary
- Adds a new tutorial at `docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd` that walks readers through cloning an `internal_tool_catalog`, writing an on-the-fly `deferred_read_csv` source, and binding the two with `xorq catalog compose` to produce a new cataloged result.
- Frames the catalog as a shared toolbox of generically applicable transforms, with the source written locally and the smoothing UDWF authored once and reused.
- Registers the tutorial under the **Analytics tutorials** section in `docs/_quarto.yml`.

The cataloged transform mirrors the exponential smoothing pattern from `examples/python_udwf.py` (`pyarrow_udwf` + `UnboundTable`).

## Test plan
- [ ] Render the docs locally (`quarto preview docs`) and confirm the new entry appears in the **Analytics tutorials** sidebar.
- [ ] Verify the page renders without errors (no broken cross-references, fenced code blocks display correctly).
- [ ] Spot-check that the CLI snippets (`xorq catalog clone`, `xorq catalog compose`, `xorq catalog run`, `xorq build`) match current command signatures.

### DIFF
--- a/docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
+++ b/docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
@@ -206,4 +206,9 @@ You separated the two halves of a pipeline along a clean seam:
 
 `xorq catalog compose` is the verb that joins them, and the result is itself a catalog entry — meaning the next person can pick it up where you left off.
 
+:::{.callout-note}
+### Runnable counterpart
+[`examples/compose_unbound_with_source.py`](https://github.com/xorq-labs/xorq/blob/main/examples/compose_unbound_with_source.py) reproduces the binding in pure Python — it builds the unbound `exp_smooth` transform, points a `deferred_read_csv` at a generated CSV, and runs the same source-replacement that `xorq catalog compose` performs internally. It's part of the test suite, so the code stays in sync with the tutorial.
+:::
+
 The catalog is where you store your internal company toolbox of transforms that are generically applicable.
--- a/docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
+++ b/docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
@@ -1,0 +1,209 @@
+---
+title: 'Compose a cataloged transform with a local CSV'
+description: "Pull a reusable unbound transform from a shared catalog and apply it to data on your laptop"
+icon: "diagram-3"
+headline: "Reuse transforms across pipelines"
+---
+
+This tutorial shows you how to take an unbound expression that lives in a shared catalog and apply it to a source expression you write on the fly. You will clone an `internal_tool_catalog`, point a `deferred_read_csv` at a CSV on your laptop, and then run `xorq catalog compose` to bind the two together into a new cataloged result.
+
+After completing this tutorial, you will know how to treat the catalog as a library of reusable transforms and pull them into ad-hoc pipelines.
+
+## Prerequisites
+
+You need:
+
+- Xorq installed: `pip install xorq`
+- A working `git` install (the catalog is a git repo)
+- A CSV file on your machine to smooth
+
+## What you will build
+
+Two pieces, composed:
+
+- **The unbound transform** — an exponential smoothing window function lives in `internal_tool_catalog` as an `UnboundExpr`. It declares an input schema with one float column named `a` and applies a UDWF to smooth it.
+- **The source** — a `deferred_read_csv` of a CSV on your laptop, written in a small Python file. This is the "on the fly" half of the composition.
+
+`xorq catalog compose` glues them together: it replaces the unbound table inside the transform with your source and persists the result back into the catalog as a new entry.
+
+## Step 1 — Clone the internal tool catalog
+
+The catalog is just a git repo, so cloning it is a one-liner. Replace the URL with the one your team uses internally.
+
+```bash
+xorq catalog clone https://github.com/your-org/internal_tool_catalog.git -n internal_tool_catalog
+```
+
+The `-n internal_tool_catalog` flag stores the clone under the name `internal_tool_catalog` so you can refer to it without typing a path. Make it the default for the rest of the session:
+
+```bash
+xorq catalog default --set internal_tool_catalog
+```
+
+Now confirm the cataloged transform is present:
+
+```bash
+xorq catalog list-aliases
+```
+
+You should see `exp-smooth` (the unbound exponential smoothing transform) in the output.
+
+Inspect its schema to see what kind of source it expects:
+
+```bash
+xorq catalog schema exp-smooth
+```
+
+```
+Type: Partial (unbound)
+
+Schema In:
+  a                        float64
+
+Schema Out:
+  a                        float64
+  exp_smooth               float64
+```
+
+The transform is unbound: it declares that it needs a table with a float column `a` and produces that same column plus a smoothed `exp_smooth` column. Any source whose schema is a superset of `Schema In` can be composed with it.
+
+:::{.callout-note}
+### What's inside the catalog
+The `exp-smooth` entry was authored once and added to the catalog with `xorq catalog add`. The Python it was built from looks like the snippet below — a `pyarrow_udwf` decorator applied to an `UnboundTable` whose schema declares the input contract. You don't need to write this yourself; it's already cataloged.
+
+```python
+import pyarrow as pa
+import xorq.api as xo
+from xorq.expr.udf import pyarrow_udwf
+from xorq.vendor import ibis
+from xorq.vendor.ibis.expr import operations as ops
+
+
+@pyarrow_udwf(
+    schema=ibis.schema({"a": float}),
+    return_type=ibis.dtype(float),
+    alpha=0.9,
+)
+def exp_smooth(self, values: list[pa.Array], num_rows: int) -> pa.Array:
+    results = []
+    curr_value = 0.0
+    arr = values[0]
+    for idx in range(num_rows):
+        if idx == 0:
+            curr_value = arr[idx].as_py()
+        else:
+            curr_value = (
+                arr[idx].as_py() * self.alpha + curr_value * (1.0 - self.alpha)
+            )
+        results.append(curr_value)
+    return pa.array(results)
+
+
+unbound = ops.UnboundTable(
+    name="placeholder", schema=ibis.schema({"a": float})
+).to_expr()
+
+expr = unbound.mutate(exp_smooth=exp_smooth.on_expr(unbound))
+```
+
+`expr` is an `UnboundExpr`: it has the shape of a transform but no source attached. That's the property the catalog records, and it's what makes the entry reusable across many sources.
+:::
+
+## Step 2 — Point a source expression at your local CSV
+
+Now write the source half on the fly. Create a file called `local_source.py` next to the CSV you want to smooth:
+
+```python
+# local_source.py
+import xorq.api as xo
+
+# <1>
+expr = xo.deferred_read_csv("readings.csv")
+```
+1. `deferred_read_csv` produces a `Source` expression bound to your local file. No data is read yet — this is just a description of where to read from.
+
+If your CSV doesn't already have a float column named `a`, add a `select`/`mutate` to line up the schema with what `exp-smooth` expects:
+
+```python
+# local_source.py
+import xorq.api as xo
+
+expr = (
+    xo.deferred_read_csv("readings.csv")
+    .mutate(a=xo._.value.cast("float64"))
+    .select("a")
+)
+```
+
+The schema only needs to be a superset of `{a: float64}` for the compose to validate.
+
+## Step 3 — Build the source and add it to the catalog
+
+`xorq catalog compose` only operates on cataloged entries, so add this one-off source first. Build it:
+
+```bash
+xorq build local_source.py -e expr
+```
+
+```
+Building expr from local_source.py
+Written 'expr' to builds/9f1c2a4e8b30
+builds/9f1c2a4e8b30
+```
+
+Then add the build to the catalog under a friendly alias:
+
+```bash
+xorq catalog add builds/9f1c2a4e8b30 --alias my-readings
+```
+
+`my-readings` is now a `Source` entry in `internal_tool_catalog`, sitting next to the `exp-smooth` unbound transform.
+
+## Step 4 — Compose them with `xorq catalog compose`
+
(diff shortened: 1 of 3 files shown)


paddymul (commented):
Adversarial review — poking holes. Comments inline. Two cross-cutting concerns that don't attach to a single line:

- **Quarto doesn't execute the snippets.** `_quarto.yml` sets `execute.eval: false`, so the prose CLI flow and inline Python are never mechanically verified at doc-build time. Combined with the runnable example only covering `replace_unbound` (see inline note on the runnable), the tutorial has no executable contract.
- **PR test plan is entirely manual checkbox items.** Render docs locally, eyeball the sidebar, eyeball the page, spot-check CLI signatures. Nothing enforced in CI.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+- **The source** — a `deferred_read_csv` of a CSV on your laptop, written in a small Python file. This is the "on the fly" half of the composition.
+
+`xorq catalog compose` glues them together: it replaces the unbound table inside the transform with your source and persists the result back into the catalog as a new entry.
+
+## Step 1 — Clone the internal tool catalog
+
+The catalog is just a git repo, so cloning it is a one-liner. Replace the URL with the one your team uses internally.
+
+```bash
+xorq catalog clone https://github.com/your-org/internal_tool_catalog.git -n internal_tool_catalog
```
paddymul, line 34: **Tutorial is unrunnable end-to-end.** This URL is fictitious and there is no public `exp-smooth` cataloged entry anywhere a reader could swap in. The 'replace with your team's URL' handwave only works for an internal audience that already has a catalog and the entry. A new reader gets stuck on step 1 with no fallback.

Either point at a real public catalog (or a `git init`-able local example), or reframe step 1 explicitly as 'assuming your org has a catalog with `exp-smooth`.'

**Nit on the same line:** `-n` only sets the *path* (`Catalog.name_to_repo_path(dest_name)`); the tutorial implies it also 'stores the clone under the name.' True in effect, but the user still has to `default --set` to make subsequent commands implicit. The two-step relationship is glossed over.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+            )
+        results.append(curr_value)
+    return pa.array(results)
+
+
+unbound = ops.UnboundTable(
+    name="placeholder", schema=ibis.schema({"a": float})
+).to_expr()
+
+expr = unbound.mutate(exp_smooth=exp_smooth.on_expr(unbound))
```
paddymul, line 106: **Ordering fragility is baked into the cataloged transform.** No `order_by` on `unbound.mutate(exp_smooth=exp_smooth.on_expr(unbound))`. Anyone composing this against arbitrary data gets engine-defined row ordering, and `exp_smooth` is order-dependent. The tutorial frames `exp-smooth` as 'generically applicable' — it isn't, unless the source is pre-ordered. Worth a callout, or fold an `.order_by(...)` into the cataloged definition.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+1. `deferred_read_csv` produces a `Source` expression bound to your local file. No data is read yet — this is just a description of where to read from.
+
+If your CSV doesn't already have a float column named `a`, add a `select`/`mutate` to line up the schema with what `exp-smooth` expects:
+
+```python
+# local_source.py
+import xorq.api as xo
+
+expr = (
+    xo.deferred_read_csv("readings.csv")
```
paddymul, line 132: **Phantom `value` column.** `xo._.value.cast("float64")` references a column named `value` that was never introduced. The example CSV (per `examples/compose_unbound_with_source.py`) has only column `a`, so this branch wouldn't be needed. Either anchor this to a concrete sample CSV with a non-`a` column or drop the variant.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+`xorq catalog compose` only operates on cataloged entries, so add this one-off source first. Build it:
+
+```bash
+xorq build local_source.py -e expr
+```
+
+```
+Building expr from local_source.py
+Written 'expr' to builds/9f1c2a4e8b30
+builds/9f1c2a4e8b30
```
paddymul, line 151: **Hand-written build hash is a footgun.** `9f1c2a4e8b30` is invented and non-deterministic. Readers who copy-paste blindly into the next step (`xorq catalog add builds/9f1c2a4e8b30 ...`) hit 'no such directory.' Use a placeholder like `<BUILD_HASH>` or shell-capture:

```bash
BUILD=$(xorq build local_source.py -e expr | tail -1)
xorq catalog add "$BUILD" --alias my-readings
```

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+```
+
+`my-readings` is now a `Source` entry in `internal_tool_catalog`, sitting next to the `exp-smooth` unbound transform.
+
+## Step 4 — Compose them with `xorq catalog compose`
+
+This is the step the rest of the tutorial was setting up. Pass the source first, then the transform:
+
+```bash
+xorq catalog compose my-readings exp-smooth --alias smoothed-readings
```
paddymul, line 167: **Failure mode for swapped argument order is undocumented.** Tutorial says 'pass the source first, then the transform.' If a reader reverses it (`xorq catalog compose exp-smooth my-readings`), `_compose_expr` (`python/xorq/catalog/cli.py:715`) does `source_entry, *transform_entries = catalog_entries.values()` and treats the unbound transform as the source — `_resolve_source` then fails with an opaque error. One sentence noting 'the first positional must be a bound source' would save a debugging round.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+```
+
+```
+Cataloged as 'smoothed-readings'
+```
```
paddymul, line 172: **Nit:** this 'Cataloged as ...' line is written to **stderr** (`python/xorq/catalog/cli.py:852`), not stdout. If a reader pipes the output expecting to capture the message, they won't.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+
+`compose` walked the unbound transform, replaced its `UnboundTable` with `my-readings`, validated that the schemas line up, built the resulting expression, and persisted the build back into the catalog as `smoothed-readings`.
```
paddymul, line 174: **Overstates what `compose` does.** 'validated that the schemas line up' is wrong — `_compose_expr` (`python/xorq/catalog/cli.py:690`) doesn't do explicit schema validation. `replace_unbound` matches the single `UnboundTable` and trusts the rest. A schema mismatch surfaces later as a less friendly build/execute error. Either show what that error actually looks like or soften the claim.

docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd
```diff
@@ -0,0 +1,214 @@
...
+You will see your original `a` column alongside the new `exp_smooth` column, smoothed with the `alpha=0.9` baked into the cataloged transform.
+
+## What you just did
+
+You separated the two halves of a pipeline along a clean seam:
+
+- The **transform** — exponential smoothing — was authored once, reviewed once, and dropped into a shared catalog. Anyone on your team can pull it down and apply it.
+- The **source** — your local CSV — stayed local and ad-hoc. You did not have to fork the catalog or PR a new pipeline to use it.
+
+`xorq catalog compose` is the verb that joins them, and the result is itself a catalog entry — meaning the next person can pick it up where you left off.
```
paddymul, line 207: **'Shareable catalog entry' framing is misleading.** The composed `smoothed-readings` entry holds a `deferred_read_csv` to your local `readings.csv`. A teammate who pulls the catalog can't execute it — the file isn't on their machine. The local-source-meets-shared-transform pattern doesn't yield a shareable composed result; it yields a personal one. Either qualify this paragraph or change the example to a path under the repo / a public URL.

examples/compose_unbound_with_source.py
```diff
@@ -0,0 +1,123 @@
...
+Companion runnable for
+``docs/tutorials/analytics_tutorials/compose_unbound_with_source.qmd``.
+
+The tutorial frames this as cloning an ``internal_tool_catalog`` (which holds
+the unbound ``exp-smooth`` transform) and binding it against a CSV on disk
+using ``xorq catalog compose``. This script reproduces the underlying
+mechanics in pure Python: it builds the unbound exponential-smoothing
+transform, points a ``deferred_read_csv`` at a temp CSV, and replaces the
+``UnboundTable`` placeholder with that source — which is what
+``xorq catalog compose`` does internally before persisting the result.
```
paddymul, line 12: **The 'runnable counterpart' doesn't verify what the tutorial claims.** PR description says 'the code stays in sync with the tutorial.' It doesn't. This script only exercises `replace_unbound` in pure Python. None of the CLI surfaces the tutorial documents — `xorq catalog clone`, `default`, `list-aliases`, `schema`, `add`, `compose`, `run` — are touched by the test. If any of those drift, the tutorial silently rots and this test still passes.

A runnable that drives the actual CLI via `subprocess` (against a temp catalog) would give the doc a real contract.

examples/compose_unbound_with_source.py
```diff
@@ -0,0 +1,123 @@
...
+
+import pyarrow as pa
+
+import xorq.api as xo
+from xorq.common.utils.graph_utils import replace_unbound
+from xorq.expr.udf import pyarrow_udwf
+from xorq.ibis_yaml.enums import ExprKind
+from xorq.vendor import ibis
+from xorq.vendor.ibis.expr import operations as ops
+from xorq.vendor.ibis.expr.types.core import ExprMetadata
```
paddymul, line 27: **Runnable leans on private internals.** `replace_unbound` (`xorq.common.utils.graph_utils`), `ExprKind` (`xorq.ibis_yaml.enums`), and `ExprMetadata` (`xorq.vendor.ibis.expr.types.core`) are not exported via `xorq.api`. If the tutorial's purpose is to show users a public pattern, a runnable using only `xorq.api` (or driving the CLI) would be more honest about what's stable surface vs. internals.

examples/compose_unbound_with_source.py
```diff
@@ -0,0 +1,123 @@
...
+    transform_meta = ExprMetadata.from_expr(transform_expr)
+    assert transform_meta.kind == ExprKind.UnboundExpr
+    assert dict(transform_meta.schema_in) == {"a": ibis.dtype("float64")}
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        csv_path = make_local_csv(Path(tmpdir))
+        source_expr = make_source_expr(csv_path)
+
+        composed = compose(source_expr, transform_expr)
+        result = xo.execute(composed).sort_values("a").reset_index(drop=True)
```
paddymul, line 110: **Order-fragile test that only happens to pass.** The UDWF is order-dependent and there's no `order_by` on the partition. The assertion only holds because the test data `[1.0, 2.1, 2.9, 4.0, 5.1, 6.0, 6.9, 8.0]` is monotonically increasing, so `sort_values("a")` is a no-op. Change the data to anything non-monotonic (e.g. `[3.0, 1.0, 2.0, ...]`) and the per-row `exp_smooth` assertion fails — smoothed values are computed in scan order, but the comparison uses sorted order.

Fix: add an explicit `.order_by("a")` to the composed expression, or assert order-independent properties (sum, count, set membership).

examples/compose_unbound_with_source.py
```diff
@@ -0,0 +1,123 @@
...
+    assert result["a"].tolist() == expected_a
+    assert result["exp_smooth"].round(6).tolist() == [
+        round(v, 6) for v in expected_smoothed(expected_a)
+    ]
+
+
+if __name__ == "__main__":
+    main()
+elif __name__ == "__pytest_main__":
+    main()
```
paddymul, line 122: **Nit:** `pytest_examples_passed` is set only on the `__pytest_main__` branch. Fine, but if `main()` raises, the pytest assertion just reports the flag missing with no diagnostic about which assertion in `main()` failed. A `try/except` wrapping `main()` and re-raising with context (or just letting the inner `AssertionError` propagate explicitly) would help debuggability.

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Low
- **Location**: `examples/compose_unbound_with_source.py:105`
- **Problem**: `result["a"].tolist() == expected_a` uses exact float equality on CSV-round-tripped floats. Values like `2.1`, `2.9`, `6.9` are not exactly representable in float64. Works in practice because Python float literals and Arrow CSV parsing both land on the same bit pattern, but it's fragile. The sibling example `python_udwf.py` uses `assert_series_equal` (which has tolerance) for the same data.
- **Fix**: `pd.testing.assert_series_equal(result["a"], pd.Series(expected_a, name="a"), check_exact=False)` or at minimum `pytest.approx`.

---

- **Severity**: Low
- **Location**: `examples/compose_unbound_with_source.py:82` / `compose_unbound_with_source.qmd` inline snippet
- **Problem**: `exp_smooth.on_expr(unbound)` is applied without `.over(ibis.window(order_by=...))`. Smoothing is order-sensitive; SQL gives no row-order guarantee without ORDER BY. The test relies on DataFusion preserving CSV insertion order (which it does for small files, and is consistent with `python_udwf.py`), but this is an implicit assumption, not enforced by the expression.
- **Fix**: Add `.over(ibis.window(order_by="a"))` to make the ordering contract explicit and match what the doc text implies.

## Summary

Adds a tutorial and runnable Python companion demonstrating how to compose a cataloged `UnboundExpr` exponential-smoothing transform with a local `deferred_read_csv` source via `replace_unbound`.

</details>

---
*Review type: default | Agent: claude-code | Job: 9*

PR #1878, opened 2026-04-27
docs(tutorials): add BSL semantic catalog + catalog collaboration tutorials

paddymul:
## Summary

Adds two new core tutorials covering the BSL + catalog story end-to-end:

- **\`build_a_semantic_catalog.qmd\`** — define a Boring Semantic Layer model over an FAA-style flights memtable, query dimensions/measures, and round-trip the model through a local catalog using \`to_tagged(model)\` / \`from_tagged(entry.expr)\`. Includes side-by-side callouts comparing BSL queries to plain xorq, and showing how an unknown dimension fails fast.
- **\`working_with_the_catalog.qmd\`** — picks up where the foundation tutorial leaves off and walks the catalog lifecycle: pushing it to GitHub, a teammate cloning and proposing a new entry via PR, pulling the merged change back, and swapping the connection profile at recovery time with \`catalog.load(name, con=...)\`. Recommends git-annex throughout (with platform install commands and a callout on S3/GCS/directory special remotes for content storage).

The naming is consistent across both tutorials: \`flights_model\` (live SemanticModel), \`flights_model_expr\` (catalog-ready tagged expression), \`flights_entry\` (catalog handle).

## Test plan

- [x] Foundation tutorial: ran the full \`flights_catalog.py\` script in a clean \`uv venv\` with \`pip install \"xorq[bsl,duckdb]\"\` — every output block in the tutorial is real run output (memtable round-trip via parquet + duckdb).
- [x] Collaboration tutorial: verified the underlying API surface end-to-end against a local bare-repo stand-in for GitHub: \`Catalog.from_repo_path(..., init=True)\`, \`Catalog.clone_from\`, \`catalog.add\`, \`catalog.pull\`, \`catalog.add_alias\`, \`Profile.from_con/save/load\`, \`catalog.load(con=...)\`. Real FAA flights parquet (≈345k rows) recovered through the catalog round-trip and queried correctly.
- [ ] Annex-enabled flow not re-verified end-to-end after the prose switched to recommending git-annex throughout. The annex APIs used in the tutorial (\`annex=LOCAL_ANNEX\`, \`catalog.fetch_entries(...)\`, \`git annex copy --to=origin\`) match the documented surface in \`python/xorq/catalog/{catalog,annex,backend}.py\`, but a full clone-fetch-recover-push round trip with annex on a bare server hasn't been replayed against the new prose.

## Notes for review

[...]

Issue #25, opened 2024-05-03
Expose the TableProvider to Python


mesejo:
Exposing the TableProvider to Python will allow us to debug the code inside the TableProvider

Issue #23, opened 2024-04-29
Implement addition of tables for sql execution


mesejo:
Currenlty this fails:

```python
    ddb = ibis.duckdb.connect()
    ddb.read_csv(csv_dir / "awards_players.csv", table_name="ddb_players")
    con.add_connection(ddb)

    left = batting[batting.yearID == 2015]
    right = awards_players[awards_players.lgID == "NL"].drop("yearID", "lgID")
    right_df = right.execute()
    left_df = left.execute()
    predicate = ["playerID"]
    result_order = ["playerID", "yearID", "lgID", "stint"]

    ddb_players = con.table("ddb_players")
    right = ddb_players[ddb_players.lgID == "NL"].drop("yearID", "lgID")
    expr = left.join(right, predicate, how="inner")
    query = ibis.to_sql(expr, dialect="datafusion")

    result = (
        con.sql(query).execute()
        .fillna(np.nan)[left.columns]
        .sort_values(result_order)
        .reset_index(drop=True)
    )

    expected = (
        check_eq(
            left_df,
            right_df,
            how="inner",
            on=predicate,
            suffixes=("_x", "_y"),
        )[left.columns]
        .sort_values(result_order)
        .reset_index(drop=True)
    )

    assert_frame_equal(result, expected, check_like=True)
    ddb.drop_view("ddb_players")
```

with error:

> Exception: DataFusion error: Plan("table 'datafusion.public.batting' not found")

PR #1890, opened 2026-04-28
docs(tutorials): add "Deploy a pipeline on a schedule" core tutorial

paddymul:
## Summary
- Adapts the deploying-xorq blog post ([xorq-website#23](https://github.com/xorq-labs/xorq-website/pull/23)) into a Quarto tutorial under `docs/tutorials/core_tutorials/deploy_a_pipeline.qmd`.
- Walks through deploying a xorq expression to a server with cron: catalog-as-submodule layout, env-var-based profile resolution, the wrapper script, parquet-to-disk + DB loader pattern, and run-log based monitoring.
- Adds a new "Core tutorials" sidebar section in `docs/_quarto.yml` exposing just the new file (other `core_tutorials/*.qmd` files in main remain hidden — separate decision).

## Notes for reviewers
- Uses `xorq catalog ... run <alias>` rather than the original blog's "resolve alias to a path then `xorq run`" pattern, to match what the catalog CLI actually exposes today.
- Drops the blog's "rough edges, named" personal-commentary section and softens the dbt comparison to keep tutorial tone instructional.
- Verified the file renders with `quarto render` locally.

## Test plan
- [ ] Render the docs site (`quarto preview docs`) and confirm the tutorial appears under "Core tutorials" in the sidebar.
- [ ] Check that all internal links (`your_first_build.qmd`, `working_with_the_catalog.qmd`, `explore_caching.qmd`, `switch_backends.qmd`) resolve when the site is built.
- [ ] Spot-check the example pipeline code compiles against current xorq (`xorq.api`, `flight_udxf`, `xo.param`, `xo.postgres.connect_env`).

### DIFF
--- a/docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
+++ b/docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
@@ -5,7 +5,7 @@ icon: "rocket-takeoff"
 headline: "From `xorq run` to a cron line"
 ---
 
-You have a xorq expression. It pulls data from somewhere, joins it against a table, computes a few derived columns, and aggregates to a result. You can run it locally with `xorq run` and the numbers look right. The next sentence out of your mouth is: *"I need this to run every morning on a server."*
+You have a xorq expression. It pulls data from somewhere, joins it against a table, computes a few derived columns, and aggregates to a result. You can run it locally with `xorq run` and the numbers look right. The next thing you want is for it to run every morning on a server, without you in the loop.
 
 This tutorial walks the whole path: how to lay out the repo, what to install on a vanilla Linux box, how to handle secrets and dev/prod splits, what the cron line actually looks like, where the output goes, and how you find out when it breaks. The reason it works without a managed scheduler is that a xorq build is a self-contained, content-addressed artifact — once it exists, "deploying" is mostly choosing where on disk it lives and what time of day to invoke `xorq run` on it.
 
@@ -16,12 +16,18 @@ You need:
 - Xorq installed: `pip install xorq` (or `uv add xorq`)
 - A working expression in a `.py` file with a top-level binding (this tutorial uses one as the running example, but bring your own if you have one)
 - Completed [Your first build](your_first_build.qmd) — you should be comfortable with `xorq build` and `xorq run`
+- Familiarity with UDXFs ([Your first UDXF](../ai_tutorials/your_first_udxf.qmd)) and saved connection profiles ([Profiles](../../api_reference/backends/profiles_api.qmd)), since the running example uses both
 - A Linux server you can SSH into (any small VPS works — the example uses a €4.50/month Hetzner CX22, but a `t3.small` on EC2 or a DigitalOcean droplet is the same shape)
 - Git, with the pipeline code in a repo you can clone onto the server
 
+:::{.callout-note}
+### No Postgres or Polymarket account?
+The running example reaches a real Postgres database and the Polymarket API. If you don't have either, treat the code as illustrative and substitute your own data sources — the deploy mechanics in §1–§6 don't depend on the example's specifics.
+:::
+
 ## The pipeline we're deploying
 
-For concreteness, here's the pipeline we'll work with. It:
+For concreteness, here's the pipeline we'll work with. (Polymarket is a prediction market for political and sports events; the only thing that matters here is that its API returns a current price per `market_id`.) It:
 
 1. Reads a Postgres table of open bets.
 2. Calls the Polymarket REST API (via a UDXF) to fetch current market odds for each bet.
@@ -83,26 +89,30 @@ valued = flight_udxf(
 )
 
 
-# 3. Compute P&L per bet, then aggregate to one row.
+# 3. Compute P&L per bet, then aggregate to one row, then stamp the
+#    logical run date as a column. Keeping run_date out of the aggregate
+#    itself avoids the "is this a grouping key or an output literal?"
+#    ambiguity.
 run_date = xo.param("run_date", "date")
 
 valued = valued.mutate(
     pnl=(valued.current_odds / valued.entry_odds - 1) * valued.stake,
 )
 
-expr = valued.aggregate(
-    run_date=run_date,
+aggregated = valued.aggregate(
     open_bet_count=valued.bet_id.count(),
     total_stake=valued.stake.sum(),
     pnl=valued.pnl.sum(),
 )
+
+expr = aggregated.mutate(run_date=run_date)
 ```
 
 Three things worth noticing before we go any further:
 
 - **No credentials in the source.** `xo.postgres.connect_env()` reads `POSTGRES_HOST` etc. from the environment at run time, and the API key is a plain `os.environ` lookup inside the UDXF. The build artifact is safe to commit; the secrets stay on whichever machine runs it.
 - **`run_date` is an `xo.param`, not a hardcoded `today()`.** That's what makes the pipeline replayable from cron — yesterday's output and today's output are determined by an explicit input, not by what the calendar says when the script executes.
-- **The network call is a `flight_udxf`, not an inline UDF.** UDXFs run in their own Arrow Flight server, separately from whatever backend is executing the query. That's the right shape for per-row external IO; if Polymarket calls ever need their own scaling story (rate limits, retries, a shared cache), the Flight server is where that lives, with no change to `pipeline.py`.
+- **The network call is a `flight_udxf`, not an inline UDF.** The UDXF runs against an *ephemeral* Arrow Flight server that `xorq run` spawns inside the same process, so the docker-compose service in §2 doesn't need a separate sidecar — one container is enough. The shape still pays off later: if Polymarket calls ever need their own scaling story (rate limits, retries, a shared cache), the Flight server is where that lives, with no change to `pipeline.py`.
 
 Locally, you build it once with `xorq build pipeline.py -e expr` (which produces a versioned artifact under `builds/<hash>/`) and run it with:
 
@@ -133,16 +143,16 @@ pipeline-repo/
 ├── pipeline.py
 ├── pyproject.toml
 ├── uv.lock
-├── Dockerfile
-├── compose.yaml
+├── Dockerfile          ← written in §2
+├── compose.yaml        ← written in §2
 ├── deploy/
-│   ├── run.sh
-│   └── crontab
+│   ├── run.sh          ← written in §4
+│   └── crontab         ← written in §4
 └── .xorq/catalogs/
-    └── polymarket/   ← git submodule, holds built artifacts
+    └── polymarket/     ← git submodule, holds built artifacts
 ```
 
-You set this up once, from inside the pipeline repo:
+You set this up once, from inside the pipeline repo **on your dev machine** (everything below is local; the server steps start in §2):
 
 ```bash
 # Initialize a new catalog and attach it as a submodule of the current repo.
@@ -188,12 +198,13 @@ usermod -aG sudo pipeline
 rsync --archive --chown=pipeline:pipeline ~/.ssh /home/pipeline
 ```
 
-Then as `pipeline`, install Docker and add yourself to the `docker` group:
+Then as `pipeline`, install Docker. The Ubuntu-packaged `docker.io` lags upstream; for a tutorial we recommend the official upstream install instructions at <https://docs.docker.com/engine/install/ubuntu/>, which give you `docker-ce` and the `docker compose` plugin together. After install:
 
 ```bash
-sudo apt update && sudo apt install -y docker.io docker-compose-v2 git
 sudo usermod -aG docker pipeline
-newgrp docker   # pick up the new group without logging out
+newgrp docker   # pick up the new group in the current shell;
+                # cron jobs running as `pipeline` get the group automatically.
+sudo apt install -y git
 ```
 
 Clone the pipeline repo:
@@ -202,6 +213,11 @@ Clone the pipeline repo:
 git clone --recurse-submodules git@github.com:you/pipeline-repo.git
 cd pipeline-repo
 ```
+
+:::{.callout-note}
+### Submodule auth
+`--recurse-submodules` clones the catalog at the same time. If the catalog repo is private, the server needs read access to it as well as to the parent repo — usually a separate [GitHub deploy key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys) on the catalog repo, or a GitHub App / fine-grained token that covers both. The first deploy fails at the submodule step otherwise, and the diagnostic isn't always obvious.
+:::
 
 ### The image: `Dockerfile`
 
@@ -222,6 +238,11 @@ ENTRYPOINT ["xorq"]
 ```
 
 `pyproject.toml` lists `xorq` plus anything the UDXF imports (`pandas`, in our case). `uv sync --frozen` reproduces the lockfile exactly — same image hash for the same lockfile.
+
+:::{.callout-warning}
+### Image vs. catalog version skew
+The xorq version in the *image* and the xorq version that produced the *catalog entry* must match. If you bump xorq locally and rebuild the catalog, also `docker compose build` on the server before the next cron run — otherwise the server runs an artifact whose format the installed xorq may not understand, and the failure mode is a confusing deserialization error rather than a clean version mismatch.
+:::
 
 ### The service: `compose.yaml`
 
@@ -245,7 +266,7 @@ services:
 2. The pipeline repo (including the catalog submodule) is mounted read-only into the container — `git pull` is enough to ship a new build artifact.
 3. Output parquet files land on the host at `/var/lib/pipeline/output/`, where the §5 DB-loader step picks them up.
 4. xorq's per-run log directory is mounted to a host path so the §6 monitoring check can read it without `docker exec`.
-5. Tells xorq to write run logs to the mounted path instead of the default `~/.local/share/xorq/runs/`.
+5. Tells xorq to write run logs to the mounted path instead of the default `~/.local/share/xorq/runs/`. This env var has to be set *before* the xorq Python process starts — Docker-loaded env_file vars are, but a `export XORQ_RUNS_LOGS_DIR=...` from inside a Python session after `import xorq` has no effect.
 
 Build the image and verify end-to-end:
 
@@ -270,6 +291,8 @@ The pipeline needs credentials: a Postgres URL, a Polymarket API key. The point 
 
 When you `xorq build pipeline.py`, the build process walks every backend the expression touches, extracts each one's connection profile, and writes them into the artifact as `profiles.yaml`. The profile structure travels with the build, but the secrets do not — they are stored as `${VAR}` references that get resolved at `xorq run` time.
 
+This assumes the profile is already registered locally before you build. The first time you connect via `xo.postgres.connect_env()`, save the resulting profile so the build picks it up by name rather than reconstructing it (see [Profiles](../../api_reference/backends/profiles_api.qmd)).
+
 What does a baked profile look like? Effectively this:
 
 ```yaml
@@ -279,7 +302,7 @@ postgres_<hash>:
   kwargs_dict:
     host: ${POSTGRES_HOST}
     port: ${POSTGRES_PORT}
-    database: ${POSTGRES_DB}
+    database: ${POSTGRES_DATABASE}
     user: ${POSTGRES_USER}
     password: ${POSTGRES_PASSWORD}
 ```
@@ -298,11 +321,13 @@ Just env vars. Put them in `/etc/pipeline/env.prod` (mode 600, owned by `pipelin
 ```
 POSTGRES_HOST=db.internal
 POSTGRES_PORT=5432
-POSTGRES_DB=pnl
-POSTGRES_USER=pipeline
+POSTGRES_DATABASE=pnl
+POSTGRES_USER=pnl_writer
 POSTGRES_PASSWORD=...
 POLYMARKET_API_KEY=...
 ```
+
+The Postgres user is named `pnl_writer` to keep it distinct from the OS user `pipeline` we created on the server — they're unrelated, and using the same name for both invites confusion.
 
 `compose.yaml` references this file via `env_file`, so Docker loads it into the container's environment on every `docker compose run`. The host shell never has the secrets in scope.
 
@@ -332,20 +357,35 @@ set -euo pipefail
 
 cd /home/pipeline/pipeline-repo
 
+# Source the env file so the host shell can pass secrets to the
+# downstream loader step (§5). The compose service still picks up
+# its own copy via env_file in compose.yaml.
+set -a
+source /etc/pipeline/env.prod
+set +a
+
 DATE=$(date -u +%Y-%m-%d)
 STAMP=$(date -u +%Y%m%d)
 OUT_CTR=/output/pnl-${STAMP}.parquet
 
-docker compose run --rm pipeline \
-  catalog -n polymarket -r . run prod \
-    --output-path "$OUT_CTR" \
-    --format parquet \
-    --params run_date="$DATE"
+# Alert on xorq exit code separately from the post-run meta.json check —
+# a hard crash never produces a meta.json, so the §6 jq check alone
+# would silently swallow it.
+if ! docker compose run --rm pipeline \
+        catalog -n polymarket -r . run prod \
+          --output-path "$OUT_CTR" \
+          --format parquet \
+          --params run_date="$DATE"; then
+    curl -fsS -X POST -d "Pipeline run failed (xorq exit non-zero) at $DATE" \
+        "$SLACK_WEBHOOK_URL" || true
+    exit 1
+fi
 ```
 
 The crontab (`crontab -e` as `pipeline`):
 
 ```
+CRON_TZ=UTC
 # m  h  dom mon dow  command
   0  6  *   *   *    /home/pipeline/pipeline-repo/deploy/run.sh >> /var/log/pipeline/cron.log 2>&1
 ```
@@ -354,6 +394,7 @@ Three things to notice:
 
 - **`docker compose run --rm` handles the environment.** The image has xorq and its deps; the `env_file` in `compose.yaml` injects secrets; the `:ro` mount of the repo brings the catalog submodule in. There's no virtualenv activation, no `apt install`, no "did the server have the right Python version" — the container is the answer to all of those.
 - **`--params run_date=...`** passes the logical date into the expression. Use this instead of `date('now')` inside the pipeline — it means the same command is deterministic and replayable, and a backfill is a bash loop over dates rather than a "wait, what does the pipeline assume about wall-clock time?" investigation.
+- **`CRON_TZ=UTC`** pins the schedule to UTC regardless of what timezone the server is in. Without it, `0 6 * * *` runs at the server's local time, which is rarely what you want for a pipeline whose `run_date` is a UTC calendar day.
 - **Resolving the alias.** `xorq catalog ... run prod` looks up the `prod` alias in the submodule catalog and runs the entry it points at. Bump the alias locally, push both the catalog and the parent repo, then `git pull --recurse-submodules` on the server (plain `git pull` updates the parent's submodule pointer but doesn't fetch new submodule contents), and the next cron tick picks up the new version — *no `docker compose build` needed*, because the catalog comes in via the read-only bind mount.
 
 ## 5. Output and downstream
@@ -363,32 +404,38 @@ Here's the most honest gap in today's story: **xorq has no "write to a destinati
 For our pipeline, that means:
 
 1. `xorq run` writes `/var/lib/pipeline/output/pnl-20260424.parquet` on the host (via the `:/output` mount in `compose.yaml`).
-2. A follow-up SQL statement loads it into Postgres.
+2. A follow-up step loads it into Postgres.
 
-The follow-up fits in a few lines appended to `run.sh`. Easiest is to run `psql` in the same compose project — add a one-shot service that points at the same env file:
+Vanilla Postgres does **not** read parquet directly — there is no `parquet_scan` table function in core Postgres, and the `parquet_fdw` extension exposes parquet only via pre-declared foreign tables, not an inline path. So the load step has to bridge the format gap. Three options, in increasing order of moving parts:
 
-```yaml
-# compose.yaml (additional service)
-  psql:
-    image: postgres:16-alpine
-    env_file:
-      - /etc/pipeline/env.prod
-    volumes:
-      - /var/lib/pipeline/output:/output:ro
-    entrypoint: ["psql"]
-```
+- **Postgres + a Python loader** (recommended for this tutorial). Reuse the `pipeline` image — it already has pandas and a Postgres driver — and append a load command to `run.sh`. The `set -a; source /etc/pipeline/env.prod; set +a` lines at the top of `run.sh` make `$POSTGRES_*` available; pass them through to a one-shot container:
 
-Then in `run.sh`:
+  ```bash
+  docker run --rm \
+    --entrypoint python3 \
+    -e PGHOST="$POSTGRES_HOST" -e PGPORT="$POSTGRES_PORT" \
+    -e PGDATABASE="$POSTGRES_DATABASE" -e PGUSER="$POSTGRES_USER" \
+    -e PGPASSWORD="$POSTGRES_PASSWORD" \
+    -e OUT_CTR="$OUT_CTR" \
+    -v /var/lib/pipeline/output:/output:ro \
+    pipeline:local \
+    -c "import pandas as pd, os, psycopg
+  df = pd.read_parquet(os.environ['OUT_CTR'])
+  with psycopg.connect(
+      f\"host={os.environ['PGHOST']} dbname={os.environ['PGDATABASE']} \"
+      f\"user={os.environ['PGUSER']} password={os.environ['PGPASSWORD']}\"
+  ) as c, c.cursor() as cur:
+      for row in df.itertuples(index=False):
+          cur.execute('INSERT INTO pnl_daily VALUES (%s,%s,%s,%s)', tuple(row))"
+  ```
 
-```bash
-docker compose run --rm psql \
-  "host=$POSTGRES_HOST dbname=$POSTGRES_DB user=$POSTGRES_USER" \
-  -c "INSERT INTO pnl_daily SELECT * FROM parquet_scan('$OUT_CTR')"
-```
+  `--entrypoint python3` overrides the image's `xorq` entrypoint for this one invocation. The image already has `pandas`; `psycopg` comes in transitively because `pyproject.toml` lists `xorq[postgres]` to support `xo.postgres.connect_env()` in the pipeline itself.
 
-(Postgres needs the `parquet_fdw` extension for `parquet_scan`; on a server you control, install it once. DuckDB/MotherDuck and Snowflake have native `COPY INTO` from parquet, which is even simpler.)
+- **Postgres + `parquet_fdw`.** Set up a foreign table once that points at a stable location and have the cron run write to that exact path each time, then `INSERT INTO pnl_daily SELECT * FROM <foreign_table>` via `psql`. Cleaner SQL, more setup, and you give up the dated filename audit trail.
 
-This is not elegant. It is also not much code, and you get a parquet file you can retain for audit — which turns out to be useful every single time you need to reconstruct what the dashboard said three weeks ago.
+- **DuckDB / MotherDuck / Snowflake.** Native `COPY ... FROM '/output/pnl-...parquet'`. If you have flexibility on the destination, this is the least friction by a wide margin — there's no format bridge to write.
+
+In all three cases, you keep the parquet file on disk as well — useful every single time you need to reconstruct what the dashboard said three weeks ago.
 
 ## 6. Monitoring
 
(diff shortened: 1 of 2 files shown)


paddymul (commented):
Adversarial self-review. Each comment is tagged `[#N]` so we can tick them off as we go. #3 (the local→catalog→server commit/push dance) is intentionally deferred — see the inline comment on the catalog setup block; we'll address it comprehensively in #1878 and link from here.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+    pnl=(valued.current_odds / valued.entry_odds - 1) * valued.stake,
+)
+
+expr = valued.aggregate(
+    run_date=run_date,
+    open_bet_count=valued.bet_id.count(),
+    total_stake=valued.stake.sum(),
+    pnl=valued.pnl.sum(),
+)
+```
```
paddymul, line 109: **[#1] The pipeline code has never been run end-to-end.** The source blog (xorq-website#23) explicitly listed "Build the polymarket pipeline example end-to-end" as pre-work that hadn't been done. We carried it forward without flagging that. Most likely failure point: putting `run_date=run_date` (a scalar `xo.param`) directly inside `valued.aggregate(...)` as an output column — that's an unusual aggregate shape and not obviously valid. **Action:** either run the example end-to-end and write down what actually happened, or swap to a smaller verified example.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+# m  h  dom mon dow  command
+  0  6  *   *   *    /home/pipeline/pipeline-repo/deploy/run.sh >> /var/log/pipeline/cron.log 2>&1
+```
+
+Three things to notice:
+
+- **`docker compose run --rm` handles the environment.** The image has xorq and its deps; the `env_file` in `compose.yaml` injects secrets; the `:ro` mount of the repo brings the catalog submodule in. There's no virtualenv activation, no `apt install`, no "did the server have the right Python version" — the container is the answer to all of those.
+- **`--params run_date=...`** passes the logical date into the expression. Use this instead of `date('now')` inside the pipeline — it means the same command is deterministic and replayable, and a backfill is a bash loop over dates rather than a "wait, what does the pipeline assume about wall-clock time?" investigation.
+- **Resolving the alias.** `xorq catalog ... run prod` looks up the `prod` alias in the submodule catalog and runs the entry it points at. Bump the alias locally, `git push`, `git pull` on the server, and the next cron tick picks up the new version — *no `docker compose build` needed*, because the catalog comes in via the read-only bind mount.
```
paddymul: **[#2] `git pull` on the server won't pull catalog submodule contents.** Plain `git pull` updates the parent's submodule pointer but doesn't fetch new submodule blobs. Need `git pull --recurse-submodules`, or `git -c submodule.recurse=true pull`, or set `git config submodule.recurse true` once. Single most common submodule footgun and we walk readers straight into it.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+# `-r .` is the parent repo to add the submodule to; `-n` is the catalog name.
+xorq catalog -n polymarket -r . init
+
+# Build the expression locally.
+xorq build pipeline.py -e expr
+# → builds/<hash>/
+
+# Add the build artifact to the catalog with a `prod` alias.
+xorq catalog -n polymarket -r . add builds/<hash> --alias prod
+```
```
paddymul, line 168: **[#3] The local→catalog→server commit/push dance is missing here.** After `xorq catalog ... add` you need to commit inside the submodule, push the submodule, then `git add .xorq/catalogs/<name> && git commit && git push` in the parent. None of that is shown. **Action:** comprehensively addressed in #1878 (catalog collaboration story). Here we add a one-line bridge: "After `xorq catalog ... add`, push the catalog and bump the parent submodule SHA — see [Working with the catalog](working_with_the_catalog.qmd) for the full collaboration loop." That keeps this tutorial focused on *deploying* the artifact and offloads the share-and-version mechanics to where they belong.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+      - /var/lib/pipeline/output:/output:ro
+    entrypoint: ["psql"]
+```
+
+Then in `run.sh`:
+
+```bash
+docker compose run --rm psql \
+  "host=$POSTGRES_HOST dbname=$POSTGRES_DB user=$POSTGRES_USER" \
+  -c "INSERT INTO pnl_daily SELECT * FROM parquet_scan('$OUT_CTR')"
```
paddymul: **[#4] The `psql` shell snippet is broken.** `$POSTGRES_HOST`, `$POSTGRES_DB`, `$POSTGRES_USER` are env_file-loaded into the *container*, not the host shell where `run.sh` runs — so bash interpolates them to empty strings. Also no password is passed (`PGPASSWORD` is never set). **Action:** either source the env file in `run.sh` before the `docker compose run psql`, or pass connection info via env vars only and let psql pick them up from the container's environment (rename to `PGHOST`/`PGDATABASE`/`PGUSER`/`PGPASSWORD` in the env_file or alias them in the psql service).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+```
+
+(Postgres needs the `parquet_fdw` extension for `parquet_scan`; on a server you control, install it once. DuckDB/MotherDuck and Snowflake have native `COPY INTO` from parquet, which is even simpler.)
```
paddymul: **[#5] `parquet_scan('...')` is not a Postgres function.** That's DuckDB syntax. `parquet_fdw` provides foreign tables, not a `parquet_scan` table function — the line as written runs on neither vanilla Postgres nor with parquet_fdw installed. **Action:** either show the parquet_fdw foreign-table flow, swap to a DuckDB/MotherDuck loader (`COPY ... FROM '/output/...parquet'`), or pipe through `\copy` from CSV.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+Xorq writes per-run logs whether you ask it to or not. Because `compose.yaml` sets `XORQ_RUNS_LOGS_DIR=/runs` and mounts `/var/lib/pipeline/runs` to that path, you'll find them on the host at `/var/lib/pipeline/runs/<expr_hash>/<run_id>/` — no `docker exec` required:
+
+- `run.jsonl` — an append-only event log of what happened during the run.
+- `meta.json` — a summary written at completion, with status, duration, row counts, and bytes written.
+
+That `meta.json` is the minimum-viable monitoring surface. A three-line bash check in `run.sh` (after the `docker compose run`) covers the common case:
+
+```bash
+META=$(find /var/lib/pipeline/runs -name meta.json -newer /tmp/last-run -print -quit)
```
paddymul: **[#6] `find -newer /tmp/last-run` errors on first run** — the file doesn't exist yet, so `find` prints "cannot access '/tmp/last-run'" and exits non-zero. Combined with `set -euo pipefail`, the wrapper script aborts. **Action:** seed the marker once (`[ -f /tmp/last-run ] || touch /tmp/last-run`) before the find, or move the `touch` to the start of the block.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+STATUS=$(jq -r .status "$META")
+[ "$STATUS" = "ok" ] || curl -X POST -d "Pipeline failed: $META" "$SLACK_WEBHOOK_URL"
```
paddymul: **[#7] The jq monitoring check only fires on success.** If `xorq run` itself crashes hard enough that no `meta.json` is written (network blip, OOM, broken catalog, missing env var), the find returns empty, jq errors on empty input, and `set -e` exits — the Slack webhook is never called. The most likely failure mode has zero alerting. **Action:** check the docker compose run exit code first (`if ! docker compose run ...; then alert && exit 1; fi`) and treat the meta.json check as the success-path sanity check, not the only alerting hook.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: **[#8] `find ... -print -quit` returns the *first* match, not the latest.** With multiple expression hashes accumulating under `runs/`, this is non-deterministic and likely picks the wrong run. **Action:** sort by mtime and take the newest (`find ... -printf '%T@ %p\n' | sort -n | tail -1 | cut -d' ' -f2-`), or use the run-id we just produced (capture from xorq output if available).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+    --output-path "$OUT_CTR" \
+    --format parquet \
+    --params run_date="$DATE"
+```
+
+The crontab (`crontab -e` as `pipeline`):
+
+```
+# m  h  dom mon dow  command
+  0  6  *   *   *    /home/pipeline/pipeline-repo/deploy/run.sh >> /var/log/pipeline/cron.log 2>&1
```
paddymul, line 390: **[#9] Cron timezone.** `0 6 * * *` runs at the *server's* local time, but the intro promised "06:00 UTC." **Action:** add `CRON_TZ=UTC` at the top of the crontab, or tell readers to `sudo timedatectl set-timezone UTC` once during server setup, or stop claiming UTC.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+Clone the pipeline repo:
+
+```bash
+git clone --recurse-submodules git@github.com:you/pipeline-repo.git
+cd pipeline-repo
+```
+
+### The image: `Dockerfile`
+
+The image is just xorq plus the dependencies your UDXFs need. The pipeline source and the catalog submodule are *not* baked into the image — they come in via a bind mount at run time, so updating the catalog is a `git pull`, not a rebuild. The image only changes when `pyproject.toml` changes.
```
paddymul, line 224: **[#10] The "image only changes when pyproject.toml changes" claim is too strong.** A xorq version bump in `pyproject.toml` ↔ build artifact format coupling means the server's image must track the same xorq the build was produced with. If a reader bumps xorq locally, rebuilds the catalog, and forgets to `docker compose build` on the server, they get a version-skew failure with no diagnostic. **Action:** add a one-liner: "any change to `pyproject.toml`/`uv.lock` requires `docker compose build` on the server before the next cron run."

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+### What the server needs
+
+Just env vars. Put them in `/etc/pipeline/env.prod` (mode 600, owned by `pipeline`):
+
+```
+POSTGRES_HOST=db.internal
+POSTGRES_PORT=5432
+POSTGRES_DB=pnl
+POSTGRES_USER=pipeline
```
paddymul: **[#11] Username collision: OS user `pipeline` and Postgres user `pipeline` are unrelated but a reader will conflate them.** **Action:** rename one — easiest is `POSTGRES_USER=pnl_writer` (or whatever) in the env file.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+├── pyproject.toml
+├── uv.lock
+├── Dockerfile
+├── compose.yaml
+├── deploy/
+│   ├── run.sh
+│   └── crontab
+└── .xorq/catalogs/
+    └── polymarket/   ← git submodule, holds built artifacts
+```
```
paddymul, line 153: **[#12] The repo layout diagram shows `Dockerfile` and `compose.yaml` that don't exist yet at this point in the tutorial.** They're introduced in §2. **Action:** annotate them in the diagram (`(written in §2)`) or move the diagram after §2.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+You set this up once, from inside the pipeline repo:
+
+```bash
```
paddymul, line 157: **[#13] Build/catalog setup runs locally on the dev laptop, but that context shift is never marked.** A first-time reader doesn't know whether "You set this up once" means on the dev box or on the server they'll SSH into in §2. **Action:** prefix this block with "On your dev machine:" (and similarly mark §2 as "On the server").

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+If the second command produces `/var/lib/pipeline/output/pnl-test.parquet` on the host, you're 80% of the way there. The rest is plumbing.
+
+## 3. Secrets and dev/prod
+
+The pipeline needs credentials: a Postgres URL, a Polymarket API key. The point of the next section is that *neither of them lives in your repo*, on your laptop or on the server.
+
+### Profiles are baked into the build artifact
+
+When you `xorq build pipeline.py`, the build process walks every backend the expression touches, extracts each one's connection profile, and writes them into the artifact as `profiles.yaml`. The profile structure travels with the build, but the secrets do not — they are stored as `${VAR}` references that get resolved at `xorq run` time.
```
paddymul, line 292: **[#14] "Profiles are baked into the build artifact" assumes a saved profile already exists.** `xo.postgres.connect_env()` reads env vars at *build time* if no saved profile exists; the resulting profile may capture dev creds. The tutorial doesn't cover registering a profile in `~/.config/xorq/profiles/` or via `XORQ_PROFILE_DIR`. **Action:** add a one-line callout linking to a profile setup reference, or add a sub-step "first time: save the profile with `xo.postgres.connect_env().profile.save()`."

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+This tutorial walks the whole path: how to lay out the repo, what to install on a vanilla Linux box, how to handle secrets and dev/prod splits, what the cron line actually looks like, where the output goes, and how you find out when it breaks. The reason it works without a managed scheduler is that a xorq build is a self-contained, content-addressed artifact — once it exists, "deploying" is mostly choosing where on disk it lives and what time of day to invoke `xorq run` on it.
+
+## Prerequisites
+
+You need:
+
+- Xorq installed: `pip install xorq` (or `uv add xorq`)
+- A working expression in a `.py` file with a top-level binding (this tutorial uses one as the running example, but bring your own if you have one)
+- Completed [Your first build](your_first_build.qmd) — you should be comfortable with `xorq build` and `xorq run`
```
paddymul, line 18: **[#15] "Completed *Your first build*" is insufficient as the only prereq.** The example uses `flight_udxf`, `xo.param`, `connect_env`, and a UDXF that hits an external API — none of which are in *Your first build*. **Action:** add prereqs for the UDXF tutorial and a Postgres profile setup reference.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+```bash
+sudo apt update && sudo apt install -y docker.io docker-compose-v2 git
+sudo usermod -aG docker pipeline
+newgrp docker   # pick up the new group without logging out
+```
+
+Clone the pipeline repo:
+
+```bash
+git clone --recurse-submodules git@github.com:you/pipeline-repo.git
```
paddymul, line 213: **[#16] Submodule auth on the server is not addressed.** `git clone --recurse-submodules git@github.com:...` requires SSH keys for *both* the parent repo and the catalog submodule. If the catalog is private (likely), you need a deploy key per repo, or a GitHub App, or HTTPS+token. First-deploy failure mode for anyone with a private catalog. **Action:** add a callout under "Clone the pipeline repo" pointing at GitHub deploy keys.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+    total_stake=valued.stake.sum(),
+    pnl=valued.pnl.sum(),
+)
+```
+
+Three things worth noticing before we go any further:
+
+- **No credentials in the source.** `xo.postgres.connect_env()` reads `POSTGRES_HOST` etc. from the environment at run time, and the API key is a plain `os.environ` lookup inside the UDXF. The build artifact is safe to commit; the secrets stay on whichever machine runs it.
+- **`run_date` is an `xo.param`, not a hardcoded `today()`.** That's what makes the pipeline replayable from cron — yesterday's output and today's output are determined by an explicit input, not by what the calendar says when the script executes.
+- **The network call is a `flight_udxf`, not an inline UDF.** UDXFs run in their own Arrow Flight server, separately from whatever backend is executing the query. That's the right shape for per-row external IO; if Polymarket calls ever need their own scaling story (rate limits, retries, a shared cache), the Flight server is where that lives, with no change to `pipeline.py`.
```
paddymul: **[#17] UDXF Flight server inside Docker is hand-waved.** The text says UDXFs run "in their own Arrow Flight server, separately from the query engine" — but inside one `docker compose run`, where? Same process, subprocess, sidecar? If it just works in-process for `xorq run`, state that. If it needs a port exposed or a sidecar service in compose, the compose file is missing it. **Action:** verify with a real run, then state the answer plainly.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+```bash
+adduser --gecos "" pipeline
+usermod -aG sudo pipeline
+rsync --archive --chown=pipeline:pipeline ~/.ssh /home/pipeline
+```
+
+Then as `pipeline`, install Docker and add yourself to the `docker` group:
+
+```bash
+sudo apt update && sudo apt install -y docker.io docker-compose-v2 git
```
paddymul: **[#18] `docker.io` is the Ubuntu-packaged Docker, which lags upstream; `docker-compose-v2` package availability is also Ubuntu-version-dependent.** The recommended path is the upstream `docker-ce` repo. **Action:** either link to https://docs.docker.com/engine/install/ubuntu/ for the official install (probably best for a tutorial), or test the apt one-liner on a fresh 24.04 box and document any caveats.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+    build: .
+    image: pipeline:local
+    env_file:
+      - /etc/pipeline/env.prod   # <1>
+    volumes:
+      - .:/work:ro                              # <2>
+      - /var/lib/pipeline/output:/output        # <3>
+      - /var/lib/pipeline/runs:/runs            # <4>
+    environment:
+      XORQ_RUNS_LOGS_DIR: /runs                 # <5>
```
paddymul, line 262: **[#19] `XORQ_RUNS_LOGS_DIR` has a sharp edge.** From `python/xorq/common/utils/logging_utils.py:164`: "modifying env var XORQ_RUNS_LOGS_DIR won't have any impact after first import." Our docker-compose flow is fine because env is set before xorq imports, but a reader who tries `export XORQ_RUNS_LOGS_DIR=...` from inside a Python session or shell after xorq import gets the default silently. **Action:** small callout noting the var must be set before the xorq process starts.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+STATUS=$(jq -r .status "$META")
+[ "$STATUS" = "ok" ] || curl -X POST -d "Pipeline failed: $META" "$SLACK_WEBHOOK_URL"
+touch /tmp/last-run
+```
+
+For anything fancier — per-stage timing, distributed traces, Prometheus metrics — xorq has OpenTelemetry instrumentation wired into the CLI. It's there if you want it; you don't need it on day one.
+
+:::{.callout-note}
+### Log hygiene
+The per-run directory tree grows forever by default. Add a companion cron to clean it up:
```
paddymul: **[#20] No cleanup for `/var/lib/pipeline/output/` itself.** Daily parquet files accumulate forever. The hygiene callout cleans up `runs/`, not `output/`. **Action:** add a parallel `find /var/lib/pipeline/output -mtime +90 -delete` or note that retention is intentional for audit (and pick a number).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
+---
+title: "Deploy a pipeline on a schedule"
+description: "Take a working xorq expression from your laptop to a server that runs it every morning under cron"
+icon: "rocket-takeoff"
+headline: "From `xorq run` to a cron line"
+---
+
+You have a xorq expression. It pulls data from somewhere, joins it against a table, computes a few derived columns, and aggregates to a result. You can run it locally with `xorq run` and the numbers look right. The next sentence out of your mouth is: *"I need this to run every morning on a server."*
```
paddymul: **[#21] Tutorial vs. blog voice.** Phrases like "the next sentence out of your mouth" and "an afternoon of yak-shaving" are blog voice. The example dumps 70 lines (UDXFs, env-resolved profiles, `xo.param`, schema construction, aggregates) up front rather than building incrementally. **Action:** decide explicitly whether this is a tutorial (build up the example) or a blog-style guide (current shape) and lean into one. Probably keep it as-is given Paddy's narrative-over-API preference, but acknowledge the choice.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+You need:
+
+- Xorq installed: `pip install xorq` (or `uv add xorq`)
+- A working expression in a `.py` file with a top-level binding (this tutorial uses one as the running example, but bring your own if you have one)
+- Completed [Your first build](your_first_build.qmd) — you should be comfortable with `xorq build` and `xorq run`
+- A Linux server you can SSH into (any small VPS works — the example uses a €4.50/month Hetzner CX22, but a `t3.small` on EC2 or a DigitalOcean droplet is the same shape)
+- Git, with the pipeline code in a repo you can clone onto the server
+
+## The pipeline we're deploying
```
paddymul, line 28: **[#22] Polymarket as the example is niche.** A reader who doesn't know it now learns the domain and the deploy story simultaneously. **Action:** consider a more universal example (weather API, GitHub stars, a public Postgres dataset) — or keep Polymarket and add a one-line "if Polymarket isn't familiar: it's a prediction market; the only thing you need to know here is that the API returns a JSON price per market_id."

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+headline: "From `xorq run` to a cron line"
+---
+
+You have a xorq expression. It pulls data from somewhere, joins it against a table, computes a few derived columns, and aggregates to a result. You can run it locally with `xorq run` and the numbers look right. The next sentence out of your mouth is: *"I need this to run every morning on a server."*
+
+This tutorial walks the whole path: how to lay out the repo, what to install on a vanilla Linux box, how to handle secrets and dev/prod splits, what the cron line actually looks like, where the output goes, and how you find out when it breaks. The reason it works without a managed scheduler is that a xorq build is a self-contained, content-addressed artifact — once it exists, "deploying" is mostly choosing where on disk it lives and what time of day to invoke `xorq run` on it.
+
+## Prerequisites
+
+You need:
```
paddymul, line 14: **[#23] No path for following along without infrastructure.** Reader needs Postgres with an `open_bets` table, a Polymarket API key, a Linux server, two git repos with SSH access, plus understanding of UDXFs. **Action:** offer a minimal-infra fallback: a SQLite or DuckDB-based variant the reader can run on their laptop end-to-end before they buy a server.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+- **`docker compose run` plus cron** replaces a managed scheduler for a single pipeline running on a server. The wrapper script invokes the compose service, which writes parquet to a host-mounted directory.
+- **Run logs land on a host-mounted volume per run.** A three-line jq check is enough to alert on failure. OTel is there when you outgrow that.
+
+Total cost of everything in this tutorial: €4.50/month for the server, $0 for the software, an afternoon of yak-shaving the first time. If you've been putting off moving your pipeline off your laptop because "I don't have Airflow set up" — you don't need it.
+
+## Next steps
+
+Now that the pipeline runs on a schedule:
+
+- [Working with the catalog](working_with_the_catalog.qmd) shows you how to share the catalog with collaborators via GitHub, including swapping profiles at recovery time.
```
paddymul, line 492: **[#24] Next-steps links target tutorials that aren't in `main`'s sidebar.** `working_with_the_catalog.qmd`, `explore_caching.qmd`, `switch_backends.qmd` exist as files but aren't in `docs/_quarto.yml` on main — pages will render but appear orphaned. **Action:** either expose them in the sidebar (as part of the new "Core tutorials" section we added), or scope next-steps to currently-exposed pages.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
```diff
@@ -0,0 +1,437 @@
...
+```
+0 3 * * 0  find /var/lib/pipeline/runs -mtime +30 -delete
+```
+:::
+
+## What you learned
+
+You took a working xorq expression and put it on a server, on a schedule. Specifically:
+
+- A **build artifact**, not a `.py` file, is the deploy unit. It's content-addressed, self-contained, and travels with the connection profiles it needs.
```
paddymul: **[#25] "What you learned" bullets list *steps*, not *abstractions*.** A good summary names concepts the reader carries forward ("deferred profile resolution," "input-addressed deploy unit," "bind mount as deploy boundary"). The current bullets just recap the procedure. **Action:** rewrite as concept-named bullets with one-sentence why-it-matters each.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in 13f19903 — server pull is now `git pull --recurse-submodules` with an explanatory parenthetical.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Addressed expediently in 13f19903 — added a paragraph after the local catalog setup that names the missing publish loop and links to `working_with_the_catalog.qmd` for the full mechanics. Comprehensive submodule story to land in #1878 (note posted there: https://github.com/xorq-labs/xorq/pull/1878#issuecomment-4338699498).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Addressed in f1b030a6: split run_date out of the aggregate into a follow-up mutate, and added a no-infra fallback callout. The full Polymarket+Postgres path still has not been exercised end-to-end against real services — flagged as illustrative.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: run.sh now sources the env file at the top, the loader is a one-shot docker run with explicit -e PG* env vars, and PGPASSWORD is set.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: replaced the broken parquet_scan with three honest options — Python loader (recommended), parquet_fdw foreign table, or a DuckDB/MotherDuck/Snowflake destination.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: meta.json check now uses find -printf with mtime sort + tail -1 (skips the -newer marker file dance entirely).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: wrapper script now alerts on a non-zero xorq exit before the meta.json check ever runs. The post-run check is positioned as the success-path sanity check.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: replaced -print -quit with find -printf '%T@ %p' | sort -n | tail -1, so we get the most recent meta.json deterministically.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: added CRON_TZ=UTC to the crontab snippet.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: callout-warning under the Dockerfile section explicitly names the image/catalog version-skew failure mode and prescribes 'docker compose build on the server'.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: env file now uses POSTGRES_USER=pnl_writer with a one-line note explaining the rename. Also caught a separate bug — POSTGRES_DB → POSTGRES_DATABASE, which is what xorq's connect_env actually reads.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: each file in the repo layout diagram is annotated with the section it's introduced in (← written in §2, ← written in §4).

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: prefixed the local catalog setup block with 'on your dev machine (everything below is local; the server steps start in §2)'.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: added a sentence after the profile YAML pointing at the profiles api reference for the first-time profile registration step.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: prereqs now also name the UDXF tutorial and the Profiles api reference.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: callout-note under the clone command linking to GitHub deploy keys, with a one-line description of the failure mode.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: verified flight_udxf uses an *ephemeral* in-process Flight server (relations.py:405 — `with flight_udxf.make_server() as server`), so no compose sidecar needed. State that explicitly in the bullet now.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: the apt one-liner is replaced with a link to https://docs.docker.com/engine/install/ubuntu/ — the upstream docker-ce path. Only the git install stays in apt.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: callout in the compose annotation noting XORQ_RUNS_LOGS_DIR must be set before the xorq Python process starts.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: log hygiene callout now has parallel finds for runs/ (30 days) and output/ (90 days), with explicit 'pick retention windows that match your audit needs.'

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Partially addressed in f1b030a6: dropped 'next sentence out of your mouth' and 'yak-shaving'. Kept the running narrative voice intentionally — Paddy's docs preference is narrative-heavy.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: one-line gloss in the section intro — '(Polymarket is a prediction market for political and sports events; the only thing that matters here is that its API returns a current price per market_id.)'.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: prereqs section gets a callout-note 'No Postgres or Polymarket account?' explicitly framing the example code as illustrative if the reader doesn't have either.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Not fixed in this commit. The next-steps targets (working_with_the_catalog.qmd, explore_caching.qmd, switch_backends.qmd) become sidebar-exposed once #1878 lands. Leaving as-is so the link target is correct when that PR merges.

docs/tutorials/core_tutorials/deploy_a_pipeline.qmd
paddymul: Fixed in f1b030a6: 'What you learned' rewritten around concept names (input-addressed deploy unit, deferred profile resolution, catalog-as-submodule deploy boundary, image-as-fixed-point/repo-as-moving-point, cron as scheduler) with one-sentence why-it-matters each.

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: High
**Location**: `deploy_a_pipeline.qmd` §4 `run.sh` — both `curl` invocations
**Problem**: `$SLACK_WEBHOOK_URL` is used inside `run.sh` which opens with `set -euo pipefail`. The variable is never defined in the env file example, never mentioned in the setup steps, and never guarded. With `-u`, expanding an unset variable is a fatal error — so the error-handling path itself crashes, silently swallowing the original failure.
**Fix**: Either add `SLACK_WEBHOOK_URL=...` to the `/etc/pipeline/env.prod` example in §3, or guard the expansions: `"${SLACK_WEBHOOK_URL:-}"` (and skip the `curl` when empty).

---

**Severity**: Medium
**Location**: `deploy_a_pipeline.qmd` §2 — "Next steps" block at end of tutorial and §6 body
**Problem**: `working_with_the_catalog.qmd` is referenced three times (§1 callout, §4 bullet, "Next steps") but the file does not exist in `docs/tutorials/core_tutorials/`. The other cross-refs (`your_first_build.qmd`, `explore_caching.qmd`, `switch_backends.qmd`) all exist; this one alone will produce a Quarto broken-link error at build time.
**Fix**: Create the file before merging, or replace the links with a `:::{.callout-note}` stub noting the page is forthcoming.

---

**Severity**: Medium
**Location**: `deploy_a_pipeline.qmd` §4 — "Three things to notice:"
**Problem**: The prose says "Three things to notice" but the list immediately below has four bullet points (`docker compose run --rm`, `--params`, `CRON_TZ=UTC`, "Resolving the alias").
**Fix**: Change to "Four things to notice:" or drop the count entirely.

---

**Severity**: Low
**Location**: `deploy_a_pipeline.qmd` §5 — Python loader `psycopg.connect(...)` f-string
**Problem**: Connection string is built via f-string: `f"host={os.environ['PGHOST']} ...password={os.environ['PGPASSWORD']}"`. A password containing a space or `=` will silently corrupt the connection string (libpq key=value format). `psycopg` keyword-argument form avoids this.
**Fix**: Use `psycopg.connect(host=..., dbname=..., user=..., password=...)` keyword form instead.

---

**Severity**: Low
**Location**: `deploy_a_pipeline.qmd` — `pipeline.py` `fetch_odds`, line `body["outcomes"][0]["price"]`
**Problem**: No bounds check on `body["outcomes"]`. An empty list or a missing key produces an `IndexError`/`KeyError` that surfaces as a generic UDXF failure rather than a diagnostic message.
**Fix**: Add a guard: `if not body.get("outcomes"): raise ValueError(f"no outcomes for market {mid}")`.

---

## Summary

Adds a new "Deploy a pipeline on a schedule" core tutorial covering repo layout, server setup via Docker Compose, secrets management, cron scheduling, output loading, and monitoring; the main actionable bugs are an unguarded `$SLACK_WEBHOOK_URL` that breaks the error-handling path and a broken cross-reference to a not-yet-created `working_with_the_catalog.qmd`.

</details>

---
*Review type: default | Agent: claude-code | Job: 8*

PR #1896, opened 2026-04-29
feat(tui): catalog visual polish + configurable panels

hussainsultan:
## Summary

- Catalog tree polish: kind icons (⊞ ⊕ ⊘ ⊛), refresh flash, two-tone leaves with column-count badge, visible `├─ ╰─` connectors, double-border focus ring, status-bar kind breakdown
- New `xorq.options.tui` section: `left_ratio`, `right_ratio`, `revisions_open`, `git_log_open` (env overrides: `XORQ_TUI_*`); replaces hardcoded panel sizes

## Test plan

- [x] `python -m pytest python/xorq/catalog/tests/test_tui.py -q` — 125 passed (4 pre-existing failures on `CatalogRowData.cached` are unrelated)
- [x] `XORQ_TUI_LEFT_RATIO=5 xorq catalog --name <repo> tui` — left column visibly wider
- [x] New entry highlights pink one cycle, settles to green/yellow


python/xorq/config.py
```diff
@@ -160,6 +160,31 @@ def get_path(self, name, board=None, **kwargs):
...
+
+    left_ratio: int = int(env_config.XORQ_TUI_LEFT_RATIO or 2)
+    right_ratio: int = int(env_config.XORQ_TUI_RIGHT_RATIO or 3)
+    revisions_open: bool = bool(
+        ast.literal_eval(env_config.XORQ_TUI_REVISIONS_OPEN or "False")
+    )
+    git_log_open: bool = bool(
+        ast.literal_eval(env_config.XORQ_TUI_GIT_LOG_OPEN or "False")
+    )
+
```
ghoersti, line 186: was thinking maybe something like  below  and avoid `ast.literal_eval`
```python
  _TRUTHY = {"1", "true", "yes", "on", "y", "t"}
  _FALSY  = {"0", "false", "no", "off", "n", "f", ""}

  def _env_int(name: str, default: int) -> int:
      raw = getattr(env_config, name, None)
      return default if raw is None or raw == "" else int(raw)

  def _env_bool(name: str, default: bool) -> bool:
      raw = getattr(env_config, name, None)
      if raw is None:
          return default
      s = raw.strip().lower()
      if s in _TRUTHY: return True
      if s in _FALSY:  return default if s == "" else False
      raise ValueError(f"{name}={raw!r} is not a boolean")

  Then the fields read cleanly:

  left_ratio: int       = _env_int("XORQ_TUI_LEFT_RATIO", 2)
  right_ratio: int      = _env_int("XORQ_TUI_RIGHT_RATIO", 3)
  revisions_open: bool  = _env_bool("XORQ_TUI_REVISIONS_OPEN", False)
  git_log_open: bool    = _env_bool("XORQ_TUI_GIT_LOG_OPEN", False)
```

 ▎ Aside on env parsing. ast.literal_eval(... or "False") works but is a sharp tool for a blunt job. It rejects shell-typical inputs like true/1/yes (only True/False parse), accepts anything else    
  that's a valid Python literal ([1,2,3], {'a':1}, None, …), and yields ValueError: malformed node or string traces that don't mention the env var. A small _env_bool(name, default) helper that        
  lower-cases and matches against {"1","true","yes","on"} / {"0","false","no","off"} would give friendlier UX, raise actionable errors on typos, and let the field defaults stay properly typed (False  
  instead of "False"). Same shape would clean up the int(... or 2) fields too.

dlovell (approved):

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Medium
**Location**: `python/xorq/config.py:173-179` (TUI class, bool fields)
**Problem**: `ast.literal_eval("true")` and `ast.literal_eval("yes")` raise `ValueError` — Python literals are case-sensitive. A user setting `XORQ_TUI_REVISIONS_OPEN=true` or `=1` crashes at module import with no helpful message. PR reviewer already flagged this.
**Fix**: Use a simple set-based check (as the PR discussion suggests) or at minimum `str.lower()` before eval: `ast.literal_eval((env_config.XORQ_TUI_REVISIONS_OPEN or "False").capitalize())`.

---

**Severity**: Medium
**Location**: `python/xorq/config.py:171-172` (TUI class, int fields)
**Problem**: `int(env_config.XORQ_TUI_LEFT_RATIO or 2)` raises unhandled `ValueError` at import time if the env var is set to a non-integer string (e.g. `"2.5"`, `"wide"`).
**Fix**: Wrap in try/except with a descriptive error, or validate the string is digits before conversion.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tests/test_tui.py:1185-1196` (`test_tui_env_var_override`)
**Problem**: Test only validates that the raw `env_config` object reads the monkeypatched env var — it doesn't assert that `TUI()` produces an instance with the overridden values. This is impossible with the current class-level evaluation pattern (defaults are baked in at import time), so the end-to-end path (env var → `TUI` instance) is untested and the test creates a false sense of coverage.
**Fix**: Either document the import-time-only behavior explicitly, or restructure `TUI` to re-read env vars at instantiation and add an assertion like `assert TUI().left_ratio == 7`.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tui.py:831-838` (`_styled_leaf_label`, else branch)
**Problem**: When no alias and `is_new=False`, hash is styled `XORQ_DARK.primary` (no bold), but when alias is present and `is_new=False`, the alias uses `name_style = f"bold {XORQ_DARK.primary}"`. Hash-only entries are visually inconsistent with aliased entries.
**Fix**: Change `style=name_style if is_new else XORQ_DARK.primary` → `style=name_style if is_new else f"bold {XORQ_DARK.primary}"`.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tui.py` (`_render_status`)
**Problem**: `rows = self._row_cache.values()` is a live dict view iterated twice (`kind_counts` then `cached_count`) on the main thread while the worker thread can modify `_row_cache`. Could produce mismatched counts.
**Fix**: Snapshot once: `rows = list(self._row_cache.values())`.

---

## Summary

Adds visual theming (styled tree nodes, flash-new highlighting) and `xorq.options.tui` layout config (column ratios, panel open state) backed by env vars, with the main risk being import-time crashes for common boolean env var conventions (`true`, `1`, `yes`).

</details>

---
*Review type: default | Agent: claude-code | Job: 7*

PR #1899, opened 2026-04-29
fix(catalog): single-remote constraint + push rejection (#1898, ADR-0011)

### LINKED ISSUE #1898: Catalog.push() returns success when the remote rejects the push
## Summary

`Catalog.push()` returns successfully when the remote rejects the push. Callers cannot tell whether their entries were actually published — the API reports success while the remote ref has not moved.

This was surfaced while writing a reproduction for #1886. `xorq#1886` is the broader docs/recovery issue; this one is the narrower API bug.

## Reproduction

Two catalogs cloned from the same bare remote, both add a different entry, User A pushes first (clean fast-forward), User B pushes second (should be rejected — `fetch first`):

```
BUG A: User B catalog.push() doesn't surface the remote rejection
========================================================================
  main  flags=REJECTED|ERROR  summary='[rejected] (fetch first)'
  -> BUG: catalog.push() returned without raising, but the remote REJECTED the push.
  after cat_b.push(): local HEAD          = d5b8a28bc73966cdb8c1d5ff189f375772a8536e
  after cat_b.push(): origin/main on wire = e8eb5a978a277668f57b270bcf798cd0efd7460b
  after cat_b.push(): -> remote ref did NOT move to local HEAD; the push the API said succeeded was actually rejected.
```

Two independent confirmations of the contradiction:

1. **GitPython's own report.** `Remote.push()` returns a `PushInfoList` with `REJECTED|ERROR` flags set and summary `[rejected] (fetch first)`.
2. **The remote itself.** `git ls-remote origin refs/heads/main` against the bare origin shows the `main` ref is still at User A's commit, not User B's local HEAD.

`catalog.push()` saw both of these and returned normally anyway.

Full reproduction script: `repro_1886_catalog_conflict.py` on `docs/bsl-semantic-catalog-tutorial`.

## Root cause

`python/xorq/catalog/catalog.py:306-312`:

```python
def push(self):
    """Push to all git remotes after verifying consistency."""
    self.assert_consistency()
    results = tuple(map(Remote.push, self._git_remotes))
    for remote in self._git_remotes:
        remote.push(ANNEX_BRANCH)
    return results
```

Two `Remote.push` calls, neither result is inspected. GitPython's `Remote.push()` does not raise on remote rejection — it returns a `PushInfoList` and the caller is expected to check `info.flags`. We don't.

## Why this matters in practice

- Every `catalog.add(sync=True)` (the default) calls `push()` internally. A diverged remote means `add(sync=True)` claims to publish and silently doesn't.
- The reasonable next thing for a user who got "success" from `push()` is to tell their teammate "it's published" — and now the catalog's source-of-truth claim is broken.
- This is the precondition for the conflict-resolution recipes in #1886: today users don't even reach the conflict, they get a false success and walk away.

## Proposed fix

After each `Remote.push()`, walk the returned `PushInfoList` and raise on any of `REJECTED | REMOTE_REJECTED | ERROR | REMOTE_FAILURE`. `UP_TO_DATE`, `FAST_FORWARD`, `NEW_HEAD` are clean. Both ref pushes (main and `git-annex`) should be checked — partial success leaves the remote in a half-state and is also a failure.

A small `CatalogPushRejected(refs=..., summaries=...)` exception (or similar) lets the CLI print a useful message and exit non-zero, and lets callers of `add(sync=True)` distinguish push failures from other errors.

## Scope

This issue is *only* the silent-success bug. The companion bugs (`catalog.pull()` failing on divergent branches before merging, the broader docs/conflict-resolution work) stay in #1886.

paddymul:
## Summary

Fixes #1898 and lands [ADR-0011](docs/adr/0011-catalog-single-git-remote.md).

The original bug: `catalog.push()` returned success when the remote rejected the push (e.g. divergent branch). Today the API silently swallows GitPython's `REJECTED|ERROR` flags. This PR walks the `PushInfoList` and raises `CatalogPushError` on rejection.

While writing the multi-remote aggregation logic for the rejection fix, it became clear we were patching one of an open-ended set of multi-remote consistency problems we had not designed for. ADR-0011 narrows the contract instead: **the catalog supports exactly one git remote**. Two or more raises `CatalogConfigurationError` at the next sync-side operation. A new `Catalog.set_remote(name, url)` makes the supported configuration path discoverable. The aggregation loop is reverted; its test is marked `pytest.skip` (not deleted) so the multi-remote setup is preserved for any future re-introduction.

## Walkthrough

The branch is structured TDD-style; see the commit list for the exact order. In summary:

1. **Push-rejection fix.** Failing test, then the fix that raises `CatalogPushError` when GitPython reports a rejection.
2. **Multi-remote aggregation.** Failing test for two-remote partial failure, then the aggregation implementation. Both are reverted/skipped in step 4 — preserved as the audit trail for what we built and chose not to keep.
3. **ADR-0011.** Proposes the single-remote constraint and explains why aggregation is the wrong long-term direction.
4. **Single-remote enforcement.** Failing tests (multi-remote refusal across `push`/`pull`/`fetch`/`sync`, plus `set_remote`), then the implementation that makes them pass. Reverts the aggregation loop and `pytest.skip`s its test.

## Test plan

- [ ] `test_push_surfaces_remote_rejection` passes — push rejection raises `CatalogPushError`
- [ ] `test_multi_remote_raises_configuration_error` passes for `push`/`pull`/`fetch`/`sync`
- [ ] `Catalog.set_remote` replaces (rather than appends) the catalog's single git remote
- [ ] No regression in `test_catalog_clone_from_push`, `test_push_skips_missing_annex_branch`
- [ ] `test_push_aggregates_failures_across_remotes` reports as skipped (not failed)

### DIFF
--- a/docs/adr/0009-catalog-single-git-remote.md
+++ b/docs/adr/0009-catalog-single-git-remote.md
@@ -49,7 +49,7 @@ catalog repo  ──push/pull──>  one git remote (e.g. github.com/org/catalo
 
 ### Configuring the remote
 
-`Catalog.set_remote(name, url, force=False)` is the supported way to wire up the git remote. It refuses by default when a remote is already configured — silent replacement turns a typo (`"origion"` for `"origin"`) into the deletion of the configured remote with no signal. `force=True` is the explicit opt-in for the deliberate replacement case.
+`Catalog.set_remote(name, url, force=False)` is the supported way to wire up the git remote. It refuses by default when a remote is already configured — silent replacement turns a typo in the remote name into the deletion of the existing remote with no signal. `force=True` is the explicit opt-in for the deliberate replacement case.
 
 The CLI mirrors the API: `xorq catalog set-remote URL [--name NAME] [--force]`.
 
@@ -102,7 +102,7 @@ Rejected on the same grounds as the flag option: a warning that is easy to ignor
 
 The first cut of `set_remote` deleted any existing git remote and created the new one without complaint. The thinking was that the caller obviously means to overwrite — they just called `set_remote`.
 
-Rejected because a typo (`"origion"` for `"origin"`) deletes the configured remote with no signal. The cost of asking the user to pass `force=True` for the rare deliberate-replacement case is small relative to the cost of silent destruction.
+Rejected because a typo in the remote name deletes the configured remote with no signal. The cost of asking the user to pass `force=True` for the rare deliberate-replacement case is small relative to the cost of silent destruction.
 
 ## Consequences
 
--- a/docs/adr/0009-catalog-single-git-remote.md
+++ b/docs/adr/0009-catalog-single-git-remote.md
@@ -1,0 +1,131 @@
+# ADR-0009: Catalog supports a single git remote
+
+- **Status:** Proposed
+- **Date:** 2026-04-30
+- **Context area:** `python/xorq/catalog/catalog.py`
+
+## Context
+
+A `Catalog` is a git repository — usually with a git-annex sidecar branch — that holds versioned build artifacts. `Catalog.push()`, `Catalog.pull()`, and `Catalog.fetch()` operate over `self._git_remotes`, which is *every* git remote on the underlying repo that has a fetch refspec (`catalog.py:304-315`). Nothing in the catalog API restricts a user to one remote; if `git remote add r2 …` is run on the working tree, the next `catalog.push()` will dutifully push to both `origin` and `r2`.
+
+Issue [#1898](https://github.com/letsql/xorq/issues/1898) surfaced a silent failure mode in this design: when a push to one remote was rejected (non-fast-forward), `catalog.push()` returned without raising. The fix landed in three commits on this branch:
+
+| Commit | Change |
+|--------|--------|
+| `fe3f608b` | Raise `CatalogPushError` when `PushInfo` flags include `REJECTED`/`REMOTE_REJECTED`/`REMOTE_FAILURE`/`ERROR` |
+| `6b2ab300` | Push every ref to every remote first, then aggregate all rejections into one error message — instead of bailing on the first failure |
+| `0b40bb35` | Test covering two-remote partial-failure aggregation |
+
+The aggregation fix exists because the obvious "fail fast on the first rejected push" implementation has a worse failure mode than the bug it replaces: with two git remotes, rejecting on `origin` skips the push of `origin`'s annex branch *and* the entire push to `r2`, so a retry sees inconsistent state across the two remotes. Aggregation patches that — but it patches *one* of an open-ended set of multi-remote consistency problems we have not enumerated:
+
+- **Atomicity (between remotes).** `push()` to N remotes is N independent operations. Any non-empty subset can succeed while the rest fail, leaving remotes out of sync. Aggregating the error tells the user *that* this happened, not how to recover. Note: single-remote does not give us atomicity either — `main` and `git-annex` are pushed in two separate operations (`catalog.py:346, 349`), so even one remote can be left half-pushed. Single-remote shrinks the inconsistency surface from `2N` operations to `2`; it does not eliminate it.
+- **Annex content placement.** `git-annex` tracks per-key location across remotes. With multiple git remotes plus a special remote, "where does the content live" stops being a yes/no question. `Catalog.fetch_entries` (`catalog.py:321-334`) just calls `backend.fetch_content`; it does not reason about which remote satisfies which key.
+- **Pull semantics.** `Catalog.pull()` calls `Remote.pull` on each remote in turn (`catalog.py:367-369`). With divergent histories on two remotes, the first pull succeeds and the second fails (or merges) silently with respect to the catalog's consistency invariants.
+- **`assert_consistency` against which remote?** Mutating operations (`add`, `remove`, `add_alias`) wrap their work in `synchronizing` (`catalog.py:371-375`), which pulls from every remote then pushes to every remote; `assert_consistency` runs inside the wrapped block (e.g. `_add_zip` at `catalog.py:233-239`). With multiple remotes, "the catalog is consistent with the remote" is undefined — we may have pulled from `origin` and pushed to `r2`, with `r1` untouched.
+- **`remote_config` is singular.** The git-annex `RemoteConfig` resolution (`catalog.py:200-211`) reads one config from `remote.log`. The data model already assumes a single source of truth for credentials and bucket location.
+
+We have no users today asking for multi-remote, no design for what the operations above *should* do, and no test coverage beyond the push-failure aggregation case. The aggregation fix is correct given the current API, but committing to it as a supported surface area locks in design debt we have not paid down.
+
+The aggregation fix landed before this ADR rather than alongside it because it is a strict improvement over silent rejection regardless of what we decide about multi-remote: the single-remote case is also better-served by raising than returning successfully on a rejected push. Writing the aggregation logic is what surfaced the larger surface-area question this ADR resolves.
+
+## Decision drivers
+
+- The catalog API is small and recent; we have room to constrain it before users build workflows that depend on multi-remote behavior.
+- Single-remote is the only configuration we test, document, or run in production.
+- Failure modes that "look like they work" are worse than failure modes that raise loudly.
+- We want to hear from users with this need rather than guess at requirements.
+
+## Decision
+
+The catalog officially supports exactly one git remote. Configurations with two or more git remotes are unsupported, and the catalog will refuse to operate on them rather than attempting best-effort multi-remote semantics.
+
+### What "supported" means
+
+- `Catalog.push()`, `Catalog.pull()`, `Catalog.fetch()`, and `Catalog.sync()` are defined for repos with zero or one git remote.
+  - **Zero remotes** is the local-only case (used in tests and by `init=True` before a remote is wired up). These operations are no-ops in that mode and do not raise. We accept the silent no-op here — unlike the rejected-push case, the user has not configured a remote, so there is no "I asked for X and got nothing" surprise. Local-only workflows (development, tests, fixture setup) want this.
+  - **One remote** is the supported configuration; behavior is unchanged from today minus the multi-remote aggregation logic.
+- A repo with two or more git remotes raises `CatalogConfigurationError` the first time one of those operations is invoked, with a message that names the remotes found and points the user at the single-remote constraint. `CatalogConfigurationError` is a new `RuntimeError` subclass defined in `python/xorq/catalog/catalog.py` alongside `CatalogPushError`, and is exported from `xorq.catalog`.
+- `Catalog.clone_from`, `Catalog.add`, `Catalog.remove`, and read-only operations are unaffected — they do not depend on a remote.
+
+### What "git remote" means here
+
+This ADR constrains *git* remotes — entries in `repo.remotes` with a fetch refspec, as enumerated by `Catalog._git_remotes` (`catalog.py:304-315`). The git-annex *special* remote (the S3 / directory / etc. configured via `RemoteConfig` and stored in `remote.log`) is unrelated and remains singular as it always has been. A catalog continues to look like:
+
+```
+catalog repo  ──push/pull──>  one git remote (e.g. github.com/org/catalog.git)
+     │
+     └── annex content  ──get/copy──>  one annex special remote (e.g. s3://bucket)
+```
+
+### Implementation sketch
+
+Add a `_assert_single_git_remote` guard called at the top of `push`, `pull`, `fetch`, and `sync`. (Putting it inside `synchronizing` is tempting but insufficient: `push`/`pull`/`fetch` are public methods callers invoke directly without going through `synchronizing`, and `sync` enters `synchronizing` *itself*, so the guard must run before the context manager body.) The guard raises `CatalogConfigurationError` when `len(self._git_remotes) > 1` and is a no-op otherwise. The aggregation loop in `push()` (`catalog.py:336-365`) collapses to a single-remote push-and-collect.
+
+Add a `Catalog.set_remote(name, url)` method that replaces (rather than appends) the catalog's git remote. This is the supported way to configure a remote: it makes the contract discoverable instead of relying on users to know that bare `git remote add` is forbidden. Raw `git remote add` against the working tree remains the configuration that triggers `CatalogConfigurationError`.
+
+The existing test `test_push_aggregates_failures_across_remotes` (`test_catalog.py:210`) is marked `pytest.skip("multi-remote unsupported per ADR-0009")` rather than deleted, so the bare-repo + two-remote + diverged-history setup is preserved against any future multi-remote re-introduction. A new test asserts `CatalogConfigurationError` is raised when a second git remote is added.
+
+### Escape hatch
+
+Users who have a real multi-remote use case are directed (in the error message and in the catalog docs) to open an issue or contact us. We will design the supported semantics in collaboration with a concrete workflow rather than speculatively.
+
+There is no in-band escape hatch. Dropping to raw `git push` bypasses `assert_consistency`, does not push the `git-annex` branch unless explicitly named, and does not run `git annex copy --to=<remote>` for content mirroring — meaning the catalog's invariants do not hold and downstream `Catalog.fetch_entries` against that remote may find metadata without content. We mention this so users do not reach for it as a workaround thinking it's "almost supported."
+
+## Alternatives considered
+
+### Keep multi-remote with the aggregation fix from #1898
+
+Continue accepting any number of git remotes; rely on the push-aggregation fix to surface failures.
+
+Rejected because:
+- The aggregation fix only addresses *push* failures. `pull`, `fetch`, `assert_consistency`, and annex content placement still have undefined behavior across multiple remotes.
+- It implicitly promises a feature ("multi-remote works") we have not designed, tested, or run.
+- Each subsequent multi-remote bug becomes an obligation to patch a surface we did not intend to support.
+
+### Support multi-remote with documented warnings
+
+Keep the implementation, document the sharp edges in the catalog guide, and treat each multi-remote failure as a separate triage item.
+
+Deferred. This is what we have today minus the explicit warning. It pushes design work onto users (who must reason about the inconsistency surface themselves) and onto future maintainers (who must keep patching aggregation-style fixes without a guiding model). We can revisit this if real users show up with use cases — see the escape hatch above.
+
+### Refuse zero remotes too
+
+Require exactly one git remote at all times.
+
+Rejected because the local-only case (a freshly initialized catalog before any remote is added, or a purely local workflow during development) is genuinely useful and exercised by tests. Zero remotes today is a silent no-op for `push`/`pull`/`fetch`; the supported-configuration section above commits to keeping that behavior, since the user has not asked the catalog to talk to anything.
+
+### Keep the surface, gate it behind a flag
+
+Allow multi-remote only when an explicit `multi_remote=True` is passed to the catalog constructor, treating it as an opt-in unsupported configuration.
+
+Rejected because flags-for-unsupported-features tend to drift into de facto support — users find them, build on them, and the flag becomes load-bearing without ever being designed. A clear error with a "contact us" path is more honest.
+
+### Warn on multi-remote, keep operating
+
+Print a deprecation-style warning the first time a multi-remote catalog is opened, but allow `push`/`pull`/`fetch`/`sync` to run with the current aggregation behavior.
+
+Rejected on the same grounds as the flag option: a warning that's easy to ignore is functionally support. It also keeps the aggregation surface alive without committing us to the rest of the multi-remote semantics enumerated in Context, which is the worst of both worlds — we'd accumulate bug reports against a configuration we tell users is "warned" rather than supported.
+
+## Consequences
+
+### Positive
+
+- The catalog's public surface matches what we test and run. No more divergence between "what the code accepts" and "what we will actually support."
+- `Catalog.push()` simplifies — no aggregation loop, no per-remote partitioning of push results. (The `main`/`git-annex` two-branch half-pushed-state hazard remains and still requires its own documentation, but is now contained to one remote.)
+- Users who hit the limit hear from us at the first `push`/`pull`/`fetch`/`sync` call after misconfiguration, not from a confusing partial-success three pushes later.
+- `Catalog.set_remote()` makes the supported configuration path discoverable; users do not have to learn by hitting the error.
+- We get a forcing function: if and when a real multi-remote use case lands, we design for it deliberately rather than retrofitting.
+
+### Negative
+
+- Users who today rely on multi-remote (we believe there are none, but we have not surveyed) will see a hard error after upgrading. Mitigation: the error message names the remotes and links to a contact path; the local-only path (zero remotes) is preserved.
+- The aggregation fix from `6b2ab300` is reverted as part of this change. If multi-remote support is later re-introduced, the aggregation logic would need to be re-derived (~30 lines); the test is preserved via `pytest.skip` so the bare-repo + two-remote + diverged-history setup is not lost.
+- "One git remote" is enforced at the catalog API boundary, not at the underlying git repository. A user who runs `git remote add` directly on the working tree and then operates on the repo with raw git can still create the inconsistent states described in Context. We accept this — the catalog is opinionated, the underlying git repo is not. The error fires on every subsequent `push`/`pull`/`fetch`/`sync` call, not just the first one, so the misconfiguration is loud and persistent rather than one-shot.
+
+## References
+
+- Issue [#1898](https://github.com/letsql/xorq/issues/1898) — silent push rejection
+- Commit `fe3f608b` — raise on rejected push
+- Commit `6b2ab300` — aggregate push failures across all remotes (to be reverted by this ADR)
+- Commit `0b40bb35` — multi-remote aggregation test (to be `pytest.skip`-ed and preserved against future multi-remote re-introduction; a new test asserting `CatalogConfigurationError` replaces it as live coverage)
+- [ADR-0003](0003-optional-git-annex-backend.md) — git-annex backend (related; unaffected by this decision since it concerns the *special* remote, not git remotes)
--- a/docs/adr/0009-catalog-single-git-remote.md
+++ b/docs/adr/0009-catalog-single-git-remote.md
@@ -1,49 +1,45 @@
 # ADR-0009: Catalog supports a single git remote
 
-- **Status:** Proposed
-- **Date:** 2026-04-30
-- **Context area:** `python/xorq/catalog/catalog.py`
+- **Status:** Accepted
+- **Date:** 2026-05-05
+- **Deciders:** Paddy Mullen
 
 ## Context
 
-A `Catalog` is a git repository — usually with a git-annex sidecar branch — that holds versioned build artifacts. `Catalog.push()`, `Catalog.pull()`, and `Catalog.fetch()` operate over `self._git_remotes`, which is *every* git remote on the underlying repo that has a fetch refspec (see the `Catalog._git_remotes` property in `python/xorq/catalog/catalog.py`). Nothing in the catalog API restricts a user to one remote; if `git remote add r2 …` is run on the working tree, the next `catalog.push()` will dutifully push to both `origin` and `r2`.
+A `Catalog` is a git repository — usually with a git-annex sidecar branch — that holds versioned build artifacts. `Catalog.push()`, `Catalog.pull()`, `Catalog.fetch()`, and `Catalog.sync()` operate over `Catalog._git_remotes`, which is *every* git remote on the underlying repo with a fetch refspec (see `python/xorq/catalog/catalog.py`). Nothing in the catalog API restricts a user to one remote: running `git remote add r2 …` on the working tree silently expands the next `Catalog.push()` to push to both `origin` and `r2`.
 
-Issue [#1898](https://github.com/letsql/xorq/issues/1898) surfaced a silent failure mode in this design: when a push to one remote was rejected (non-fast-forward), `catalog.push()` returned without raising. The fix raises `CatalogPushError` whenever `PushInfo` flags include `REJECTED`/`REMOTE_REJECTED`/`REMOTE_FAILURE`/`ERROR`, and — to support multi-remote — pushes every ref to every remote first, then aggregates all rejections into one error message rather than bailing on the first failure. A two-remote partial-failure test covers the aggregation behavior.
+Issue [#1898](https://github.com/letsql/xorq/issues/1898) surfaced a silent failure mode in this design: a non-fast-forward push was rejected, and `Catalog.push()` returned without raising. The fix raises `CatalogPushError` on `REJECTED`/`REMOTE_REJECTED`/`REMOTE_FAILURE`/`ERROR` flags. Writing the multi-remote variant of that fix — push to all remotes first, aggregate failures, then raise — surfaced an open-ended set of multi-remote consistency questions we have no design for:
 
-The aggregation fix exists because the obvious "fail fast on the first rejected push" implementation has a worse failure mode than the bug it replaces: with two git remotes, rejecting on `origin` skips the push of `origin`'s annex branch *and* the entire push to `r2`, so a retry sees inconsistent state across the two remotes. Aggregation patches that — but it patches *one* of an open-ended set of multi-remote consistency problems we have not enumerated:
-
-- **Atomicity (between remotes).** `push()` to N remotes is N independent operations. Any non-empty subset can succeed while the rest fail, leaving remotes out of sync. Aggregating the error tells the user *that* this happened, not how to recover. Note: single-remote does not give us atomicity either — `main` and `git-annex` are pushed in two separate operations inside `Catalog.push`, so even one remote can be left half-pushed. Single-remote shrinks the inconsistency surface from `2N` operations to `2`; it does not eliminate it.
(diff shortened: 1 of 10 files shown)


python/xorq/catalog/catalog.py
```diff
@@ -313,13 +344,23 @@ def fetch_entries(self, *entries):
...
+        Raises ``CatalogPushError`` if any remote rejects the push (e.g.
+        the branch has diverged). Both the main branch and the annex
+        branch are checked — partial success leaves the remote in a
+        half-state and is also reported as a failure.
+        """
         self.assert_consistency()
-        results = tuple(map(Remote.push, self._git_remotes))
+        results = tuple(remote.push() for remote in self._git_remotes)
+        for remote, infos in zip(self._git_remotes, results):
+            _assert_push_accepted(infos, remote.name)
```
dlovell: we should either push-and-assert one by one or raise for all failed remotes

dlovell (changes requested):

python/xorq/catalog/tests/test_catalog.py
```diff
@@ -178,6 +178,34 @@ def test_catalog_clone_from_push(repo_cloned_bare, tmpdir):
...
+    )
+
+    with build_expr_context_zip(xo.memtable({"a-only": ["a"]})) as zp:
+        user_a.add(zp, sync=False)
+    user_a.push()
+
+    with build_expr_context_zip(xo.memtable({"b-only": ["b"]})) as zp:
+        user_b.add(zp, sync=False)
+
+    with pytest.raises(Exception, match="(?i)reject"):
```
dlovell: `with pytest.raises(CatalogPushError, match="(?i)reject")`

python/xorq/catalog/catalog.py
paddymul: made ADR 0009 which puts in "We only support one remote" decision

paddymul:
there are tests for this, but I also wrote a simple reproduction script to see what the errors look like, I commented out the first section in the second run
```
"""Demo: what exceptions does Catalog.push() actually raise?

Two scenarios on PR #1899's branch:
  A. non-fast-forward rejection -> CatalogPushError (the #1898 fix)
  B. transport failure (unreachable remote) -> git.exc.GitCommandError (uncaught)

Run:  python demo_push_exceptions.py
"""

import shutil
import tempfile
import traceback
from pathlib import Path

from git import Repo as GitRepo
from git.exc import GitCommandError

import xorq.api as xo
from xorq.catalog import CatalogPushError
from xorq.catalog.catalog import Catalog
from xorq.catalog.constants import MAIN_BRANCH
from xorq.catalog.expr_utils import build_expr_context_zip


def banner(title):
    line = "=" * 72
    print(f"\n{line}\n{title}\n{line}")


def show_exc(exc):
    print(f"  exception type: {type(exc).__module__}.{type(exc).__name__}")
    print(f"  isinstance CatalogPushError? {isinstance(exc, CatalogPushError)}")
    msg = str(exc).strip().replace("\n", "\n    ")
    print(f"  str(exc):\n    {msg}")


def scenario_a_non_ff_rejection(workdir):
    banner("A. Non-fast-forward rejection (the #1898 case)")

    bare = workdir / "bare"
    GitRepo.init(bare, bare=True, initial_branch=MAIN_BRANCH)

    seed_path = workdir / "seed"
    seed = Catalog.from_repo_path(seed_path, init=True, annex=False)
    seed.set_remote("origin", str(bare))
    seed.repo.git.push("-u", "origin", MAIN_BRANCH)

    user_a_path = workdir / "user_a"
    user_b_path = workdir / "user_b"
    GitRepo.clone_from(str(bare), user_a_path)
    GitRepo.clone_from(str(bare), user_b_path)

    user_a = Catalog.from_repo_path(user_a_path, init=False, annex=False)
    user_b = Catalog.from_repo_path(user_b_path, init=False, annex=False)

    print("user_a.add(...) + push -> clean fast-forward")
    with build_expr_context_zip(xo.memtable({"a-only": ["a"]})) as zp:
        user_a.add(zp, sync=False)
    user_a.push()
    print("  ok")

    print("user_b.add(...) + push -> origin/main has moved, expect rejection")
    with build_expr_context_zip(xo.memtable({"b-only": ["b"]})) as zp:
        user_b.add(zp, sync=False)
        user_b.push()

def scenario_b_bad_transport(workdir):
    banner("B. Transport failure (remote URL points nowhere)")

    bare = workdir / "bare"
    GitRepo.init(bare, bare=True, initial_branch=MAIN_BRANCH)

    repo_path = workdir / "lonely"
    catalog = Catalog.from_repo_path(repo_path, init=True, annex=False)
    catalog.set_remote("origin", str(bare))
    catalog.repo.git.push("-u", "origin", MAIN_BRANCH)

    with build_expr_context_zip(xo.memtable({"x": ["x"]})) as zp:
        catalog.add(zp, sync=False)

    bogus_url = f"file://{workdir}/does-not-exist"
    catalog.repo.git.remote("set-url", "origin", bogus_url)
    print(f"remote set-url origin -> {bogus_url}")
    print("catalog.push() -> expect git transport error")

    catalog.push()

def main():
    # workdir = Path(tempfile.mkdtemp(prefix="xorq-push-demo-"))
    # try:
    #     print("scenario_a_non_ff_rejection")
    #     scenario_a_non_ff_rejection(workdir / "a")

    # finally:
    #     shutil.rmtree(workdir, ignore_errors=True)
    workdir = Path(tempfile.mkdtemp(prefix="xorq-push-demo-"))
    print("=" *80)
    try:
        print("scenario_b_bad_transport")
        scenario_b_bad_transport(workdir / "b")
    finally:
        shutil.rmtree(workdir, ignore_errors=True)

if __name__ == "__main__":
    main()
```

paddymul:
Failure #1 
```
scenario_a_non_ff_rejection

========================================================================
A. Non-fast-forward rejection (the #1898 case)
========================================================================
user_a.add(...) + push -> clean fast-forward
/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/python/xorq/ibis_yaml/translate.py:595: UserWarning: The Read op path is using a local filesystem path, running the build may not work in other environments.
  warn_on_local_path(op.read_kwargs)
Building wheel...
Successfully built /var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/tmp8tzd6nm8/xorq-0.3.21-py3-none-any.whl
  ok
user_b.add(...) + push -> origin/main has moved, expect rejection
/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/python/xorq/ibis_yaml/translate.py:595: UserWarning: The Read op path is using a local filesystem path, running the build may not work in other environments.
  warn_on_local_path(op.read_kwargs)
Building wheel...
Successfully built /var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/tmp7ijxkgtv/xorq-0.3.21-py3-none-any.whl
Error lines received while fetching: error: failed to push some refs to '/var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/xorq-push-demo-lvvjnuds/a/bare'
Traceback (most recent call last):
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 105, in <module>
    main()
    ~~~~^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 92, in main
    scenario_a_non_ff_rejection(workdir / "a")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 65, in scenario_a_non_ff_rejection
    user_b.push()
    ~~~~~~~~~~~^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/python/xorq/catalog/catalog.py", line 408, in push
    raise CatalogPushError(f"push failed: {'; '.join(failures)}")
```

paddymul:
Failure 2
```
scenario_b_bad_transport

========================================================================
B. Transport failure (remote URL points nowhere)
========================================================================
/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/python/xorq/ibis_yaml/translate.py:595: UserWarning: The Read op path is using a local filesystem path, running the build may not work in other environments.
  warn_on_local_path(op.read_kwargs)
Building wheel...
Successfully built /var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/tmpksqc6g2o/xorq-0.3.21-py3-none-any.whl
remote set-url origin -> file:///var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/xorq-push-demo-wsjfnq_w/b/does-not-exist
catalog.push() -> expect git transport error
Traceback (most recent call last):
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 105, in <module>
    main()
    ~~~~^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 100, in main
    scenario_b_bad_transport(workdir / "b")
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/demo_push_exceptions.py", line 86, in scenario_b_bad_transport
    catalog.push()
    ~~~~~~~~~~~~^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/python/xorq/catalog/catalog.py", line 394, in push
    main_result = remote.push()
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/.venv/lib/python3.13/site-packages/git/remote.py", line 1210, in push
    return self._get_push_info(proc, progress, kill_after_timeout=kill_after_timeout)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/.venv/lib/python3.13/site-packages/git/remote.py", line 975, in _get_push_info
    proc.wait(stderr=stderr_text)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/Users/paddy/.claude-worktrees/xorq/pr-1899-demo/.venv/lib/python3.13/site-packages/git/cmd.py", line 423, in wait
    raise GitCommandError(remove_password_if_present(self.args), status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git push --porcelain -- origin
  stderr: 'fatal: '/var/folders/_h/8prtx3692_16pzx62n6rjptm0000gn/T/xorq-push-demo-wsjfnq_w/b/does-not-exist' does not appear to be a git repository
fatal: Could not read from remote repository.'

```

docs/adr/0011-catalog-single-git-remote.md
dlovell, line 1: make sure `docs/adr/template.md` is in context and review this

python/xorq/catalog/catalog.py
```diff
@@ -283,18 +313,48 @@ def list(self):
...
+        if len(self._git_remotes) > 1:
+            names = ", ".join(r.name for r in self._git_remotes)
+            raise CatalogConfigurationError(
+                f"catalog supports a single git remote (ADR-0009); "
+                f"found {len(self._git_remotes)}: {names}. "
+                f"Use Catalog.set_remote(name, url) to replace existing remotes, "
+                f"or open an issue if you have a multi-remote use case."
+            )
+
+    def set_remote(self, name, url):
```
dlovell: should set_remote silently delete remotes if it is not in the expected {0,1} remote state?

docs/adr/0009-catalog-single-git-remote.md
```diff
@@ -0,0 +1,131 @@
...
+## Decision drivers
+
+- The catalog API is small and recent; we have room to constrain it before users build workflows that depend on multi-remote behavior.
+- Single-remote is the only configuration we test, document, or run in production.
+- Failure modes that "look like they work" are worse than failure modes that raise loudly.
+- We want to hear from users with this need rather than guess at requirements.
+
+## Decision
+
+The catalog officially supports exactly one git remote. Configurations with two or more git remotes are unsupported, and the catalog will refuse to operate on them rather than attempting best-effort multi-remote semantics.
```
dlovell: is supporting a catalog with no remotes in conflict with the language of "exactly one"?

docs/adr/0009-catalog-single-git-remote.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Context
+
+A `Catalog` is a git repository — usually with a git-annex sidecar branch — that holds versioned build artifacts. `Catalog.push()`, `Catalog.pull()`, and `Catalog.fetch()` operate over `self._git_remotes`, which is *every* git remote on the underlying repo that has a fetch refspec (see the `Catalog._git_remotes` property in `python/xorq/catalog/catalog.py`). Nothing in the catalog API restricts a user to one remote; if `git remote add r2 …` is run on the working tree, the next `catalog.push()` will dutifully push to both `origin` and `r2`.
+
+Issue [#1898](https://github.com/letsql/xorq/issues/1898) surfaced a silent failure mode in this design: when a push to one remote was rejected (non-fast-forward), `catalog.push()` returned without raising. The fix landed in three commits on this branch:
+
+| Commit | Change |
+|--------|--------|
+| `fe3f608b` | Raise `CatalogPushError` when `PushInfo` flags include `REJECTED`/`REMOTE_REJECTED`/`REMOTE_FAILURE`/`ERROR` |
```
dlovell: let's avoid commit shas
if necessary, we can use pr and issue numbers

mesejo:
## roborev: Fail

<details>

The `sync` method at line 348-351 calls `synchronizing()` which calls `pull()` then `push()` — both now guarded. So `sync` gets the guard transitively, which is consistent with the test passing.

## Review Findings

- **Severity**: Medium
- **Location**: `python/xorq/catalog/catalog.py:361–375` (new `push`)
- **Problem**: `self._git_remotes` is a property (re-evaluated each call). `push()` calls it three times: inside `_assert_at_most_one_git_remote()`, in `if not self._git_remotes:`, and in `(remote,) = self._git_remotes`. If git config changes between calls (e.g. concurrent `git remote add` on the working tree), the unpack raises `ValueError: too many values to unpack` instead of `CatalogConfigurationError`.
- **Fix**: Cache once — `remotes = self._git_remotes` — and use that local variable throughout.

---

- **Severity**: Medium
- **Location**: `python/xorq/catalog/catalog.py:307` (new `_git_remotes`)
- **Problem**: Exception catch narrowed from `except (KeyError, Exception)` (effectively `except Exception`) to `except (NoOptionError, NoSectionError)`. Any exception from `remote.config_reader.get("fetch")` other than configparser exceptions — e.g. a GitPython `GitCommandError` or `IOError` on a corrupt config — now propagates out of `_git_remotes` as an unhandled exception rather than treating the remote as not having a fetch refspec.
- **Fix**: Either add the broader exception as a fallback (`except Exception: return False`) with a logged warning, or confirm that GitPython's `config_reader.get` only raises `NoOptionError`/`NoSectionError` and document that assumption.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py:318–321` (`_assert_at_most_one_git_remote`)
- **Problem**: `self._git_remotes` is evaluated twice — once for `len()`, once to collect names in the error message. If a remote is removed between the two calls, the count in the message won't match the listed names. Minor inconsistency in error output.
- **Fix**: Cache: `remotes = self._git_remotes; if len(remotes) > 1: names = ...`.

---

- **Severity**: Low
- **Location**: `docs/adr/0009-catalog-single-git-remote.md:3`
- **Problem**: ADR status is "Proposed" but the implementation is merged in this PR. Stale status makes the doc misleading as a future reference.
- **Fix**: Change to `**Status:** Accepted`.

---

## Summary

Implements ADR-0009 single-git-remote constraint: adds `CatalogConfigurationError` for multi-remote configurations, raises `CatalogPushError` on rejected pushes, introduces `Catalog.set_remote()` as the safe reconfiguration path, and backs the decision with an ADR and new tests.

</details>

---
*Review type: default | Agent: claude-code | Job: 6*

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: High
**Location**: `python/xorq/catalog/catalog.py` — `set_remote` method (diff lines +339–348)
**Problem**: `set_remote(name, url)` has no `force` parameter, yet two tests depend on it. `test_set_remote_refuses_when_remote_exists` expects `CatalogConfigurationError` when called without `force=True` on a catalog that already has a remote; `test_set_remote_force_replaces_existing` calls `catalog.set_remote("origin", str(bare_2), force=True)`. The implementation unconditionally deletes all remotes and creates a new one — the guard-without-force behavior is entirely absent. Both tests will fail: the first because no error is raised, the second with `TypeError: set_remote() got an unexpected keyword argument 'force'`.

**Fix**: Add `force=False` parameter; when `force is False` and `self._git_remotes` is non-empty, raise `CatalogConfigurationError` naming the existing remote(s) and mentioning `force=True`.

---

**Severity**: High
**Location**: `python/xorq/catalog/cli.py` — `set_remote` command (diff lines +422–447)
**Problem**: The CLI command defines no `--force` option, but `test_set_remote_cli_refuses_when_remote_exists` expects `exit_code != 0` without `--force`, and `test_set_remote_cli_force_replaces_existing` passes `--force` as an argument. Click will treat `--force` as an unknown option and error with exit code 2 regardless of the catalog's behavior, so the tests' assertions about catalog semantics are never reached.

**Fix**: Add `@click.option("--force", is_flag=True, default=False, ...)` to the `set-remote` command and pass it through to `catalog.set_remote(name, url, force=force)`.

---

**Severity**: Medium
**Location**: `python/xorq/catalog/catalog.py` — `set_remote` method
**Problem**: The delete loop and `create_remote` have no atomicity. If `create_remote` raises (e.g. invalid URL, git error), all previously existing remotes are already gone and the catalog is left with zero remotes. No rollback.

**Fix**: Collect existing remotes before any mutation; delete only after the new remote is confirmed created, or hold the old remote's name/url and re-create on failure.

---

**Severity**: Low
**Location**: `docs/adr/0009-catalog-single-git-remote.md` — Status field (line 3)
**Problem**: Status is `Proposed` but the implementation is being merged in this PR.

**Fix**: Change to `Accepted`.

## Summary

Enforces ADR-0009's single-git-remote constraint via `_assert_at_most_one_git_remote` guards on `push`/`pull`/`fetch`/`sync`, raises `CatalogPushError` on rejected pushes, and adds `set_remote` as the supported reconfiguration path — but the `force` parameter required by the tests is missing from both `set_remote` and the CLI `set-remote` command, making four new tests fail at runtime.

</details>

---
*Review type: default | Agent: claude-code | Job: 34*

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `python/xorq/catalog/catalog.py`, `push()` (new implementation)
- **Problem**: TOCTOU — `_git_remotes` is a live property computed on each access. `push()` calls it three times: inside `_assert_at_most_one_git_remote()`, in `if not self._git_remotes`, and in `(remote,) = self._git_remotes`. If a second remote is added (e.g. by a concurrent `git remote add` or a test fixture) between the assert and the tuple-unpack, the unpack raises `ValueError: too many values to unpack` instead of `CatalogConfigurationError`. The same pattern appears in `set_remote` (two separate calls), where a concurrent add between the guard check and the delete loop could produce a partially-deleted state.
- **Fix**: Capture `remotes = self._git_remotes` once at the top of `push()` and `set_remote()`, then pass or iterate over the local variable.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/exceptions.py`
- **Problem**: `CatalogPushError` and `CatalogConfigurationError` share no common base beyond `RuntimeError`. Callers wanting to catch any catalog error must enumerate both types; adding a third exception later requires updating every call site.
- **Fix**: Add `class CatalogError(RuntimeError): pass` and derive both from it.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py`, `set_remote()` delete loop
- **Problem**: The `for existing in self._git_remotes: self.repo.delete_remote(existing)` loop has no error handling. If `delete_remote` raises mid-loop (e.g. git config write failure), subsequent remotes survive undeleted and `create_remote` still runs, leaving the repo in an inconsistent multi-remote state that the caller didn't request.
- **Fix**: Wrap in a `try/except` or collect all failures before proceeding; at minimum document that partial-delete on error leaves the repo dirty.

---

## Summary

Enforces ADR-0009's single-remote constraint across `push`/`pull`/`fetch`/`sync`, adds `set_remote` with an overwrite guard, raises on rejected pushes, and extracts catalog exceptions into a dedicated module — all with comprehensive test coverage.

</details>

---
*Review type: default | Agent: claude-code | Job: 36*

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py:337–340` (`synchronizing`)
- **Problem**: `sync()` calls `pull()` then `push()` via `synchronizing()`. Each now calls `_validated_git_remotes()` independently — two separate reads of `_git_remotes`. The docstring on `_validated_git_remotes` states "callers want a stable snapshot for the duration of a single operation." A concurrent `git remote add` between the two reads could slip past the guard on the second call. Very low probability but inconsistent with stated design intent.
- **Fix**: `sync` (or `synchronizing`) should read and validate remotes once at entry, then pass the snapshot to `pull`/`push` for the duration.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py:287–294` (`_has_fetch_refspec`)
- **Problem**: Exception narrowed from catch-all `except (KeyError, Exception)` to `except (NoOptionError, NoSectionError)`. Correct for GitPython ≥ 3.1.46 (configparser raises those, not `KeyError`). But if GitPython's `SectionConstrainedConfigParser` raises anything else on a corrupted or atypical config (e.g., `OSError` reading `.git/config`), it now propagates and breaks `_git_remotes` entirely rather than silently excluding the remote.
- **Fix**: Either keep the catch-all or add `except Exception` as a fallback with a log warning, so `_git_remotes` stays functional under unexpected config errors.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/exceptions.py` (new file)
- **Problem**: `CatalogConfigurationError` and `CatalogPushError` are not re-exported from `xorq.catalog`. External callers who need to catch them must import from the internal `xorq.catalog.exceptions` module path. The `__init__.py` is empty by design, but these are user-facing errors that appear in CLI-visible tracebacks.
- **Fix**: Add `from xorq.catalog.exceptions import CatalogConfigurationError, CatalogPushError` to `python/xorq/catalog/__init__.py`, or document `xorq.catalog.exceptions` as the public import path in the ADR.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py:379–417` (`push`)
- **Problem**: Return type changed from a flat `tuple[PushInfoList, ...]` (one entry per remote per push) to `tuple[PushInfoList, PushInfoList]` or `tuple[PushInfoList]`. Any caller that checks `len(results)` or indexes into results by remote will silently get different data. CLI ignores the return value; tests don't assert on it.
- **Fix**: Document the new return shape in the method docstring if public consumers exist; otherwise acceptable as-is.

## Summary

Enforces ADR-0009's at-most-one-git-remote constraint via `_validated_git_remotes`, adds `set_remote` with force-guard, raises `CatalogPushError` on rejected pushes, and extracts exceptions into a dedicated module — all with thorough test coverage.

</details>

---
*Review type: default | Agent: claude-code | Job: 38*

python/xorq/catalog/catalog.py
paddymul: I added a --force option to overwrite the existing remote

paddymul:
The first errors aren't worth adressing, they are to far fetched
> ## roborev: Fail
> Review findings
> ## Review Findings
> 
>     * **Severity**: Low
> 
>     * **Location**: `python/xorq/catalog/catalog.py:337–340` (`synchronizing`)
> 
>     * **Problem**: `sync()` calls `pull()` then `push()` via `synchronizing()`. Each now calls `_validated_git_remotes()` independently — two separate reads of `_git_remotes`. The docstring on `_validated_git_remotes` states "callers want a stable snapshot for the duration of a single operation." A concurrent `git remote add` between the two reads could slip past the guard on the second call. Very low probability but inconsistent with stated design intent.
> 
>     * **Fix**: `sync` (or `synchronizing`) should read and validate remotes once at entry, then pass the snapshot to `pull`/`push` for the duration.
> 
> 
>     * **Severity**: Low
> 
>     * **Location**: `python/xorq/catalog/catalog.py:287–294` (`_has_fetch_refspec`)
> 
>     * **Problem**: Exception narrowed from catch-all `except (KeyError, Exception)` to `except (NoOptionError, NoSectionError)`. Correct for GitPython ≥ 3.1.46 (configparser raises those, not `KeyError`). But if GitPython's `SectionConstrainedConfigParser` raises anything else on a corrupted or atypical config (e.g., `OSError` reading `.git/config`), it now propagates and breaks `_git_remotes` entirely rather than silently excluding the remote.
> 
>     * **Fix**: Either keep the catch-all or add `except Exception` as a fallback with a log warning, so `_git_remotes` stays functional under unexpected config errors.
> 
> 



These two comments are worth addressing, commit coming



>     * **Severity**: Low
> 
>     * **Location**: `python/xorq/catalog/exceptions.py` (new file)
> 
>     * **Problem**: `CatalogConfigurationError` and `CatalogPushError` are not re-exported from `xorq.catalog`. External callers who need to catch them must import from the internal `xorq.catalog.exceptions` module path. The `__init__.py` is empty by design, but these are user-facing errors that appear in CLI-visible tracebacks.
> 
>     * **Fix**: Add `from xorq.catalog.exceptions import CatalogConfigurationError, CatalogPushError` to `python/xorq/catalog/__init__.py`, or document `xorq.catalog.exceptions` as the public import path in the ADR.
> 
> 
>     * **Severity**: Low
> 
>     * **Location**: `python/xorq/catalog/catalog.py:379–417` (`push`)
> 
>     * **Problem**: Return type changed from a flat `tuple[PushInfoList, ...]` (one entry per remote per push) to `tuple[PushInfoList, PushInfoList]` or `tuple[PushInfoList]`. Any caller that checks `len(results)` or indexes into results by remote will silently get different data. CLI ignores the return value; tests don't assert on it.
> 
>     * **Fix**: Document the new return shape in the method docstring if public consumers exist; otherwise acceptable as-is.
>

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py`, `_format_push_failures`
- **Problem**: `info.local_ref` is a GitPython `Reference` object; `str(Reference)` gives `refs/heads/main`, not `main`. Error messages will read `origin/refs/heads/main: rejected (fetch first)` instead of `origin/main: …`, which is verbose and slightly confusing.
- **Fix**: Use `info.local_ref.name if info.local_ref else '?'` (or `str(info.local_ref).split('refs/heads/')[-1]`) to emit just the short branch name.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py`, `push()` lines 379–383
- **Problem**: `assert_consistency()` is called before the `if not remotes: return ()` guard. For zero-remote catalogs the consistency check runs (and could surface unrelated errors) even though `push` is a documented no-op in that mode. Intention is unclear — if you intend to skip the no-op silently, the guard should come first.
- **Fix**: Move `if not remotes: return ()` above `self.assert_consistency()`, or add a comment explaining why local consistency is checked even in the zero-remote case.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/tests/test_catalog.py`, `test_multi_remote_raises_configuration_error` (`op="sync"`)
- **Problem**: `sync` is not updated in the diff — it raises `CatalogConfigurationError` only indirectly (via `synchronizing` → `pull` → `_validated_git_remotes`). The ADR states the guard runs at the top of each public entry point. If `sync` is ever refactored to short-circuit `pull` (e.g., a no-pull variant), the guard silently disappears.
- **Fix**: Add `_validated_git_remotes()` at the top of `sync()` directly, consistent with the other three operations.

## Summary

Adds a single-remote enforcement contract to the Catalog API (ADR-0009): `push`/`pull`/`fetch` validate the remote count via `_validated_git_remotes`, `set_remote` replaces the raw `repo.create_remote` call, `push` now raises `CatalogPushError` on rejected pushes, and a dedicated `exceptions.py` module is extracted — all backed by comprehensive new tests.

</details>

---
*Review type: default | Agent: claude-code | Job: 39*

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py` — `_validated_git_remotes` error message
- **Problem**: Error message says "Use `Catalog.set_remote(name, url, force=True)` to replace existing remotes" but doesn't clarify that `set_remote` operates on `_git_remotes` (not `_validated_git_remotes`), so it is *not* blocked by the 2+ remote guard. A user hitting this error may not realize a single `set_remote(..., force=True)` call fixes the N→1 state.
- **Fix**: Add a brief note: "A single `set_remote(name, url, force=True)` call will delete all existing remotes and set one."

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/tests/test_catalog.py` — `test_set_remote_force_replaces_existing`
- **Problem**: Only tests the 1→1 remote replacement. The `set_remote(force=True)` path when `len(_git_remotes) > 1` (i.e., the recovery path for ADR-0009 violations created via raw `git remote add`) is untested. The delete-loop iterates a snapshot tuple so it's safe by inspection, but the case isn't exercised.
- **Fix**: Add a test that starts with 2 remotes created via `repo.create_remote`, calls `set_remote("origin", url, force=True)`, and asserts exactly one remote remains.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py` — `pull()`
- **Problem**: `pull()` silently ignores fetch/merge failures (returns raw `FetchInfo` results without inspection), while `push()` explicitly raises `CatalogPushError`. This asymmetry means a rejected pull (e.g., merge conflict) surfaces as a GitPython exception rather than a typed catalog error. Pre-existing gap, but this PR fixed the push side without addressing the pull side.
- **Fix**: Consider adding failure-flag inspection to `pull()` similar to `_format_push_failures`, raising `CatalogPushError` (or a new `CatalogPullError`) on `FetchInfo` error flags.

## Summary

Enforces ADR-0009 single-git-remote constraint across `push`/`pull`/`fetch`/`sync`, adds `set_remote` API and CLI command, raises `CatalogPushError` on rejected pushes, and extracts exceptions into a dedicated module — all well-tested and cleanly implemented.

</details>

---
*Review type: default | Agent: claude-code | Job: 40*

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `python/xorq/catalog/catalog.py` — `set_remote()`
- **Problem**: Non-atomic `force=True` path. The method deletes all existing remotes in a loop before calling `create_remote`. If `create_remote` raises (invalid name, git error), the repo is left with zero remotes and no rollback path.
- **Fix**: Collect all existing remote names first, create the new remote, then delete the old ones — or at minimum document that partial failure leaves zero remotes.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/catalog.py:443`
- **Problem**: `sync()` calls `_validated_git_remotes()` and discards the result. `pull()` and `push()` inside `synchronizing()` each call it again — three git config reads per `sync()` where two would suffice. The ADR documents the early-guard intent but the returned tuple is unused.
- **Fix**: Either use the returned value (pass to inner operations) or accept the waste — but document why the return value is discarded to avoid confusion.

---

- **Severity**: Low
- **Location**: `python/xorq/catalog/tests/test_catalog.py`
- **Problem**: `test_push_no_remote_skips_consistency_check` covers the zero-remote no-op for `push()` explicitly. `fetch()` and `pull()` have the same zero-remote no-op contract (both call `_validated_git_remotes()` which returns `()`) but lack analogous tests.
- **Fix**: Add `test_fetch_no_remote_is_noop` and `test_pull_no_remote_is_noop` mirroring the push test.

## Summary

Implements ADR-0009 by enforcing an at-most-one git remote constraint on catalog sync operations, adding `set_remote` with an overwrite guard, raising on rejected pushes, and extracting exceptions into a dedicated module — all well-tested and clearly documented.

</details>

---
*Review type: default | Agent: claude-code | Job: 43*

mesejo:
## roborev: Fail

<details>

## Review Findings

**Severity**: Low
**Location**: `python/xorq/catalog/catalog.py` — `sync()` + `push()` + `pull()`
**Problem**: `sync()` calls `_validated_git_remotes()` explicitly as a guard, but `pull()` and `push()` each call it again inside `synchronizing()` — 3 git-config reads per `sync()` call. The comment justifies it as a defensive guard, but the guard result is discarded, so it provides no snapshot stability guarantee either.
**Fix**: Either pass the validated remotes into `synchronizing` / `pull` / `push`, or drop the redundant top-level call in `sync()` and rely on the checks already present in `pull()` and `push()`.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tests/test_catalog.py`
**Problem**: Zero-remote no-op is tested for `push()` (`test_push_no_remote_skips_consistency_check`) but not for `fetch()` or `sync()`. The ADR explicitly guarantees all four operations are no-ops at zero remotes.
**Fix**: Add tests asserting `catalog.fetch() == ()` and `catalog.sync()` completes without error when `_git_remotes == ()`.

---

**Severity**: Low
**Location**: `python/xorq/catalog/tests/test_catalog.py:test_push_annex_branch_rejection`
**Problem**: `monkeypatch.setattr(type(cloned.repo.remotes.origin), "push", patched_push)` patches `git.remote.Remote.push` at the class level. If any other test runs concurrently in the same process (e.g., via `-n auto` with pytest-xdist) and creates a `Remote` instance, it picks up the patched method.
**Fix**: Use `monkeypatch.setattr(cloned.repo.remotes.origin, "push", ...)` to patch the instance rather than the class, keeping the change contained.

## Summary

Implements ADR-0011 — single git-remote constraint for the catalog — by adding `_validated_git_remotes()`, `set_remote()`, `CatalogPushError`/`CatalogConfigurationError` exceptions, a `set-remote` CLI command, and comprehensive tests covering rejection surfacing, multi-remote refusal, and recovery paths.

</details>

---
*Review type: default | Agent: claude-code | Job: 67*

dlovell (approved):

PR #1920, opened 2026-05-01
docs(tutorials): BSL semantic catalog + working-with-catalog tutorials (redo of #1878)

paddymul:
## Summary

Reintroduces the BSL semantic catalog tutorials originally landed in #1878 and reverted in #1919, rebased onto latest `main` for further iteration. Opening as a draft while work continues.

Two tutorials under `docs/tutorials/core_tutorials/`:
- `build_a_semantic_catalog.qmd` — building a catalog from scratch
- `working_with_the_catalog.qmd` — git-annex-centric workflow for catalog collaboration

Plus runnable snippet scripts and smoke tests in `python/xorq/tests/test_tutorial_snippets.py`.

## Test plan

- [ ] `pytest python/xorq/tests/test_tutorial_snippets.py` passes locally
- [ ] Tutorials render cleanly via Quarto
- [ ] Snippets execute from a fresh scratch directory (not just inside the xorq tree)


paddymul:
Heads-up: the **Working with the catalog** tutorial in this PR depends on a fix that hasn't shipped yet on PyPI.

The tutorial's `catalog.push()` step fails against the current PyPI release (`xorq==0.3.21`) because that version unconditionally pushes a `git-annex` ref even when the catalog was created in plain (non-annex) mode — which is the mode this tutorial uses:

\`\`\`
git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git push --porcelain -- origin git-annex
  stderr: 'error: src refspec git-annex does not match any
\`\`\`

The fix is on `main` (commit \`75d84d25\`, PR #1892 — *fix(catalog): guard annex branch push when branch is missing*), so the tutorial works end-to-end against `main`. **It will not work for readers on PyPI until the next release.** Suggest holding the merge / docs deploy until that release is out, or queuing this PR behind it.

mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `docs/tutorials/core_tutorials/build_a_semantic_catalog.snippets.py:30`, `working_with_the_catalog.snippets.py:27`
- **Problem**: Fixture `pyproject.toml` uses `"[tool.setuptools]\npackages = []\n"` but the tutorial prose prescribes `py-modules = []`. These are different directives: `packages = []` suppresses package discovery but not top-level module discovery. The tutorial explicitly explains that `py-modules = []` is required to prevent the multi-module wheel build failure when two `.py` files are present — the smoke test uses the wrong key and may not reproduce the reader's actual experience.
- **Fix**: Change `packages = []` to `py-modules = []` in both snippet fixtures.

---

- **Severity**: Medium
- **Location**: `python/xorq/tests/test_tutorial_snippets.py:63`
- **Problem**: `timeout=30` seconds. Each `catalog.add(...)` invocation builds a wheel (spawns a subprocess, runs setuptools, `egg_info`, etc.) and performs git commits. `working_with_the_catalog.snippets.py` calls `catalog.add(...)` twice and runs several extra `subprocess.run` git commands. 30 s is almost certainly too short on a cold CI runner.
- **Fix**: Raise timeout to at least 120 s (or 300 s for the catalog tutorial). Can be per-test if needed via a dict keyed by slug.

---

- **Severity**: Medium
- **Location**: `docs/tutorials/core_tutorials/working_with_the_catalog.snippets.py:170`
- **Problem**: `Profile.from_con(sqlite_con).save(alias="local_dev_sqlite", clobber=True)` writes a profile file and symlink to the user's persistent profile directory (default from `xo.options.profiles.profile_dir`). This side effect is never cleaned up in the `finally` block, so the `local_dev_sqlite` symlink accumulates in the real profile dir after every test run.
- **Fix**: Capture the returned path from `.save(...)` and delete it (and its target) in the `finally` block, or pass `profile_dir=_workdir / "profiles"` to scope the write to the temp directory.

---

- **Severity**: Low
- **Location**: `docs/tutorials/core_tutorials/working_with_the_catalog.snippets.py:136`
- **Problem**: `Catalog.clone_from(_BARE, _USER_B)` passes a `Path` object as the URL. The tutorial prose uses `f"file://{Path(...).expanduser()}"`. While gitpython's `Repo.clone_from` likely stringifies the path, passing a bare `Path` diverges from what the reader is instructed to do and may behave differently on some platforms.
- **Fix**: Pass `f"file://{_BARE}"` to match the tutorial's local-bare-repo tab.

---

- **Severity**: Low
- **Location**: `docs/tutorials/core_tutorials/working_with_the_catalog.snippets.py:33`
- **Problem**: `git init --bare -b main` requires Git ≥ 2.28 (released 2020-07). Older system git on some CI images would fail here with an unrecognised `-b` flag.
- **Fix**: Either assert the git version in the fixture or use the more portable `git symbolic-ref HEAD refs/heads/main` post-init as a fallback.

---

- **Severity**: Low
- **Location**: `docs/tutorials/core_tutorials/build_a_semantic_catalog.snippets.py:29`, `working_with_the_catalog.snippets.py:28`
- **Problem**: `requirements.txt` is written into the temp project dir but nothing in `catalog.add(...)` or the wheel builder reads it. It appears to be dead code in the fixture.
- **Fix**: Remove the `requirements.txt` write; it adds noise without contributing to test coverage.

---

## Summary

Adds two new BSL semantic-catalog tutorials (build + collaborate) with corresponding snippet-based smoke tests, but the smoke-test fixtures use the wrong setuptools key (`packages` instead of `py-modules`), have a too-short subprocess timeout for wheel builds, and leave a persistent profile artifact on disk after each run.

</details>

---
*Review type: default | Agent: claude-code | Job: 3*

mesejo:
## roborev: Fail

<details>

`sqlite` extra exists and includes `adbc-driver-sqlite`. Tutorial prose is accurate. Now checking the fixture `pyproject.toml` discrepancy from the diff more carefully — the snippet uses `packages = []` while the tutorial explains `py-modules = []` is critical.

## Review Findings

- **Severity**: Medium
- **Location**: `docs/tutorials/core_tutorials/build_a_semantic_catalog.snippets.py`, fixture block (~line 32)
- **Problem**: Fixture `pyproject.toml` uses `[tool.setuptools]\npackages = []\n` (suppresses package discovery) but the tutorial explicitly explains `py-modules = []` is required to suppress *module* auto-discovery. These are distinct setuptools directives. When `catalog.add(...)` runs inside the smoke test, setuptools sees the `.py` files in `_workdir` and may still fail on multiple top-level modules — the exact failure the tutorial warns about.
- **Fix**: Change the fixture's `pyproject.toml` string to use `py-modules = []` instead of `packages = []`, matching the tutorial instructions.

---

- **Severity**: Low
- **Location**: `python/xorq/tests/test_tutorial_snippets.py:20`
- **Problem**: `assert SCRIPTS, f"no *.snippets.py files..."` runs at collection time. If the glob finds nothing (e.g., a mis-configured `TUTORIALS_DIR`), pytest aborts collection with a bare `AssertionError` instead of a clean test failure.
- **Fix**: Replace with `pytest.skip` inside a fixture or use `pytest.mark.skipif` on the parametrize call; alternatively add a dedicated test that asserts `SCRIPTS` is non-empty so the failure is reported as a test failure, not a collection crash.

---

- **Severity**: Low
- **Location**: `docs/tutorials/core_tutorials/working_with_the_catalog.snippets.py`, fixture block (~line 43)
- **Problem**: `git init --bare -b main` requires Git ≥ 2.28 (the `-b` / `--initial-branch` flag). CI runners on Ubuntu 20.04 or older ship Git 2.25, which will reject the flag and fail the test.
- **Fix**: Drop `-b main` and instead run `git symbolic-ref HEAD refs/heads/main` after `git init --bare`, which works on all Git versions; or add a minimum-git-version check / skip.

---

- **Severity**: Low
- **Location**: `docs/tutorials/core_tutorials/working_with_the_catalog.snippets.py`, profile-swap block (~line 175)
- **Problem**: `Profile.from_con(sqlite_con).save(alias="local_dev_sqlite", clobber=True)` writes to a shared global profile store. When tests run in parallel (e.g., `pytest-xdist`), two workers racing to write the same alias could corrupt the profile file even with `clobber=True` if the write is not atomic.
- **Fix**: Use a unique alias per run (`f"local_dev_sqlite_{os.getpid()}"`) and clean it up in the `finally` block, or use a profile path inside `_workdir` if `Profile.save` accepts an explicit path.

---

- **Severity**: Low
- **Location**: `python/xorq/tests/test_tutorial_snippets.py:62`
- **Problem**: `timeout=120` may be too tight for a cold CI run that must build a wheel (the `catalog.add(...)` call invokes the setuptools wheel builder). The tutorial prose itself warns readers to expect wheel-build output. On resource-constrained CI runners this regularly exceeds 2 minutes.
- **Fix**: Raise to `timeout=300` or make it configurable via a pytest ini option / environment variable.

## Summary

Adds two Quarto tutorials (BSL semantic catalog + multi-user catalog collaboration) with matching smoke-test snippets; the main correctness risk is a `packages = []` vs `py-modules = []` mismatch between the fixture and the tutorial's own explanation, which could silently fail to reproduce the exact wheel-build constraint the tutorial warns about.

</details>

---
*Review type: default | Agent: claude-code | Job: 60*

mesejo (approved):

PR #1934, opened 2026-05-08
perf(tui): move SQL highlighting off the UI thread

mesejo:
Pygments lexing and Rich Syntax rendering previously ran synchronously on the main thread on every NodeHighlighted event, freezing the TUI for deeply nested CTEs (~1k+ lines). Fix:

- Add `_pygments_tokens()` that tokenises SQL with Pygments and builds a pre-styled `rich.Text` using `XorqSQLStyle.style_for_token()`, cached via `lru_cache(maxsize=256)` keyed on the SQL string.
- Add `_render_sql_text()` helper that applies the 500-line fallback (plain text with a `-- syntax highlighting disabled` note) or delegates to `_pygments_tokens`.
- Add `_load_sql_preview` worker (`exclusive=True`, `group="sql_render"`) that runs off-thread, cancelling stale in-flight renders on rapid navigation.
- Pass multi-query `sqls` tuple to the worker so `_render_sql_dag` also runs off the UI thread.
- Guard worker result against `_current_sql_hash` and `is_attached` to prevent stale updates after navigation or screen dismissal.
- Add `priority=True` to `Binding("1", ...)` and `Binding("2", ...)` so DataTable focus no longer swallows the view-switch keys.

Closes XOR-305.


mesejo:
## roborev: Fail

<details>

## Review Findings

- **Severity**: Medium
- **Location**: `tui.py:100–114` (`_pygments_to_text`) and CSS `#sql-preview`
- **Problem**: `word_wrap=True` dropped. Original `Syntax` wrapped long lines; new `Text()` has no wrap directive. `Static` defaults to clipping on overflow. Long CTEs/subselects will be cut off.
- **Fix**: Add `overflow: fold;` or `overflow-x: auto;` to `#sql-preview` in CSS (line 1470), or set `text.no_wrap = False` on the returned `Text`.

---

- **Severity**: Medium
- **Location**: `tui.py:901–908` (`_load_sql_preview`)
- **Problem**: Same mutable `Text` object stored in cache and passed directly to `Static.update(rich_text)`. If Textual or Rich mutates the object during rendering (span resolution, cell-width wrapping), subsequent cache hits return a corrupted object.
- **Fix**: Pass a copy: `self.query_one(...).update(rich_text.copy())` at line 908, or cache the raw SQL string and build `Text` fresh each call.

---

- **Severity**: Low
- **Location**: `tui.py:514`, `_sql_render_cache`
- **Problem**: Dict grows unbounded. Large catalogs accumulate one `Text` per entry with no eviction.
- **Fix**: Cap at ~100 entries with `functools.lru_cache` on `_pygments_to_text`, or replace dict with an `OrderedDict` and pop oldest when `len > 100`.

---

- **Severity**: Low
- **Location**: `tui.py:890`, `@work(exit_on_error=False)`
- **Problem**: Worker errors silently swallowed. If `pygments_get_lexer("sql")` or `_pygments_to_text` raises (e.g., lexer install missing), SQL preview stays blank forever with no user feedback.
- **Fix**: Wrap in `try/except` inside the worker and call `call_from_thread` with a fallback `Static.update("(highlight error)")`.

---

## Summary

Moves SQL syntax highlighting off the UI thread to fix freezes on large CTEs, using a per-entry cache and staleness guard, but loses `word_wrap=True`, passes mutable cached `Text` objects directly to `Static.update`, and has an unbounded cache with silent error swallowing.

</details>

---
*Review type: default | Agent: claude-code | Job: 74*

hussainsultan (approved):

PR #1936, opened 2026-05-10
test(gizmosql): start GizmoSQL via gizmosql PyPI package, drop Docker

prmoore77:
## Summary

Replaces the Docker-managed GizmoSQL test container with the new [`gizmosql`](https://pypi.org/project/gizmosql/) PyPI package, which wraps the same server binary as a managed Python subprocess (download-on-first-use, free-port auto-pick, context-managed teardown).

- `python/xorq/backends/gizmosql/tests/conftest.py` — replace docker container fixture with `gizmosql.Server(...)`; read host/port/credentials from the server object.
- `compose.yaml` — drop the `gizmosql` service and its dedicated network; the gizmosql tests no longer participate in compose.
- `.github/workflows/ci-test.yml` — flip the gizmosql matrix entry from `services: gizmosql` to `services: false`.
- `pyproject.toml` — add `gizmosql>=1.26.0` to the existing `[gizmosql]` optional-dependencies group so `--extra gizmosql` brings it in.

## Why

The `gizmosql` PyPI package is released in lockstep with the GizmoSQL server, so tests always exercise the latest stable server without managing image tags. Local contributors no longer need Docker running for the GizmoSQL backend tests, and CI shaves a Docker service container off the gizmosql matrix entry. The package picks a free port on each run, so multiple test runs can coexist without stepping on a fixed `31337`.

## Test plan

- [ ] CI `linux / gizmosql` job is green (it'll exercise the new fixture end-to-end on the matrix Python versions).
- [ ] `uv run --no-sync pytest -m gizmosql -k 'not script_execution and not slow' python/xorq/backends/gizmosql/tests` passes locally with `--extra gizmosql --extra duckdb`.
- [ ] No remaining references to `gizmodata/gizmosql:latest`, port `31337`, or the `gizmosql` compose service in test infrastructure (the only remaining `31337` literals are in `test_coverage.py`, which exercises URL-parsing).

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -73,7 +73,7 @@ jobs:
               - --extra sqlite
               - --extra duckdb
           - name: gizmosql
-            services: gizmosql
+            services: false
             extras:
               - --extra gizmosql
               - --extra duckdb
--- a/compose.yaml
+++ b/compose.yaml
@@ -116,25 +116,9 @@ services:
       - $PWD/docker/trino/catalog/hive.properties:/etc/trino/catalog/hive.properties:ro
       - $PWD/docker/trino/jvm.config:/etc/trino/jvm.config:ro
 
-  gizmosql:
-    image: gizmodata/gizmosql:latest
-    environment:
-      GIZMOSQL_USERNAME: ibis
-      GIZMOSQL_PASSWORD: ibis_password
-      TLS_ENABLED: "1"
-      PRINT_QUERIES: "0"
-      DATABASE_FILENAME: ":memory:"
-    ports:
-      - 31337:31337
-    networks:
-      - gizmosql
-    volumes:
-      - ./ci/ibis-testing-data/parquet:/data/parquet:ro
-
 networks:
   postgres:
   trino:
-  gizmosql:
 
 
 volumes:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,8 @@ sqlite = [
 gizmosql = [
     "adbc-driver-gizmosql>=1.1.5",
     "duckdb>=1.1.3",
-    "gizmosql>=1.26.0",
+    "gizmosql>=1.26.0,<2",
+    "cryptography>=42",
 ]
 ibis = [
     "ibis-framework>=10.5.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,6 +131,7 @@ sqlite = [
 gizmosql = [
     "adbc-driver-gizmosql>=1.1.5",
     "duckdb>=1.1.3",
+    "gizmosql>=1.26.0",
 ]
 ibis = [
     "ibis-framework>=10.5.0",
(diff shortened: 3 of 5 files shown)


prmoore77:
Addressed both roborev findings in 8532309:

- **Medium / TLS coverage**: regenerated TLS path. Session-scoped fixture now mints a short-lived self-signed RSA cert (CN=`localhost`, SAN=`localhost`, ~1 day validity) via `cryptography`, writes the PEM cert + key into a pytest tmp dir, and passes `--tls <cert> <key>` through `gizmosql.Server.extra_args`. The client reconnects with `use_encryption=True`, `disable_certificate_verification=True` (cert is freshly minted per session and only sees loopback traffic, so a CA bundle would be overkill).
- **Low / unbounded gizmosql pin**: changed `gizmosql>=1.26.0` to `gizmosql>=1.26.0,<2` in the `[gizmosql]` extra to guard against future `Server` API changes on `.host` / `.port` / `.username` / `.password`.
- Added `cryptography>=42` to the `[gizmosql]` extra (needed by the new test fixture).

The encrypted Flight SQL path is now exercised end-to-end again, equivalent to the prior Docker setup.

prmoore77:
Addressed both round-2 roborev findings:

- **Medium / API surface assertion**: added a fail-fast `assert hasattr(srv, attr)` loop on `host` / `port` / `username` / `password` at fixture entry. The `gizmosql>=1.26.0,<2` pin still guards against major drift; this gives a clearer error than a generic `AttributeError` if a patch release ever renames one.
- **Low / unencrypted key comment**: added an inline comment on the key write explaining why `NoEncryption()` is intentional (server reads it at startup with no passphrase prompt, tmp dir is per-session + cleaned up, cert is loopback-only and ~1 day valid).

prmoore77:
Round-3 roborev findings addressed:

- **TLS CLI contract**: added a docstring note that the `--tls <cert> <key>` argument shape is covered by the `gizmosql>=1.26.0,<2` version pin in `pyproject.toml`.
- **Lost convenience fallback**: documented in the fixture docstring that the previous `_port_is_listening`-based reuse of a pre-running local GizmoSQL is gone — the fixture always starts its own subprocess on a freshly-picked free port.
- **Cert dir permissions**: `cert_dir.chmod(0o700)` immediately after `mktemp`, before the unencrypted key is written, so the directory is owner-only on shared CI runners.

prmoore77:
Round-4 roborev findings addressed:

- **PRINT_QUERIES parity**: explicitly pass `extra_env={"PRINT_QUERIES": "0"}` (matching the old Docker config) so a future default-flip in the binary can't silently change CI log volume.
- **Skip-not-error on API drift**: replaced the `assert hasattr(srv, ...)` block with `pytest.skip(...)` so a future `gizmosql` patch that renames a public attribute degrades the gizmosql test session to skips rather than fixture errors. The version pin (`<2`) is still the primary guard.

mesejo (approved):

PR #1943, opened 2026-05-11
fix: add expr_builder to KIND_ORDER, bump bsl 3.13

ghoersti:
Summary                                                                                     
                                                                                              
  - python/xorq/catalog/tui.py: add "expr_builder" to KIND_ORDER and add a matching           
  KIND_STYLES entry (⊡, secondary color). Without the styles entry, _styled_branch_label would
   KeyError for expr_builder rows.                                                            
  - pyproject.toml: bump the bsl extra from boring-semantic-layer>=0.1.0 to >=0.3.13. 0.3.13
  is the first BSL release that ships the xorq.from_tag_node entry point for the bsl tag;
  older versions silently degrade BSL entries to ExprKind.Expr.
  - uv.lock: refreshed (BSL 0.3.10 → 0.3.13).

### DIFF
--- a/python/xorq/catalog/tui.py
+++ b/python/xorq/catalog/tui.py
@@ -95,7 +95,7 @@ XORQ_DARK = Theme(
     },
 )
 
-KIND_ORDER = ("source", "expr", "unbound_expr", "composed")
+KIND_ORDER = ("source", "expr", "unbound_expr", "composed", "expr_builder")
 
 
 @frozen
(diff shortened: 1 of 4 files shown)


python/xorq/catalog/tui.py
```diff
@@ -95,7 +95,7 @@ class XorqSQLStyle(PygmentsStyle):
     },
 )
 
-KIND_ORDER = ("source", "expr", "unbound_expr", "composed")
+KIND_ORDER = ("source", "expr", "unbound_expr", "composed", "expr_builder")
```
dlovell: these should use `xorq.ibis_yaml.enums.ExprKind`

dlovell (changes requested):

dlovell (approved):

Issue #1940, opened 2026-05-11
xorq uv run emits cryptic "Invalid type NoneType for attribute 'limit'" warning when --limit is not passed


paddymul:
## Repro

```bash
xorq init --path moneyball-tutorial
cd moneyball-tutorial
# any minimal expr.py with an `expr` at module scope
xorq uv build expr.py -e expr --builds-dir builds
xorq uv run builds/<hash> --output-path out.parquet
```

Output:
```
Invalid type NoneType for attribute 'limit' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

The run succeeds and writes the parquet; the warning is harmless.

## Why it's worth fixing

1. **No context.** Nothing in the message identifies it as an OpenTelemetry warning. A first-time user reads "Invalid type NoneType" and assumes something is wrong with their expression or with a `--limit` argument they didn't pass.
2. **No suppression knob.** `xorq uv run --help` exposes no verbosity/quiet/log-level option, and no documented env var.
3. **Distracting in tutorials.** The init tutorial (`docs/getting_started/get-started-with-xorq-init.qmd`) has to spend a sentence explaining the warning is "harmless noise."

## Likely source

`python/xorq/cli.py` builds a `run_params` tuple including `("limit", limit)` and threads it through OTel. The `span.add_event` call on line 308 already filters None values, so the warning likely comes from another code path (decorator-attached span attributes, or `otel_utils`).

## Suggested fix

A fix was implement in PR 1747.  but the template the tutorial used pointed at xorq 0.3.11, prior to PR #1747 .

The fix for this particular bug is to point to a current version of xorq.

The problem of other unactionable error/warning messages from otel calls probably exists in other places in our code base. 

Strip None values before any OTel attribute assignment, or coerce `None → ""` at the params construction site. A user-facing `--quiet` flag (or respect for `OTEL_LOG_LEVEL`) would be a separate, smaller improvement.

PR #1944, opened 2026-05-11
fix(logging): drop None attribute values from RunLogger.log_span_event (#1940)

paddymul:
## Summary

When calling `add_event` with a dictionary with `None` values a warning is thrown that a user can't act on of:

```
Invalid type NoneType for attribute 'limit' value. Expected one of
['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

Call sites previously had to protect calls to `log_span_event` with dictionary comprehensions like
```
span.add_event("run.params", {k: v for k, v in run_params if v is not None})
```
This PR builds that dictionary comprehension into `log_span_event`.  it doesn't protect an `add_event` call like above.

### DIFF
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -305,9 +305,11 @@ def run_command(
         ("limit", limit),
     )
 
+    span.add_event("run.params", {k: v for k, v in run_params if v is not None})  # see PR review
+
     try:
         with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
-            rl.log_span_event(span, "run.start", dict(run_params))
+            rl.log_event("run.start", dict(run_params))  # see PR review
 
             with timed() as get_elapsed:
                 try:
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -307,6 +307,7 @@ def run_command(
 
     try:
         with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
+            # here to hang PR comment off of
             rl.log_span_event(span, "run.params", dict(run_params))
 
             with timed() as get_elapsed:
@@ -538,6 +539,7 @@ def run_unbound_command(
     cache_dir = _get_cache_dir(cache_dir)
 
     span = trace.get_current_span()
+    # here to hang PR comment off of
     params = {
         "expr_path": str(expr_path),
         "output_format": str(output_format),
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -307,7 +307,7 @@ def run_command(
 
     try:
         with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
-            rl.log_span_event(span, "run.start", dict(run_params))
+            rl.log_span_event(span, "run.params", dict(run_params))
 
             with timed() as get_elapsed:
                 try:
@@ -538,15 +538,15 @@ def run_unbound_command(
     cache_dir = _get_cache_dir(cache_dir)
 
     span = trace.get_current_span()
-    span.add_event(
-        "run_unbound.params",
-        {
-            "expr_path": str(expr_path),
-            "to_unbind_hash": str(to_unbind_hash),
-            "to_unbind_tag": str(to_unbind_tag),
-            "output_format": str(output_format),
-        },
-    )
+    params = {
+        "expr_path": str(expr_path),
+        "output_format": str(output_format),
+    }
+    if to_unbind_hash is not None:
+        params["to_unbind_hash"] = to_unbind_hash
+    if to_unbind_tag is not None:
+        params["to_unbind_tag"] = to_unbind_tag
+    span.add_event("run_unbound.params", params)
 
     # Resolve build identifier
     expr_path = ensure_build_dir(expr_path)
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -305,11 +305,9 @@ def run_command(
         ("limit", limit),
     )
 
-    span.add_event("run.params", {k: v for k, v in run_params if v is not None})
-
     try:
         with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
-            rl.log_event("run.start", dict(run_params))
+            rl.log_span_event(span, "run.start", dict(run_params))
 
             with timed() as get_elapsed:
                 try:
--- a/python/xorq/cli.py
+++ b/python/xorq/cli.py
@@ -305,11 +305,11 @@ def run_command(
         ("limit", limit),
     )
 
-    span.add_event("run.params", {k: v for k, v in run_params if v is not None})  # see PR review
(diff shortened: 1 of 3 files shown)


paddymul:
Replying to the roborev findings:

- **High — "implementation missing"**: stale review. It fired against the test-only commit (`7864f35b`) before the fix landed. The fix is now in `f3756eba`; ci-test-library is green on the fix commit. See [run 25689953728](https://github.com/xorq-labs/xorq/actions/runs/25689953728).
- **Low — JSON round-trip nit**: noted, not changing. The two `is None` assertions explicitly document that the file log preserves None values, which is the intended contract per the docstring.

Worth noting in passing: this fix is **defense-in-depth at one OTel surface**, not a universal None-attribute filter. A future regression that calls `span.add_event` or `span.set_attribute` directly (bypassing `log_span_event`) would still leak. A pytest fixture that fails on any `opentelemetry.attributes` warning would be the next-level lockdown — happy to do that as a follow-up if useful.

paddymul:
**Aside (out-of-diff observation — flagging since I noticed it next door)** — `cli.py:541-549` in `run_unbound_command`:

```python
span.add_event(
    "run_unbound.params",
    {
        "expr_path": str(expr_path),
        "to_unbind_hash": str(to_unbind_hash),
        "to_unbind_tag": str(to_unbind_tag),
        "output_format": str(output_format),
    },
)
```

`to_unbind_hash` and `to_unbind_tag` are both `Optional[str]` — exactly one is supplied at a time (hash *or* tag identifies the node), so the other defaults to `None`. Wrapping them unconditionally in `str()` records the literal string `"None"` for the unset attr in the OTel span — arguably worse than the `Invalid type NoneType` warning the `str()` was working around, since downstream consumers can't distinguish "unset" from "literally named None."

Optional cleanup — only include each key when set, matching the `serve.params` pattern at cli.py:668:

```python
params = {
    "expr_path": str(expr_path),
    "output_format": str(output_format),
}
if to_unbind_hash is not None:
    params["to_unbind_hash"] = to_unbind_hash
if to_unbind_tag is not None:
    params["to_unbind_tag"] = to_unbind_tag
span.add_event("run_unbound.params", params)
```

Out of scope for the core #1940 fix (None warning on `xorq run --limit`), but worth a follow-up. Originally added in 2caa9dd15 (#1479).

python/xorq/cli.py
```diff
@@ -305,11 +305,11 @@ def run_command(
         ("limit", limit),
     )
 
-    span.add_event("run.params", {k: v for k, v in run_params if v is not None})
+    span.add_event("run.params", {k: v for k, v in run_params if v is not None})  # see PR review
 
     try:
         with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
-            rl.log_event("run.start", dict(run_params))
+            rl.log_event("run.start", dict(run_params))  # see PR review
```
paddymul: **Codex (Medium):** _"The refactor silently drops the `run.params` OTel span event ... event name change is an observable breaking change."_

The deletion of `span.add_event("run.params", ...)` at line 308 and the swap of `rl.log_event(...)` → `rl.log_span_event(...)` at line 312 are **one connected refactor** — both fall out of moving the run-params emission into the new `log_span_event` helper. Codex flagged the rename specifically, but the deletion is the same root cause.

The two `# see PR review` end-of-line markers exist only to anchor this comment across the full 308–312 range; the suggested change replaces them with the real refactor.

Single suggested change — collapse both sites into the helper, but split the **span** event name (`run.params`) from the **file** audit event name (`run.start`) so each surface keeps the name it has on `main`:

```suggestion
    try:
        with RunLogger.from_expr_hash(expr_hash, params_tuple=run_params) as rl:
            rl.log_span_event(span, "run.params", dict(run_params))
            rl.log_event("run.start", dict(run_params))
```

Net behavior:
- Span: `run.params` event (codex addressed — no observable rename for OTel consumers)
- File `run.jsonl`: `run.params` (from `log_span_event`) **plus** `run.start` (from `log_event`) — original audit name preserved, with one extra `run.params` entry for completeness
- None-filter centralized in `log_span_event` — the #1940 fix in this PR applies to every future caller automatically

If accepted, drop both `# see PR review` markers when applying.

python/xorq/common/utils/logging_utils.py
```diff
@@ -212,10 +212,13 @@ def log_event(self, event: str, fields: dict = None):
         self._fh.flush()
 
     def log_span_event(self, span, event: str, fields: dict = None):
-        """Log to both the run log and an OTel span."""
+        """Log to both the run log and an OTel span. Cleans up OTel re #1940."""
```
dlovell: Drop the ticket reference — it'll be opaque once #1940 is closed. The behavior is clear from the code.

```suggestion
        """Log to both the run log and an OTel span."""
```

python/xorq/common/utils/tests/test_logging.py
```diff
@@ -84,6 +84,30 @@ def test_log_span_event_none_span(tmp_path):
     assert events[0]["event"] == "my.event"
 
 
+@pytest.mark.library
+def test_log_span_event_drops_none_attributes_from_span(tmp_path):
+    # Regression test for #1940: span.add_event must drop None attrs
+    # (OTel rejects them); file log keeps everything for audit.
```
dlovell: The test name already says what it tests, and `git blame` → commit message links back to #1940. Remove the inline comment — the PR description is the right place for this context.

```suggestion
```

dlovell (changes requested):
Core fix is correct, minimal, and well-tested — two small cleanup requests before merge.

paddymul:
This is a very minimal fix that wouldn't actually prevent against the problem raised in 1940.  Because this protects `log_span_event` not raw `add_event` which is what I think actually caused the warning, although at this point I'm not sure.


in the 0.3.11 release of xorq this call site was unprotected https://github.com/xorq-labs/xorq/blob/v0.3.11/python/xorq/cli.py#L194

That call site was updated with protection in PR 1747.  

I think that was the cause of the warning message displayed to the user on `xorq run`/`xorq run-cached`.  

The better fix for the tutorial was updating the template used to point at the most recent version of xorq.

@dlovell given this, do we want to update other call sites of `add_event` or do some patching of `add_event` to protect from this bug

dlovell:
> This is a very minimal fix that wouldn't actually prevent against the problem raised in 1940. Because this protects `log_span_event` not raw `add_event` which is what I think actually caused the warning, although at this point I'm not sure.
> 
> in the 0.3.11 release of xorq this call site was unprotected https://github.com/xorq-labs/xorq/blob/v0.3.11/python/xorq/cli.py#L194
> 
> That call site was updated with protection in PR 1747.
> 
> I think that was the cause of the warning message displayed to the user on `xorq run`/`xorq run-cached`.
> 
> The better fix for the tutorial was updating the template used to point at the most recent version of xorq.
> 
> @dlovell given this, do we want to update other call sites of `add_event` or do some patching of `add_event` to protect from this bug

please create issues for follow on work

dlovell (approved):

PR #1945, opened 2026-05-11
fix(packager): trust uv.lock; stop byte-comparing on-disk requirements.txt (#1941)

### LINKED ISSUE #1941: xorq uv build refuses to run when template-shipped requirements.txt doesn't byte-match uv.lock across uv versions
## Repro

```bash
xorq init --path tutorial
cd tutorial
xorq uv build expr.py -e expr --builds-dir builds
```

If the local `uv` version differs from the one that generated the template's commit-pinned `requirements.txt` and `uv.lock`, the build refuses to run with a byte-comparison error against `uv.lock`.

Workaround:
```bash
rm requirements.txt
```
After which `xorq uv build` re-generates `requirements.txt` from `uv.lock` and proceeds.

## Why it's worth fixing

1. **First-run failure.** This is among the very first commands a tutorial reader runs after `xorq init`. It teaches them that the tool is broken at minute one.
2. **`rm` as a setup step is unintuitive.** The reader is asked to delete a file the tool just created, with the explanation "byte-comparison error against uv.lock." That sounds like a bug because it is one.
3. **Tutorial pollution.** The init tutorial (`docs/getting_started/get-started-with-xorq-init.qmd`) has to spend a paragraph explaining and working around this.

## Likely fix paths

In rough order of how much they improve the UX:

1. **Don't ship `requirements.txt` in the init template.** Let `xorq uv build` regenerate it from `uv.lock` against the user's local uv at build time.
2. **Have `xorq uv build` regenerate `requirements.txt`** instead of byte-comparing the on-disk file. Same effect — works regardless of what the template ships.
3. **Regenerate `requirements.txt` during `xorq init`** using the user's local uv, so it matches `uv.lock` from the start.

All three remove the `rm` step from the tutorial.

paddymul:
## Summary

`xorq uv build` currently fails with a `RuntimeError: requirements.txt
... does not match` whenever the local `uv` version exports a slightly
different `requirements.txt` than the one shipped with the template.
This is the first command in the init tutorial after `xorq init` and
forces every reader to `rm requirements.txt` as a setup step.

The byte-comparison was a "did you forget to re-export?" safety check.
The cost is much higher than the benefit: `uv export` output is not
stable across uv versions, so the check breaks first-run reproducibly
across nearly any machine that wasn't used to publish the template.

## Fix

When `uv.lock` exists, the lockfile is authoritative. The packager
now generates a fresh export from the lock and bundles it in the
build directly. The user's on-disk `requirements.txt` is left
untouched — we don't rewrite their working tree.

## Why this is safe

- The build artifact already contains the freshly exported requirements
  pinned to `uv.lock`. Reproducibility comes from the lockfile, not
  from a stale on-disk export.
- The packager's existing test infra already worked around this bug by
  `unlink()`-ing `requirements.txt` before each invocation
  (`test_packager.py:prep_template_tmpdir`). That workaround can be
  removed in a follow-up; the slow-marker tests start exercising the
  new code path naturally.
- `_ensure_wheel_artifacts` and downstream consumers all read from the
  bundle's `requirements_path` (in the build dir), not from the user's
  working tree.

## Test plan

- [x] Failing test pushed first (TDD); confirmed failing on CI before fix.
- [x] After the fix, `test_write_requirements_path_prefers_export_over_stale_on_disk` passes.
- [x] Existing `@pytest.mark.slow` end-to-end packager tests
  (`test_wheel_packager`, `test_wheel_builder`, `test_wheel_runner`)
  continue to pass against all three templates.
- [ ] Manual: run `xorq init` then `xorq uv build` without the
  `rm requirements.txt` workaround. Build should succeed.

## Follow-up not in this PR

- `prep_template_tmpdir` still calls `stale_reqs.unlink()`. The
  workaround is no longer required after this fix — worth a quick
  cleanup in a separate change once this lands.
- The init tutorial currently documents the `rm requirements.txt`
  workaround with a link to #1941; that paragraph can come out once
  this ships in a release.

Fixes #1941.

### DIFF
--- a/python/xorq/ibis_yaml/packager.py
+++ b/python/xorq/ibis_yaml/packager.py
@@ -217,23 +217,19 @@ class WheelPackager:
         existing = self.project_path / DumpFiles.requirements
 
         if has_lockfile:
+            # uv.lock is authoritative. We regenerate `requirements.txt`
+            # from the lock here and bundle it in the build artifact.
+            # We don't compare against the user's on-disk file --- `uv
+            # export` output is not stable across uv versions, and any
+            # template shipped with a pre-baked requirements.txt would
+            # otherwise break the build on the first invocation by a
+            # user whose local uv differs (see #1941).
             with tracer.start_as_current_span("packager.uv_export_requirements"):
                 exported = uv_export_requirements(
                     self.project_path,
                     self.python_version,
                     extras=self.extras,
                     all_extras=self.all_extras,
-                )
-            if existing.exists() and existing.read_text() != exported:
-                raise RuntimeError(
-                    f"{DumpFiles.requirements} in {self.project_path} does not match "
-                    f"`uv export` output from {UVLOCK_NAME} (byte-exact comparison). "
-                    f"This happens when {UVLOCK_NAME} changes, or when the in-tree "
-                    f"{DumpFiles.requirements} was produced by a different uv version "
-                    f"than the one running now. To resolve: delete "
-                    f"{existing} and let the packager regenerate it, "
-                    f"or re-export manually with `uv export --locked --no-dev "
-                    f"--no-emit-project --no-header --no-annotate > {existing.name}`."
                 )
             (self.tmpdir / DumpFiles.requirements).write_text(exported)
         else:
--- a/python/xorq/ibis_yaml/tests/test_packager.py
+++ b/python/xorq/ibis_yaml/tests/test_packager.py
@@ -348,7 +348,14 @@ def test_wheel_bundle_from_build_path_no_wheel(tmp_path):
 # ---------------------------------------------------------------------------
 
 
-def test_write_requirements_path_rejects_stale_requirements(tmp_path, monkeypatch):
+@pytest.mark.library
+def test_write_requirements_path_prefers_export_over_stale_on_disk(
+    tmp_path, monkeypatch
+):
+    """When uv.lock exists, the lockfile is authoritative. An on-disk
+    `requirements.txt` exported by a different uv version must not break
+    the build --- the packager regenerates from the lock and uses that.
+    Regression for #1941."""
     _make_pyproject(tmp_path)
     (tmp_path / UVLOCK_NAME).write_text("# lockfile")
     (tmp_path / DumpFiles.requirements).write_text("requests==2.31.0\n")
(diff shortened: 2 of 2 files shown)


paddymul:
Replying to the roborev findings:

- **High — "implementation missing"**: stale review. It fired against the test-only commit (`78aa93d9`) before the fix landed. The fix is now in `d6de2863`; ci-test-library is green on the fix commit. See [run 25690267281](https://github.com/xorq-labs/xorq/actions/runs/25690267281).
- **Medium — "missing test coverage for the no-lockfile branch"**: valid, addressed in `81dc5e6e` with `test_write_requirements_path_copies_existing_when_no_lockfile`. Guards against accidentally re-routing the else-branch through `uv export`.
- **Low — "`@pytest.mark.library` excludes from fast CI"**: rebutted. `library` is the marker for ci-test-library, the standard PR-gate unit-test workflow. It's not a slow/optional subset — pyproject.toml's registered markers and ci-test-library.yml's `-m "library or xorq"` confirm this. Same convention used in #1944.

Follow-up not in this PR (already noted in the PR body): `prep_template_tmpdir`'s `stale_reqs.unlink()` workaround is no longer needed and can be cleaned up once this lands.

paddymul:
@dlovell — heads up that this PR makes part of [ADR-0008](https://github.com/xorq-labs/xorq/blob/main/docs/adr/0008-wheel-based-packaging-pipeline.md) stale.

The **"Sync-check is byte-exact"** section (around lines 43–50) still reads as the current design:

> When both \`uv.lock\` and \`requirements.txt\` are present in the project, \`WheelPackager._write_requirements_path\` re-runs \`uv export\` and compares its output byte-for-byte against the existing \`requirements.txt\`. Any divergence — including whitespace, ordering, or format changes — raises \`RuntimeError\`. **This is intentional** and it *will* trip…

After this PR lands that's no longer true — the packager trusts \`uv.lock\`, regenerates the export into the build dir, and leaves the user's on-disk \`requirements.txt\` alone. The ADR status header still says "Accepted," so a reader will treat the byte-check paragraph as current behavior.

Worth either updating that section in this PR or noting it as a follow-up so the ADR doesn't drift.

paddymul:
Forward-looking note: the reason we're still threading \`requirements.txt\` through the packager at all (rather than just shipping \`uv.lock\` or a \`pylock.toml\` in the build dir) is that \`uv tool run --with-requirements\` is the only runtime-injection path uv currently exposes, and it only consumes requirements.txt-format input.

Tracking [astral-sh/uv#19117](https://github.com/astral-sh/uv/issues/19117) — *\`uvx\` with a \`pylock.toml\` (E.g. \`--with-pylock\` or reuse \`--with-requirements\`)*. It's open, assigned to @zanieb (core uv maintainer), no milestone yet.

Verified on uv 0.10.11 that \`uv tool run --with-requirements pylock.toml\` recognizes the PEP 751 filename but silently drops every package in the lock — pylock-aware resolution today only fires in \`uv pip install/sync/compile\`, not in the runner path. The runner's \`--with-requirements\` help text reportedly already advertises pylock support (per a comment on #19117), so part of this is a doc/runtime mismatch and may land sooner than the full feature.

When #19117 ships, we should be able to:
- Drop \`uv export\` from the build pipeline (the byte-compare drama you just fixed in this PR goes away entirely).
- Ship \`pylock.toml\` (or \`uv.lock\` reprojected through pylock) inside \`builds/<hash>/\` instead of \`requirements.txt\` + wheel-as-sidecar.
- Keep portability since pylock is PEP-standard and consumable by any compliant installer.

No action needed here — just a marker so we revisit the packager design once uv plumbs this through.

zanieb:
What do you need `uvx --with-requirements` for? It's not currently supported because it's a weird concept to layer a lockfile onto a tool invocation.

Issue #1942, opened 2026-05-11
xorq uv build pays ~13s on every repeat invocation: wheel rebuilt into fresh tmpdir defeats uv tool env cache


paddymul:
## Repro

```bash
# Scaffold a minimal project and time two consecutive builds with no code change.
xorq init --path tmp-build-cost
cd tmp-build-cost
printf 'a,b\n1,2\n' > test.csv
cat > expr.py <<'PY'
import xorq.api as xo
con = xo.connect()
expr = xo.deferred_read_csv(con=con, path="test.csv", table_name="t")
PY
rm requirements.txt   # workaround for #1941

time xorq uv build expr.py -e expr --builds-dir builds
time xorq uv build expr.py -e expr --builds-dir builds   # identical, no edits
```

Observed on macOS (uv 0.11.6, Apple Silicon):

```
--- xorq uv build (first) ---
Building wheel...
Successfully built /var/folders/.../tmpwj30u2dr/xorq_template_..._.whl
builds/6a5585813a52
real    13.69s

--- xorq uv build (second, no code change) ---
Building wheel...
Successfully built /var/folders/.../tmpex6q_9cb/xorq_template_..._.whl
builds/6a5585813a52
real    12.95s
```

The build path (`builds/6a5585813a52`) is identical — proving the expression hashes the same — yet every invocation pays the full ~13s.

For contrast, `xorq uv run` against a stable build directory is fast on repeat:

```
--- xorq uv run (first) ---
Installed 76 packages in 192ms
real    7.17s

--- xorq uv run (second) ---
real    0.64s   # uv tool cache hit
```

## Root cause

`WheelPackager` (`python/xorq/ibis_yaml/packager.py`) instantiates a fresh `tempfile.TemporaryDirectory()` per instance, and `PackagedBuilder.from_script_path` constructs a new `WheelPackager` on every `xorq uv build` invocation. Result: the project wheel lands at a unique tmp path every time.

Downstream, `uv_tool_run(..., with_=self.wheel_path, ...)` shells out to:

```
uv tool run --isolated \
  --with /var/folders/.../tmpXXXX/<project>.whl \
  --with-requirements requirements.txt \
  xorq build ...
```

uv keys its tool-env cache on `(--with paths, --with-requirements content, ...)`. A different `--with` path on every invocation forces a cache miss → uv installs the full dependency set into a fresh ephemeral env every build.

[...]

PR #1946, opened 2026-05-11
fix(packager): default to uv --link-mode hardlink on macOS (#1942)

paddymul:
## Summary

Per [standup discussion with Dan](https://github.com/xorq-labs/xorq/issues/1942): partial mitigation for #1942's repeat-invocation `xorq uv build` cost on macOS.

- Adds `options.uv.use_hardlink` (settable via `XORQ_UV_USE_HARDLINK`), defaulting to `True` on macOS and `False` elsewhere.
- When `True`, `WheelPackager._build_wheel` and `uv_tool_run` pass `--link-mode hardlink` to the uv subprocess. This avoids the per-invocation syspolicyd rescan that uv's default `--link-mode=clone` triggers on macOS (astral-sh/uv#18577).
- Does **not** address the root cause from #1942 (fresh tmp wheel path on every build defeats uv's tool-env cache) — that's still the larger fix on the umbrella issue.

A/B from #1942 (with this default on macOS, second-build steady state, 3-trial average):

| Step | `clone` (current default) | `hardlink` (this PR's default on macOS) |
|---|---|---|
| `xorq uv build` (2nd) | ~14s | ~7s |

## Why is it still ~7s? Deep-dive

Decomposing `xorq uv build` over 3 trials (averaged, macOS, this branch):

```
                                         hardlink     clone     delta
uv build --wheel                            0.32s     0.21s    -0.11s
uv export (requirements.txt)                0.03s     0.03s    -0.00s
uv tool run xorq build                      7.03s    14.35s    +7.32s
                                         ─────────────────────────────
TOTAL                                       7.38s    14.59s    +7.21s
```

Wheel build + uv export are negligible (~0.25s). All cost is in `uv tool run`. Decomposing that further:

```
                                         hardlink     clone
uv tool run + python -c 'pass'              2.76s     0.49s    ← env install only
uv tool run + import xorq.api               5.43s     7.31s    ← + ~2.7s of imports
uv tool run + xorq build (full)             7.03s    14.35s    ← + ~1.6s of build work
```

**What's eating the 7s on hardlink:**

| Component | Time | Why |
|---|---|---|
| uv installs 76 packages into fresh ephemeral env | **2.76s** | `--isolated` + new wheel path every build defeats tool-env cache |
| `import xorq.api` (ibis, datafusion, etc.) | **2.67s** | Python import cost; pays per invocation |
| `xorq build` work (parse, serialize, write) | **1.57s** | Actual build logic |

**Asymmetry on `python -c 'pass'`:** hardlink is *slower* (2.76s) than clone (0.49s) for env-creation-only workloads — hardlink pays more kernel overhead per file linked, and uv's cache layout for this minimal case favors clone. But once you start *reading* .so files (the `import xorq.api` and the actual build), syspolicyd rescans on clone destroy that lead. Hardlink wins on workloads that read many native extensions, which is our case.

**The remaining ~7s is structural, not link-mode:**

- **2.76s is recoverable** if we kill the per-invocation env install. Issue #1942's suggested fix #1 (content-address the wheel → stable `--with` path → uv tool-env cache hit) would drop this to ~0.5s, putting total around **~4.5s**.
- **2.67s `import xorq.api`** is the lower bound for any uv-tool-run-style flow. To go below that you'd need either AOT-compiled imports or a long-running build daemon.
- **1.57s** of actual build work is what's left after fixing both.

So this PR captures the easy ~50% (the syspolicyd workaround); the next ~40% requires the content-addressed wheel work tracked on #1942.

## Design notes (from chat with Dan)

> settable EnvConfig that propagates to xorq.config.options; defaults to True on osx, False otherwise … will propagate to xorq.ibis_yaml.packager.{PackagedBuilder,PackagedRunner}

Implementation matches that spec: `XORQ_UV_USE_HARDLINK` → `options.uv.use_hardlink` → packager reads it at uv-subprocess-invocation time and passes `--link-mode hardlink` (the uv CLI flag, not the `UV_LINK_MODE` env var — per Dan's preference for command-line flag over env-var).

## Test plan

- [x] `_link_mode_args()` returns `("--link-mode", "hardlink")` when option True, `()` when False
- [x] `uv_tool_run` passes `--link-mode hardlink` to the subprocess when option True; omits flag when False
- [x] `options.uv.use_hardlink` defaults `True` on darwin, `False` on linux
- [x] `XORQ_UV_USE_HARDLINK=False` overrides the darwin default
- [x] First commit (failing tests at baseline) was seen failing on CI (linux core job) before the fix landed
- [x] CI green on the fix commit


paddymul:
> ## roborev: Fail
> Review findings
> ## Review Findings
> 
> **Severity**: Low **Location**: `python/xorq/ibis_yaml/tests/test_packager.py:361–373`, `436–447` **Problem**: All wheel-build and `uv_tool_run` tests stub out `subprocess.run`, so they verify that `--link-mode hardlink` is injected into the argument list but cannot catch a `uv` version where `uv build` does not accept `--link-mode`. On macOS the default is `use_hardlink=True`, meaning every `WheelPackager._build_wheel()` call would pass the flag; a version incompatibility would only surface at runtime, not in CI. **Fix**: Document the minimum supported `uv` version that accepts `--link-mode` in `uv build` (or add a smoke integration test that runs a no-op `uv build --help` / `uv tool run --help` and checks for the flag).
> ## Summary
> 
> Adds `--link-mode hardlink` as an opt-in (macOS-default) setting for `uv build` and `uv tool run` invocations, plumbed through a new `options.uv.use_hardlink` config attribute and backed by a thorough suite of unit tests that mock `subprocess.run`.
> 
> _Review type: default | Agent: claude-code | Job: 88_

I'm not worried about this.  it looks like link_mode has been around since 2024
https://github.com/search?q=repo%3Aastral-sh%2Fuv+link-mode&type=commits&p=3

dlovell (approved):

PR #1949, opened 2026-05-12
refactor(config): add parse_bool_env for consistent env var parsing

dlovell:
## Summary
- Adds a `parse_bool_env` helper to `env_utils.py` that accepts `true/false/1/0/yes/no/on/off` (case-insensitive, whitespace-trimmed) and raises `ValueError` on unrecognised input
- Replaces inconsistent `ast.literal_eval` usage in `config.py` (`XORQ_DEBUG`, `XORQ_TUI_REVISIONS_OPEN`, `XORQ_TUI_GIT_LOG_OPEN`) — these previously crashed on shell-style values like `true`/`false`
- Fixes a bug in `snowflake_utils.py` where `bool("False")` evaluated to `True`

## Test plan
- [x] 17 parametrized tests for `parse_bool_env` covering truthy, falsy, case variants, whitespace, default, and garbage rejection
- [x] Verify `xorq.config.options` loads correctly with default `.env.xorq.template` values
- [x] Verify `SNOWFLAKE_CREATE_OBJECT_UDFS=False` now correctly evaluates to `False`

### DIFF
--- a/python/xorq/common/utils/env_utils.py
+++ b/python/xorq/common/utils/env_utils.py
@@ -148,6 +148,22 @@ class EnvConfigable:
         )
 
 
+_TRUTHY = frozenset({"true", "1", "yes", "on"})
+_FALSY = frozenset({"false", "0", "no", "off"})
+
+
+def parse_bool_env(value: str) -> bool:
+    normalised = value.strip().lower()
+    if normalised in _TRUTHY:
+        return True
+    if normalised in _FALSY:
+        return False
+    raise ValueError(
+        f"Cannot interpret {value!r} as bool; "
+        f"expected one of {sorted(_TRUTHY | _FALSY)}"
+    )
+
+
 @toolz.curry
 def maybe_substitute_env_var(obj, ctx=os.environ):
     if isinstance(obj, str) and (match := compiled_env_var_substitution_re.match(obj)):
--- a/python/xorq/config.py
+++ b/python/xorq/config.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
-import ast
 import pathlib
 from typing import Any, Optional
 
 from xorq.common.utils.env_utils import (
     EnvConfigable,
     env_templates_dir,
+    parse_bool_env,
 )
 from xorq.vendor import ibis
 from xorq.vendor.ibis.backends import BaseBackend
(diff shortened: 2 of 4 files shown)


python/xorq/config.py
```diff
@@ -177,11 +177,11 @@ class TUI(Config):
 
     left_ratio: int = max(int(env_config.XORQ_TUI_LEFT_RATIO or 2), 1)
     right_ratio: int = max(int(env_config.XORQ_TUI_RIGHT_RATIO or 3), 1)
-    revisions_open: bool = bool(
-        ast.literal_eval(env_config.XORQ_TUI_REVISIONS_OPEN or "False")
+    revisions_open: bool = bool(env_config.XORQ_TUI_REVISIONS_OPEN) and parse_bool_env(
```
paddymul, line 180: Why this pattern here?  are you verifying that XORQ_TUI_REVISIONS_OPEN is truthy (not None) so you can call parse_bool_env on it.

It looks like you repeat this pattern in every place you use it.  a `safe_parse_bool_env` that encapsulates the pattern and reduce repetition would be nice.

python/xorq/common/utils/env_utils.py
```diff
@@ -148,6 +148,22 @@ def subclass_from_env_file(cls, env_file, prefix="", name_mapper=None):
         )
 
 
+_TRUTHY = frozenset({"true", "1", "yes", "on"})
+_FALSY = frozenset({"false", "0", "no", "off"})
+
+
+def parse_bool_env(value: str) -> bool:
```
paddymul, line 155: I looked at EnvConfiggable, maybe not closely enough, but it's not obvious from looking there that every value returned is a string, there aren't types.  So how do we know that this function will be called with a string?

I would expect `def __getitem__(self, key):` to have a return type of `str`
https://github.com/xorq-labs/xorq/blob/main/python/xorq/common/utils/env_utils.py#L64

probably out of scope for this PR, it looks like EnvConfiggable will always have `str` values. It struck me looking at the code.  Also if you passed in an int or a bool, this would fail at line 156.

python/xorq/config.py
dlovell: if people use our env templates, then every env var we could use is at least defined, but possibly empty
but whether an empty value should be false depends on the variable / particular use

python/xorq/common/utils/env_utils.py
dlovell: EnvConfigable/EnvConfig doesn't do type casting, its just there to specify what possible values could be used and to get them rather than having os.environ scattered throughout the codebase

paddymul (approved):

PR #1951, opened 2026-05-12
Replace dask-based tokenization with xorq-dasher

### LINKED ISSUE #1972: Falsy-or path fallback silently swallows empty-string paths
## Summary

Two sites use `read_kwargs.get("read_path") or read_kwargs["hash_path"]` to select a path, which treats any falsy `read_path` (including `""`) as absent and falls through to `hash_path`:

- `python/xorq/caching/strategy.py:46`
- `python/xorq/common/utils/dasher/_opaque.py:159`

## Impact

If `read_path` is set to an empty string (or another falsy value), the code silently uses `hash_path` instead of raising an error. This could produce incorrect cache keys — a particularly dangerous failure mode since it's silent and affects correctness.

## Suggested fix

Use an explicit `None` check instead of falsy `or`:

```python
# strategy.py:46
path = read_kwargs.get("read_path")
if path is None:
    path = read_kwargs["hash_path"]

# _opaque.py:159
anchor = read_kwargs.get("read_path")
if anchor is None:
    anchor = read_kwargs.get("hash_path")
```

This preserves the intended fallback while correctly raising `KeyError` / propagating empty strings rather than swallowing them.

Barre:
## Migrate tokenization from dask to xorq-dasher

Replaces the `xorq.common.utils.dask_normalize/` package with a new
`xorq.common.utils.dasher` package backed by `xorq-dasher`. Drops dask
as a runtime dep for cache-key computation.

### Why

`dask.base.tokenize` mutates global registry state and needs monkey-patching
to swap behavior per call (which is what `SnapshotStrategy` was doing).
`dasher.Hasher` supports per-call rule overrides and a contextvars channel
for transitive tokenize calls — no globals to mutate, no patching.

### Changes

- New `dasher/` package: `__init__.py` (public API + HASHER), plus
`_gap_rules` / `_paths` / `_relations` / `_opaque` submodules.
- `SnapshotStrategy` rewritten on top of `snapshot_hasher` + a `_current_hasher`
contextvar; no more monkey-patching `dask.base.normalize_token`.
- Per-backend `DatabaseTable` normalizers (DataFusion, DuckDB) extract file
paths and stat them so `ModificationTimeStrategy` invalidates on file edits.
- Catalog-extract tempdir paths stripped to their build-hashed suffix so
`load_expr_from_zip` is reload-stable (ADR-0007).
- Opaque-leaf placeholders use content-addressed names instead of `id()`
fixes catalog-reload tokenization (ADR-0010).
- `_normalize_computed_kwargs_expr` is data-free — schema-only for
`InMemoryTable`, path-stripped for `Read`, cache-class for `CachedNode`.
Fixes the `FittedPipeline.predict`-folds-training-data-into-hash issue.

### Correctness fixes (review round)

- Fix falsy-or path fallback in `snapshot_normalize_read` and
  `_xorq_opaque_to_placeholder` — use explicit `is not None` so
  empty-string `read_path` is not silently swallowed (closes #1972)
- Restrict ReadParquet path extraction to `expressions[0]` to avoid
  capturing string-valued kwargs as file paths
- Add `_MISSING` sentinel for FlightUDXF `exchange_f` to distinguish
  missing attribute from explicit `None`
- Use atomic pointer swap in `FQNDispatch.register` for thread safety
- Make catalog-path regex greedy to strip through last catalog segment


codecov.yml
dlovell, line 1: please remove this

.gitignore
dlovell, line 1: please remove this

pyproject.toml
mesejo, line 33: I think we can drop cityhash, it's only there to guarantee that the algorithm selection by dask was deterministic

Barre: What about https://github.com/xorq-labs/xorq/blob/24ffdf2677cc124e98a062b008814ff7c6928012/python/xorq/vendor/ibis/backends/sqlite/udf.py#L249-L251?

python/xorq/caching/strategy.py
```diff
@@ -43,8 +43,7 @@ def snapshot_normalize_read(read):
     # Materialized build-bundle reads carry a content-hash-named read_path that is
     # stable across environments. Their hash_path is an absolute tmpdir path that
     # changes every run, so prefer read_path when available.
-    rp = read_kwargs.get("read_path")
-    path = rp if rp is not None else read_kwargs["hash_path"]
+    path = read_kwargs.get("read_path") or read_kwargs["hash_path"]
```
dlovell: this was intentional so `""` would be treated different than None

dlovell (approved):

PR #1968, opened 2026-05-19
ci(codecov): fix fragmented coverage reporting and scope flags to paths

mesejo:
**Coverage flags and CI reporting**

Each CI workflow uploads under a named Codecov flag (`core`, `library`, `examples`, `slow`, `databricks`, `snowflake`, `gcs`, `s3`). Flags make Codecov treat each upload as a separate named group rather than merging everything into one anonymous blob, so PR comments and the dashboard show per-backend coverage deltas instead of a single repo-wide number.

Without explicit flag definitions, every flag is scored against the full repo tree. This makes `examples` look artificially low (it never exercises `python/xorq/` internals) and hides which backend actually caused a coverage drop on a given PR.

**Changes**

- `ci-test.yml`: add `flags: core` and `fail_ci_if_error: false` to the Codecov upload step (both were missing)
- `codecov.yml`: add per-flag `paths` filters so each flag's coverage percentage is scoped to the relevant source tree; add `carryforward: true` so skipped workflows (e.g. snowflake not running on a PR) carry the last known value instead of dropping to 0%; add coverage status thresholds and PR comment config
- `catalog-*` flags use dynamic names and are covered by `default_rules`

**`[tool.coverage.run]` — data collection**

| Setting                                       | Effect                                                                                                            |
| --------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
| `source = ["python/xorq"]`                    | Measure only this package; ignore third-party and stdlib                                                          |
| `omit`                                        | Exclude test files and vendored code from stats                                                                   |
| `parallel = true`                             | Each process writes its own `.coverage.N` file; required with xdist and multiprocessing                           |
| `concurrency = ["multiprocessing", "thread"]` | Tells the tracer which concurrency primitives are in use so it follows execution across process/thread boundaries |
| `sigterm = true`                              | Flush data on SIGTERM; prevents loss when xdist kills workers mid-run                                             |

**`[tool.coverage.report]` — display**

- `show_missing = true`: print line numbers of uncovered lines
- `skip_covered = false`: show 100%-covered files too, not just files with gaps

**Why `parallel` and `concurrency` must be set together**

`parallel = true` splits writes correctly across processes. `concurrency` is what makes the tracer actually follow execution into threads and subprocesses. Without it, anything running in a worker is silently missed, producing falsely low numbers.

**Coverage reach graph**

The reach graph shows what percentage of the codebase is reachable by any test at all — distinct from line coverage, which measures how many reachable lines were actually executed. A repo with high line coverage but low reach has large swaths of code that no test suite even loads. Codecov plots reach over time so you can see whether new code is being exercised at all, not just whether existing tests still pass.

https://docs.codecov.com/v4.6/docs/graphs#section-coverage-reach


.github/workflows/ci-test.yml
```diff
@@ -160,3 +160,6 @@ jobs:
         uses: codecov/codecov-action@v5.5.4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          flags: core
+          fail_ci_if_error: false
+          verbose: true
```
dlovell: is `verbose: true` intentional? if so, what is its purpose?

codecov.yml
dlovell, line 1: claude says
```
  2. codecov.yml — no explicit flag definitions. The workflow files set flags: core, flags: library, etc., but codecov.yml only configures flag_management.default_rules. This works
  (Codecov auto-creates flags from uploads), but adding explicit flag definitions with paths filters would give more accurate per-flag coverage and prevent a stale flag if a workflow is
  renamed. Not blocking, but worth considering.
```
and suggests this
```
  flags:
    core:
      paths:
        - python/xorq/
      carryforward: true
    library:
      paths:
        - python/xorq/
      carryforward: true
    examples:
      paths:
        - examples/
      carryforward: true
    slow:
      paths:
        - python/xorq/
      carryforward: true

  The paths filter tells Codecov which source files each flag's coverage applies to, so the per-flag coverage % is scoped to the right code. Without this, every flag reports coverage
  against the entire repo — meaning the examples flag shows low coverage because it never touches python/xorq/ internals, even though that's expected. It also lets Codecov detect stale
  flags if a workflow stops uploading.

  That said, this is a polish item — the current setup works, it just reports noisier per-flag numbers.
```

dlovell (approved):

PR #1963, opened 2026-05-18
chore(init): bump template pins to pin-uv branches

paddymul:
## Summary
- Bump `cached_fetcher`, `sklearn`, and `penguins` template pins in `init_templates.py` to the latest template commits.
- The new template commits pin `uv>=0.9.18`, export `requirements.txt` with the flags xorq's packager compares against (`--locked --no-dev --no-emit-project --no-header --no-annotate`), and bump xorq to v0.3.24 in the templates' `uv.lock`.

Without this bump, fresh `xorq init` projects fail the packager's byte-exact `requirements.txt` check on first `xorq uv build` (#1941).

Companion PRs:
- xorq-labs/xorq-template-cached-fetcher#7
- xorq-labs/xorq-template-sklearn#12
- xorq-labs/xorq-template-penguins#13

## Test plan
- [x] `uv run pytest python/xorq/tests/test_cli.py::test_init_command_sklearn` — passes (downloads each new SHA from GitHub)
- [ ] Manual `xorq init --template <each>` smoke
- [ ] Slow test: `test_init_uv_build_uv_run` end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1869, opened 2026-04-24
refactor(xorq_datafusion): flatten DataFusionBackend inheritance

mesejo:
Merge two-level xorq_datafusion → datafusion.Backend hierarchy into a
single Backend class. Deletes the datafusion/ subdirectory entirely.
    
Fixes bundled with the merge:
  - register(): match/case dispatch, single deregister_table, one return;
    add unbound-expression guard; ir.Table/ir.Expr arms replace verbose
    ibis.expr.types.* references; drop redundant hasattr guards
 - _compile_pyarrow_udaf: param_name loop var shadowed outer name;
    materialize parameters to list, guard empty case
 - _to_pyarrow_batches / to_parquet: remove redundant _import_pyarrow()
    calls — pyarrow unconditionally imported at module level
 - register_table_provider: narrow source to ir.Table; guard None name
 - translate_sort: move to Backend._translate_sort as `@staticmethod`
 - Promote pandas/pyarrow.dataset/pyarrow.parquet to module-level imports
    
BREAKING CHANGE: truncate_table removed — DataFusion does not implement
DELETE DML; the method was a silent no-op.
    
BREAKING CHANGE: read_postgres removed — _from_url kwargs silently ignored database param.update(); no callers found in codebase.

## Follow-up

`register()` has 9 dispatch arms. This PR adds tests for three paths:
`pd.DataFrame` → `pa.Table`, `ir.Table` via `IbisTableProvider`, and
`ir.Expr` (memtable → `execute()` path).

The remaining six arms — `pa.Table`, `pa.RecordBatch`,
`pa.RecordBatchReader`, `ds.Dataset`, DataFusion `Table`, DataFusion
`DataFrame`, and the `str`/`Path` (parquet/csv) early-return — are not
covered by new tests. They are exercised indirectly by the broader test
suite but lack dedicated unit tests. A follow-up PR will add targeted
tests for each arm, including the `ordering`/`sort_order` kwarg path
through `_translate_sort`.

PR #1975, opened 2026-05-20
test(packager): regression test for xorq-datafusion 0.2.7 execute_stream crash

paddymul:
## Summary
- Adds a failing test that pins `xorq-template-sklearn` to SHA `d94dd60f` (its pin-uv branch — the same SHA PR #1963 wants to bump `init_templates.py` to).
- The packaged build of that template crashes inside `frame.execute_stream()` with **`Arrow error: C Data interface error: Invalid: Mismatching number of fields and child arrays`** when the cache layer writes `fitted_pipeline.predict(test)`.
- Surfaces as exit 139 (SIGSEGV) on Linux CI and as a Python `ValueError` on macOS.

## Why a new test pinned to a SHA

The regression only manifests when:
1. xorq-datafusion is **0.2.7** (which xorq 0.3.25 pins — see #1869, #1952), AND
2. The expression goes through the packager's YAML round-trip + `uv tool run --isolated` execution path.

Running the same sklearn-pipeline expression in-process on this branch's venv succeeds — only the isolated/packaged path crashes. The existing `test_wheel_runner[sklearn]` still passes because `init_templates.py` on `main` still points at the old SHA (whose `uv.lock` pins `xorq==0.3.13`). Pinning this new test to the post-bump SHA captures the regression independently of #1963 landing.

When the upstream fix in xorq-datafusion (or the `xorq_datafusion` backend wrapper from #1869) lands, this test will pass and can be folded back into the parametrized `test_wheel_runner`.

## Test plan
- [x] Test fails locally on this branch (worktree synced to main + `xorq-datafusion==0.2.7`): `UvToolRunError ... exit status 139`.
- [ ] CI shows the same failure on `ci-test-slow` (slow level 2).
- [ ] Verify on macOS runners too, where the inner error surfaces as `ValueError` not SIGSEGV.

### DIFF
--- a/python/xorq/ibis_yaml/tests/fixtures/execute_stream_regression/README.md
+++ b/python/xorq/ibis_yaml/tests/fixtures/execute_stream_regression/README.md
@@ -1,0 +1,14 @@
+# execute_stream_regression fixture
+
+Snapshot of `xorq-labs/xorq-template-sklearn` at commit
+`d94dd60fe2328af70ba5652effb442e022719186`. Vendored so the regression test
+does not depend on cloning a separate template repo at test time.
+
+Drives `test_wheel_runner_xorq_datafusion_execute_stream_regression` in
+`../test_packager.py`. The packaged build of this project hits an Arrow C Data
+Interface mismatch in `frame.execute_stream()` on xorq-datafusion 0.2.7 — see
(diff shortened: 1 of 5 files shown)


mesejo:
This PR is stale

PR #1977, opened 2026-05-21
feat(relations): replace SafeTee with batchcorder StreamCache

### LINKED ISSUE #983: `asof_join` with `tolerance` and an `into_backend` gives empty result
```# https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.asof_join
from datetime import datetime, timedelta

import xorq as xo


sensors_parquet = "sensors.parquet"
events_parquet = "events.parquet"
sensors = xo.memtable(
    {
        "site": ["a", "b", "a", "b", "a"],
        "humidity": [0.3, 0.4, 0.5, 0.6, 0.7],
        "event_time": [
            datetime(2024, 11, 16, 12, 0, 15, 500000),
            datetime(2024, 11, 16, 12, 0, 15, 700000),
            datetime(2024, 11, 17, 18, 12, 14, 950000),
            datetime(2024, 11, 17, 18, 12, 15, 120000),
            datetime(2024, 11, 18, 18, 12, 15, 100000),
        ],
    }
)
events = xo.memtable(
    {
        "site": ["a", "b", "a"],
        "event_type": [
            "cloud coverage",
            "rain start",
            "rain stop",
        ],
        "event_time": [
            datetime(2024, 11, 16, 12, 0, 15, 400000),
            datetime(2024, 11, 17, 18, 12, 15, 100000),
            datetime(2024, 11, 18, 18, 12, 15, 100000),
        ],
    }
)
tolerance = timedelta(seconds=1)
sensors.to_parquet(sensors_parquet)
events.to_parquet(events_parquet)


con = xo.duckdb.connect()
expr0 = (
    sensors.into_backend(con)
    .asof_join(
        events.into_backend(con),
        on="event_time",
        predicates="site",
        # tolerance=tolerance,
    )
    .drop(
        "event_time_right"
    )
    .order_by("event_time")
)
expr1 = (
    con.read_parquet("sensors.parquet")
    .asof_join(
        con.read_parquet("events.parquet"),
        on="event_time",
        predicates="site",
        tolerance=tolerance,
    )
    .drop(
        "event_time_right"
    )
    .order_by("event_time")
)
expr2 = (
    sensors.into_backend(con)
    .asof_join(
        events.into_backend(con),
        on="event_time",
        predicates="site",
        tolerance=tolerance,
    )
    .drop(
        "event_time_right"
    )
    .order_by("event_time")
)


# this works but gives a result without tolerance
df0 = expr0.execute()
# this works but requires read from parquet
df1 = expr1.execute()
# this is empty
df2 = expr2.execute()
assert df2.empty
```

mesejo:
Fix #983: asof_join with tolerance and self-join returned empty results
because SafeTee exhausted a one-shot Arrow iterator on re-scan.

StreamCache (batchcorder 0.1.3) wraps any Arrow stream in a replayable
Rust-backed cache. DuckDB/DataFusion acquire independent StreamCacheReader
handles per scan — thread-safe concurrent reads. Only memory mode is used;
batches live in a Rust Arc<Vec<RecordBatch>>. Disk spill is not configured
and not enabled.

## Bounded memory via `max_readers`

Each StreamCache is constructed with `max_readers` = a per-RemoteTable scan
count, so batches are evicted once all readers advance past them (instead of
retaining the whole stream). batchcorder enforces `max_readers` as a hard
cap: an under-count crashes (`ValueError: Maximum number of readers reached`
when the backend creates reader N+1), an over-count silently disables
eviction. Omitting it (passing `None`) leaves the cache unbounded — always
safe, no eviction.

The scan count is **not** visible in the expression graph: a single graph
reference can compile to several physical scans (e.g. a self-join, the
canonical multi-scan shape). `count_remote_table_readers` therefore rewrites
each RemoteTable to a placeholder under a fresh unique name, compiles the
expression to a sqlglot AST, and counts `Table` nodes bearing that name —
robust across compiler lowering and immune to inflation from column refs,
aliases, or string literals. Counts are floored at 1 (a bare RemoteTable is
still scanned once; `max_readers=0` would forbid the reader that is in fact
created); on SQL-compile failure (e.g. a non-SQL backend) the mapping is
empty and the cache is left unbounded.

The count is a best-effort **lower bound**, not an exact scan count: it sees
only the scans the compiled SQL spells out, not re-scans the backend's
optimiser introduces below the SQL layer. Known gap: a DuckDB `PARTITION BY`-
only aggregate window lowers to a `GROUP BY` self-join that scans its input
twice while the AST counts one — an under-count that currently raises. See
the ADR ("Counting accuracy") for the tradeoff.

DuckDB scans the cache directly, so the count drives eviction and an
under-count crashes. DataFusion reads the outer cache once (`from_stream`)
and re-wraps it into an inner cache it scans per reference, so an over-count
on the outer cache there merely disables eviction (never crashes).

## Core changes

- **remote_table_exec.py** (new): `register_and_transform_remote_tables`,
  `count_remote_table_readers`, `RemoteTableScope`, `bind_scope_to_reader`,
  and `prepare_create_table_from_expr` — moved out of relations.py. The
  replacer creates one StreamCache per RemoteTable node with `max_readers`
  from `count_remote_table_readers`, and returns `(expr, scope)` where the
  `RemoteTableScope` owns every materialized resource (upstream readers,
  StreamCaches, placeholder tables, tee drains) behind one idempotent
  `close()` that tears down in dependency order.
- **relations.py**: SafeTee, the old graph-walk
  `register_and_transform_remote_tables`, and `prepare_create_table_from_expr`
  removed (relocated to remote_table_exec.py). RemoteTable caches its
  reader's physical schema for the StreamCache.
- **api.py**: `_transform_expr` returns `(expr, scope)`. Eager call sites
  (`_pandas_execute`, `get_plans`, caching/storage) use the new
  `remote_table_scope` context manager, which closes the scope on exit and
  raises tee/WAP drain failures only on the success path. The streaming
  `to_pyarrow_batches` path uses `bind_scope_to_reader`, which defers
  `scope.close()` to the result reader's generator `finally` (with a
  `weakref.finalize` backstop for readers abandoned before first read) —
  replacing the removed `rbr_wrapper`/`_run_cleanup` helpers.

## Backend `read_record_batches` plumbing

- **logical schema**: `read_record_batches` takes the logical schema and
  projects/casts the physical reader down to it, dropping extra physical
  columns. `project_and_cast_reader` short-circuits when the schema already
  matches. Redundant schema args were removed where the reader carries it.
- **xorq_datafusion**: registers `StreamCache` via `StreamCache.cast` (a
  replayable casting view over the same buffer) so DataFusion reads purely
  from Rust — prevents the GIL/Mutex deadlock fixed upstream in
  xorq-datafusion ef86e2b (Python::attach moved to spawn_blocking). Uses
  `CastingStreamCache` so fan-out scans still evict; `read_record_batches`
  collapsed into a single match with a shared `_casting_reader` helper and
  rollback hoisted across all source types. Bumps xorq-datafusion 0.2.7 →
  0.2.9.
- **postgres**: adds a `to_pyarrow_batches` override on the xorq Backend;
  uses PgADBC for the ADBC primary path + psycopg fallback for session-local
  temp tables, logging the swallowed ADBC connection failure before falling
  back. Plugs an ADBC leak and restores remote-table kwargs. (The vendored
  postgres backend is not touched.)
- **gizmosql**: materializes empty `StreamCache` streams and emits a
  zero-row batch so an empty stream still ingests over Flight SQL.
- **sqlite / pandas / pyiceberg**: handle `StreamCache` input in
  `read_record_batches` (shared `coerce_to_arrow_table` helper).
- **snowflake**: pass keypair as a PEM string for adbc>=1.11 compat; drop
  snowflake from `_ADBC_BACKENDS` `mode="replace"`; restore `create_table`
  permissiveness for non-Expr objects.
- **flight**: use `to_reader` instead of the pyarrow-14-only `from_stream`.

## Tests

- expr/tests/test_relations.py: unit tests for `count_remote_table_readers`
  (bare=1, single scan=1, many fields=1, self-join=2, asof double-scan=[1,2],
  unbindable → empty mapping)
- expr/tests/test_transform_scope.py + test_stream_cache_contract.py:
  RemoteTableScope lifecycle and StreamCache reader contract
- backends/duckdb/tests/test_into_backend.py and
  backends/xorq_datafusion/tests/test_into_backend.py: into_backend fan-out
  edge cases — bare, filter, self-join, two scalar subqueries, 3-way union,
  two distinct tables, empty-table fan-out, nested chain, asof double-scan
- backends/duckdb/tests/test_asof_join.py: assert single-scan tolerance
  lowering after #2086; pin the max_readers under-count crash for the
  partition-by window
- backends/xorq_datafusion/tests/test_stream_cache_deadlock.py: subprocess
  regression test for the two-scan GIL/Mutex deadlock; plus reentrancy and
  register coverage
- backends/{gizmosql,postgres,sqlite,pandas,snowflake} tests: backend
  read_record_batches / into_backend / to_pyarrow_batches coverage

Decision recorded in docs/adr/0013-batchcorder-stream-cache-for-remote-
table-fan-out.md. Covers: root cause, SafeTee failure modes, StreamCache
resource lifecycle (RemoteTableScope; explicit close at all executing call
sites), memory-only tradeoffs, the max_readers counting-accuracy gap, and
alternatives considered (tee, pre-ingest to pa.Table).

Closes #983

### DIFF
--- a/docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
+++ b/docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
@@ -3,7 +3,7 @@
 - **Status:** Accepted
 - **Date:** 2026-05-22
 - **Deciders:** Daniel
-- **Context area:** `python/xorq/expr/relations.py`, `python/xorq/expr/api.py`, `python/xorq/backends/pandas/__init__.py`
+- **Context area:** `python/xorq/expr/remote_table_exec.py`, `python/xorq/expr/api.py`, `python/xorq/expr/relations.py`
 
 ## Context
 
@@ -27,7 +27,7 @@ expr.execute()  # empty — should have 5 rows
 
 ### Root cause: multi-scan of a one-shot Arrow iterator
 
-`register_and_transform_remote_tables` (`relations.py`) walks an expression graph,
+`register_and_transform_remote_tables` (`remote_table_exec.py`) walks an expression graph,
 finds every `RemoteTable` node, materialises the remote expression as an Arrow
 `RecordBatchReader`, and registers it with the local DuckDB backend so the expression
 can execute locally.
@@ -94,22 +94,33 @@ batches lazily from the upstream reader and stores them internally. Any number o
 independent `StreamCacheReader` handles can replay the full stream from position zero
 without knowing the total reader count in advance.
 
-In `replacer`, each `RemoteTable` now gets a single `StreamCache`:
+In `replacer` (inside `register_and_transform_remote_tables` in
+`remote_table_exec.py`), each `RemoteTable` gets a single `StreamCache` whose
+resources are tracked by a `RemoteTableScope`:
 
 ```python
-cache = StreamCache(
-    pa.RecordBatchReader.from_batches(schema, remote_expr.to_pyarrow_batches())
+reader = scope.adopt_reader(remote_expr.to_pyarrow_batches())
+cache = scope.adopt_cache(
+    StreamCache(reader, max_readers=reader_counts.get(node))
 )
-caches.append(cache)
-table_name = gen_name()
-result = node.source.read_record_batches(cache, table_name=table_name)
-created[table_name] = node.source
+table_name = scope.adopt_table(node.source, gen_name())
+result = node.source.read_record_batches(cache, table_name=table_name, **read_kwargs)
 ```
 
 DuckDB acquires a fresh `StreamCacheReader` via `__arrow_c_stream__` each time it
 scans the registered table. Every reader starts at batch 0 and advances
 independently. The upstream remote expression is called exactly once — ingestion is
 lazy and shared across all readers via an internal `Arc<Mutex<DatasetInner>>`.
+
+#### Bounded eviction via `max_readers`
+
+`count_remote_table_readers` compiles the expression to a sqlglot AST with sentinel
+table names and counts how many `Table` nodes bear each sentinel — giving the exact
+physical scan count whatever lowering the backend compiler applies. The count becomes
+`StreamCache`'s `max_readers`, allowing it to evict batches once all readers advance
+past them. This is a memory optimisation, not a correctness requirement: if no AST
+can be produced (non-SQL backend), `max_readers` is omitted and the cache retains all
+batches.
 
 ### Storage modes
 
@@ -125,29 +136,29 @@ streams that exceed available RAM. Because `close()` is already called explicitl
 all executing call sites, enabling disk mode later is a configuration change only —
 no new cleanup logic is required.
 
-### Resource lifecycle
+### Resource lifecycle — `RemoteTableScope`
 
-`register_and_transform_remote_tables` returns `(expr, created, caches)` where
-`caches: list[StreamCache]` is owned by the call site.
+`register_and_transform_remote_tables` returns `(expr, scope)` where
+`scope: RemoteTableScope` owns every resource the transform materialised:
+upstream `RecordBatchReader`s, `StreamCache`s, and placeholder tables.
+`scope.close()` tears down in dependency order (tables → caches → readers),
+LIFO within each category, and is idempotent.
 
-**Executing call sites close caches explicitly:**
+Call sites use one of two patterns depending on whether execution is eager or
+streaming:
 
-- **`_pandas_execute`**: `finally` block immediately after `con.execute()` returns.
-  Pandas execution is synchronous and fully materialises before returning.
-- **`get_plans`**: `finally` block after the `EXPLAIN` SQL completes. The upstream
-  stream is never read during `EXPLAIN`; closing here releases it without waiting
-  for GC.
-- **`to_pyarrow_batches`**: `cache.close()` inside the `clean_up` callback passed
-  to `rbr_wrapper`. The callback fires when the returned `RecordBatchReader` generator
-  exits — either on full exhaustion or when the generator is closed/GCed.
-- **`prepare_create_table_from_expr`**: `finally` block after the transformed
-  expression is returned. Caches are closed before the caller materialises the table.
+- **Eager (fully materialised before returning):**
+  `remote_table_scope` is a context manager in `api.py` that yields the
+  transformed expression and closes the scope on exit.  Used by
+  `_pandas_execute`, `get_plans`, and any backend whose
+  `read_record_batches` eagerly ingests the stream (Snowflake, Postgres via
+  ADBC through `prepare_create_table_from_expr`).
 
-### Effect on `created`
-
-`created` maps `table_name → backend`. For a self-join that previously generated
-three `SafeTee` copies, `len(created)` drops from 3 to 1 — one entry per unique
-`RemoteTable` node.
+- **Streaming (`to_pyarrow_batches`):**
+  `bind_scope_to_reader` wraps the result `RecordBatchReader` in a generator
+  whose `finally` block calls `scope.close()`.  A `weakref.finalize`
+  backstop covers readers abandoned before the first read (a never-started
+  generator never runs its `finally`).
 
 ## Alternatives considered
 
@@ -182,13 +193,15 @@ ingestion and an optional disk spill path.
 
 ### Positive
 
-- **No pre-counting.** `replacer` creates one `StreamCache` per `RemoteTable` node
-  without knowing how many times DuckDB will scan it.
+- **No pre-counting for correctness.** `replacer` creates one `StreamCache` per
+  `RemoteTable` node without knowing how many times DuckDB will scan it.
+  `max_readers` is a memory-eviction hint derived from the sqlglot AST; omitting
+  it is safe — the cache simply retains all batches.
 - **Thread-safe reads.** Concurrent `StreamCacheReader` handles share an
   `Arc<Mutex<DatasetInner>>` and are designed for concurrent use.
-- **Prompt resource release.** Explicit `close()` at execution boundaries releases
-  memory immediately rather than waiting for GC.
-- **`created` cardinality is natural.** One entry per unique `RemoteTable` node.
+- **Prompt resource release.** `RemoteTableScope.close()` releases tables, caches,
+  and readers in dependency order. Eager call sites use `remote_table_scope`;
+  streaming call sites use `bind_scope_to_reader`.
 - **Fixes #983 and the full multi-scan class.** ASOF join with `tolerance` and
   self-join both work correctly.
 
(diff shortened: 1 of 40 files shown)


Barre:
I think gizmosql, pyiceberg, snowflake, postgres, and databricks backends should get a StreamCache guard in their read_record_batches as well.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
dlovell, line 1: we already have `docs/adr/0012-click-option-decorators-and-runner-hierarchy.md`

mesejo:
> I think gizmosql, pyiceberg, snowflake, postgres, and databricks backends should get a StreamCache guard in their read_record_batches as well.

If you are referring to wrapping the StreamCache in a RecordBatchReader, I think you don't actually need it; the new version of batchcorder is fully compatible with pyarrow

Aditionally snowflake, Postgres, and Databricks have no benefit from using a replayable RecordBatchReader because we create durable tables from it in those backends.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
```diff
@@ -0,0 +1,212 @@
...
+`RemoteTable` would be consumed — before allocating the tee copies. This forced a
+two-pass graph walk:
+
+1. Walk the full graph, counting occurrences of each `RemoteTable` expression hash.
+2. Walk again, allocating `SafeTee(upstream, count)` on first encounter and popping
+   one tee copy per subsequent encounter.
+
+Problems:
+
+- **Pre-counting is fragile.** DuckDB can scan a registered source more than once
```
dlovell: the objective of using batchcorder is not to eliminate pre-counting, we will still be doing pre-counting

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
```diff
@@ -0,0 +1,212 @@
...
+- **Thread safety is incomplete.** `itertools.tee` is explicitly documented as not
+  thread-safe. DuckDB scans registered tables from parallel threads within a single
+  query.
+- **`created` cardinality was inflated.** The function returned one entry per tee
+  copy, so a self-join with `n=3` copies produced `len(created) == 3` even though
+  only one remote table was involved.
+
+## Decision drivers
+
+- Multi-scan must be transparent — no pre-counting at call sites.
```
dlovell: this is not a decision driver

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
```diff
@@ -0,0 +1,212 @@
...
+
+- **Pre-counting is fragile.** DuckDB can scan a registered source more than once
+  inside a single query (ASOF joins, self-joins, joins with intermediate aggregations).
+  The scan count is not part of DuckDB's public API and varies by query shape,
+  DuckDB version, and optimiser decisions. Undercounting exhausted the tee pool;
+  overcounting leaked unconsumed iterators.
+- **Thread safety is incomplete.** `itertools.tee` is explicitly documented as not
+  thread-safe. DuckDB scans registered tables from parallel threads within a single
+  query.
+- **`created` cardinality was inflated.** The function returned one entry per tee
```
dlovell, line 91: is the claim that SafeTee is not necessary and a self-join on a record batch reader based table works?

Barre:
This blows up on duckdb:

```python
rt = xo.memtable(df).into_backend(duckdb_con, "t")
rt.mutate(s=rt.v.sum().over(xo.window(group_by="k"))).execute()
# ValueError: Maximum number of readers (1) has been reached
```

count_remote_table_readers says 1 because there's one table ref in the SQL, but duckdb rewrites aggregate windows with PARTITION BY (and no ORDER BY) into a group-by joined back on itself, so it scans the source twice. You can see it in EXPLAIN, there's no WINDOW operator at all, just HASH_JOIN(SEQ_SCAN, GROUP_BY(SEQ_SCAN)). Forcing the count to 2 makes it pass. Same thing one level deeper: join a partition-window subquery back to its base and you get predicted 2 / needed 3.

I poked at a bunch of other shapes (self joins, unions, scalar subqueries, row_number, ordered windows, asof) and they all count correctly, so I think the counter itself is fine.

Some leak paths I ran into while testing failure cases:

- api.py:479: `con.to_pyarrow_batches` runs before clean_up exists and isn't guarded, so any planning error (e.g. `rt.mutate(c=rt.b.cast("int64"))` on a string column) leaves the caches open and the placeholder tables registered. _pandas_execute and get_plans both got the try/finally, this one didn't.
- _transform_expr: caches are created at line 415 but _transform_deferred_reads right after can raise (deferred read of a missing file), and then nobody ever gets the caches to close them.
- clean_up itself: the drop_view in the except is unguarded. On duckdb register() creates a view, so drop_table always throws and drop_view is the path that actually runs if it ever fails, the cache.close() loop below never executes.

One more thing from checking the postgres path: StreamCache.close() doesn't close the reader it wraps. The upstream generator's finally only runs when the cache object gets collected, so an early-terminated scan (limit pushdown etc) keeps the ADBC connection open until GC even though we close() the cache everywhere. Worth holding onto the source reader and closing it next to cache.close(), or fixing in batchcorder. Also in the new to_pyarrow_batches, cur.execute is only caught for ADBCProgrammingError, so an OperationalError (network hiccup, the realistic failure) skips both cleanup branches and leaks the per-call connection.

Nit: the datafusion inner cache is built without max_readers, so the whole stream sits in memory for the duration of the query regardless of the count (~350MB resident for a ~320MB stream when I measured, vs ~60MB with max_readers=1). Not a leak since the table drop frees it, but it means the memory bounding effectively only exists on duckdb. And get_plans never drops the tables it registers.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
mesejo: Stale — file is now `0013`, no collision with `0012-click-option-decorators-and-runner-hierarchy.md`.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
mesejo: Reworded. The objective was never to eliminate pre-counting — we still pre-count for `max_readers`. The SafeTee problem was that correctness *depended* on an exact count; under batchcorder the count is only an eviction hint.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
mesejo: Removed that bullet from Decision drivers.

docs/adr/0013-batchcorder-stream-cache-for-remote-table-fan-out.md
mesejo: Yes — clarified in Decision: StreamCache makes the registered table replayable, so SafeTee is unnecessary and a self-join over a RemoteTable works (each side acquires its own StreamCacheReader replaying from batch 0).

dlovell:
Heads-up on a dependency from #2105 / #2107 (the `drain=True` read-ahead race fix).

This PR's `RemoteTableScope.close()` ordering — drains first (close-then-join), *then* tables → caches → readers — closes/joins the tee `DrainingIterator`s while their readers are still alive. That's exactly the cleanup path that opens the #2105 race: `close()` starts the background drain while datafusion's bounded read-ahead may still have an in-flight foreground `__next__`, and both then advance the same `write_through` generator → `ValueError('generator already executing')`, with the write silently never published.

That ordering is now only safe because #2107 makes `DrainingIterator` serialize foreground/background advancement on a shared per-batch lock. So there's a real merge-order dependency: this branch's drains-first close assumes the serialized `DrainingIterator` from #2107 (currently on `fix/tee-drain-concurrent-pull-race`). If this lands first, the multi-batch `drain=True` path can race.

Suggestion: rebase on / merge #2107 first, and add a one-line cross-reference in ADR-0013 where it describes the drains-first close ordering, pointing at ADR-0014 / #2105 for why that ordering is safe (ADR-0014 was updated in #2107 to document the serialization invariant).

PR #1979, opened 2026-05-21
docs(plans): template system redesign

paddymul:
## Summary

Plan-only PR. No code changes. Describes a redesign of `xorq init` to fix the recurring template ↔ xorq version drift bugs (e.g. the chain `00e62199`, `0feaf6c7`, `647cd3e1` that re-pinned template SHAs).

**Core idea:** templates ship `"xorq[extras] @ LATEST"` as a placeholder; `xorq init` resolves the running xorq's install mode (PyPI / editable / git / archive) to a concrete PEP 508 spec, rewrites `pyproject.toml`, deletes the (already-stale) shipped `uv.lock` and `requirements.txt`, and runs `uv lock`. CI runs `xorq init` against the in-tree xorq via editable auto-detect, so template drift fails CI **before** release instead of after.

**Why the current shipped `uv.lock` is "stale theatre":** every template's `pyproject.toml` has `[tool.uv.sources].xorq = { git = "https://github.com/xorq-labs/xorq" }`, which makes uv resolve xorq from HEAD of `main`. The shipped lock pins a specific commit, but that pin is stale within hours.

The plan walks ten resolved design decisions (placeholder syntax, dispatch table by install mode, `--xorq-spec` override, failure handling, rollout sequence across xorq + the three template repos) and an addendum surveying how `uv init` / `cargo init` / `npm init` / `pnpm init` / `poetry new` handle lockfiles in scaffolds — none of them ship one.

Reference: abandoned commit `d5de1380` covered ~40% of this design.

## Test plan

- [ ] Plan reviewed for completeness; no further open questions before implementation PR
- [ ] Implementation PR will follow on a separate branch, exercising the slow `test_init_uv_build_uv_run` against in-tree xorq

### DIFF
--- a/docs/plans/template_system_redesign.md
+++ b/docs/plans/template_system_redesign.md
@@ -1,0 +1,194 @@
+# Template System Redesign
+
+Status: plan finalized after design review. Ready to implement on a fresh branch off `main`.
+
+## Problem
+
+`xorq init` downloads a template repo (`xorq-labs/xorq-template-{name}`) at a pinned commit SHA. Each template ships its own `pyproject.toml`, `uv.lock`, and `requirements.txt`. The templates also carry `[tool.uv.sources].xorq = { git = "https://github.com/xorq-labs/xorq" }`, which makes uv resolve `xorq` from HEAD of main — so the version range in `dependencies` is decorative and the shipped `uv.lock` is stale within hours of being written.
+
+Today the only way to keep them in sync is the "arduous update cycle":
+
+1. Cut a new `xorq` release.
+2. Manually open each `xorq-template-*` repo, bump its xorq dep, re-export `uv.lock` and `requirements.txt`, commit.
+3. Bump the three commit SHAs in `python/xorq/init_templates.py`.
+4. Cut another `xorq` release.
+
+Consequences:
+
+- Bugs in the template ↔ xorq integration (e.g. the `xorq uv build` failures that drove `00e62199`, `0feaf6c7`, `647cd3e1`) are not caught until after release.
+- `test_init_uv_build_uv_run` papers over the issue by `rm requirements.txt` after init.
+- Tests can't exercise "current tree of xorq against current templates" — they always run the pinned, stale combination.
+
+An abandoned fix exists at commit `d5de1380` ("initial pass at rewriting the templates to latest version of xorq on init") — useful reference, ~40% of the final design.
+
+## Goals
+
+1. After `xorq init`, the generated project's `xorq` dependency matches the `xorq` that ran `init` — whether that was a PyPI install, an editable dev install, or a git install.
+2. The `xorq` test suite runs templates against the **current working tree** of xorq, so template breakage is caught in CI before release.
+3. Releasing xorq no longer requires bumping template repos and re-pinning SHAs.
+
+## Non-goals
+
+- Inlining the template files into this repo. Templates stay in `xorq-labs/xorq-template-*`.
+- Replacing `uv` / introducing a different package manager.
+- Reworking which templates exist or what they contain.
+- Preserving the ability to `git clone + uv sync` a template repo directly. By design, templates become usable only through `xorq init`. Template contributors iterate via `xorq init --xorq-spec '…' --template <name> --branch <pr-branch>`, which exercises the same path users hit.
+
+## Design
+
+### 1. Placeholder in template `pyproject.toml`
+
+Each template's `pyproject.toml`:
+
+```toml
+[project]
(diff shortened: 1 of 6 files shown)


docs/plans/template_system_redesign.md
```diff
@@ -0,0 +1,194 @@
...
+
+Status: plan finalized after design review. Ready to implement on a fresh branch off `main`.
+
+## Problem
+
+`xorq init` downloads a template repo (`xorq-labs/xorq-template-{name}`) at a pinned commit SHA. Each template ships its own `pyproject.toml`, `uv.lock`, and `requirements.txt`. The templates also carry `[tool.uv.sources].xorq = { git = "https://github.com/xorq-labs/xorq" }`, which makes uv resolve `xorq` from HEAD of main — so the version range in `dependencies` is decorative and the shipped `uv.lock` is stale within hours of being written.
+
+Today the only way to keep them in sync is the "arduous update cycle":
+
+1. Cut a new `xorq` release.
```
mesejo, line 11: nit: There is no need to cut a new xorq release to update the templates; the templates point to commits in the git repo

docs/plans/template_system_redesign.md
```diff
@@ -0,0 +1,194 @@
...
+`run_uv_lock`:
+- `subprocess.run(["uv", "lock"], cwd=path, check=False)`.
+- On failure: **leave the substituted `pyproject.toml` in place**, do not delete the target directory, print the uv error verbatim, exit non-zero. ("During development we want failures to scream; a dev needs to look at the half-done state.")
+- Skippable via `--no-lock`.
+
+### 4. CLI surface
+
+`xorq init` gains:
+
+- `--xorq-spec SPEC` — explicit override of the auto-detected spec. Full PEP 508 form (`"xorq[duckdb] == 0.3.25"`, `"xorq @ git+https://github.com/xorq-labs/xorq@abc123"`).
```
mesejo, line 106: Is the point of this command to allow the user to point to a different xorq version? The name of the parameter seems off, and of course, this could be done manually by the user after running xorq init

docs/plans/template_system_redesign.md
```diff
@@ -0,0 +1,194 @@
...
+- **cargo generate** (Rust): depends on the template author. Templates that represent finished applications sometimes ship `Cargo.lock`; skeleton templates usually don't.
+
+So if you squint: a fully-built, version-pinned template could justify shipping a lock. *But* the xorq case has a property those don't:
+
+### Why xorq templates are a special case
+
+The `xorq` dep's version is **literally not knowable at template-build time** — it depends on which `xorq` CLI the user runs `xorq init` with. A normal cookiecutter template could lock `pandas == 2.1.0`, `sklearn == 1.4.0`, etc. and ship a working lock. The xorq template cannot, because there's no fixed value for `xorq` until the user invokes init.
+
+This is why the current templates' shipped `uv.lock` is **not** a meaningful reproducibility artifact: the templates use `[tool.uv.sources].xorq = { git = "https://github.com/xorq-labs/xorq" }`, which resolves to HEAD of main, which moves continuously. The shipped lock pins a specific `xorq` commit, but that pin is stale within hours of being written. The lockfile's contract — "this graph reproduces" — is already broken before the user sees it.
```
mesejo, line 180: What is the meaning of stale here?

mesejo (commented):
Nice proposal! I really liked the use of  `@ latest`, I'm using something similar to test the templates in here #1978 

Overall, this proposal adds work to the init command that could be done offline, rather than every time a template is downloaded. 

Since this is PR for a template system redesign, it is also the place to discuss the following items: 

1. Do we want to keep the templates in different repos, or do we want to have a template mono-repo? My inclination is for a template mono-repo
2. Should we support user-defined templates  (not sure what that looks like)

paddymul:
> 
>     1. Do we want to keep the templates in different repos, or do we want to have a template mono-repo? My inclination is for a template mono-repo
So mono-repo would make the template list a dictionary of template name to a tuple of (address git repo url, path to template dir, and commit).  

I think the churn on the templates will decrease signficantly with this release, so the usefulnees of a mono repo woill decrease.
>     2. Should we support user-defined templates  (not sure what that looks like)

I think user defined templates would just be a URL to a git repo, or git repo at a commit.

Given that user supplied templates would be in a different form than our mono repo,  I think we can stick with separate repos

docs/plans/template_system_redesign.md
paddymul: Yes but you have to cut a new xorq release so that users of xorq can get access to those new templates.

PR #1988, opened 2026-05-23
fix(devcontainer): auto-resolve workspace to git root

dlovell:
## Summary
- Running `dev/devcontainer` from a subdirectory (not the repo root) mounted only that subdir as the container workspace, leaving `/workspaces/src` without a `.git` entry — git and Claude Code inside the container failed with "not a git repository"
- Adds a guard after workspace path resolution that detects subdirectories and auto-resolves to `git rev-parse --show-toplevel`, printing a note to stderr
- Skipped when the directory doesn't exist (only happens on `down`/`clean` paths for already-removed worktrees)

## Test plan
- [ ] Run `dev/devcontainer up` from the repo root — no change in behavior
- [ ] Run `dev/devcontainer up` from a subdirectory (e.g. `python/xorq/`) — should print `note: resolving workspace to git root: ...` and start normally
- [ ] Run `dev/devcontainer -w python/xorq up` — should also auto-resolve
- [ ] Run `dev/devcontainer down` for a non-existent worktree path — should skip the check and proceed

### DIFF
--- a/dev/devcontainer
+++ b/dev/devcontainer
@@ -62,6 +62,21 @@ else
     export DEV_WORKSPACE="$(cd "$DEV_WORKSPACE" && pwd)"
 fi
 
+# Auto-resolve to git root when workspace is a subdirectory of a checkout.
+# Mounting a subdirectory as the container workspace leaves /workspaces/src
+# without a .git entry, so git (and tools like Claude Code) can't find the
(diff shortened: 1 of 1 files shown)


dlovell:
merged in https://github.com/xorq-labs/xorq/pull/1974

PR #2010, opened 2026-05-28
refactor(packager): CliCapabilities registry for inner xorq flag detection

dlovell:
## Summary

- Replace `_inner_xorq_supports_emit_option` and `PackagedUnboundRunner._unbind_flags` with a single frozen `CliCapabilities` class that probes `xorq <subcommand> --help` and exposes `has_flag()`, `resolve_flag()`, and `optional_args()` methods
- Fix a latent `NameError` in `PackagedUnboundRunner._extra_args` when only non-unbind flags (`typ`, `batch_size`, `instream`) were set without unbind args
- Add 10 unit tests for `CliCapabilities` and migrate 7 existing tests to mock `CliCapabilities.probe` instead of branching on `"--help" in args`

## Test plan

- [x] All 91 existing packager tests pass
- [ ] Verify emit-path fallback still works when inner xorq lacks `--emit-build-path-to`
- [ ] Verify hyphenated/underscore unbind flag resolution against real inner xorq

### DIFF
--- a/python/xorq/ibis_yaml/packager.py
+++ b/python/xorq/ibis_yaml/packager.py
@@ -16,6 +16,7 @@ PackagedRunner   build directory → execution output (via `uv tool run xorq run
 import functools
 import itertools
 import json
+import logging
 import operator
 import os
 import shutil
@@ -45,6 +46,9 @@ from xorq.common.utils.process_utils import in_nix_shell
 from xorq.ibis_yaml.enums import DumpFiles
 
 
+logger = logging.getLogger(__name__)
+
+
 class UvToolRunError(subprocess.CalledProcessError):
     def __str__(self):
         parts = [super().__str__()]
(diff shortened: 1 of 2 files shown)


dlovell:
Cross-PR note re: #2117.

#2117 threads `--relocate-reads/--no-relocate-reads` through `xorq uv build` into the inner `xorq build` subprocess (`PackagedBuilder._build`), currently appending the flag **unconditionally** with a comment that `--no-relocate-reads` needs a new-enough inner xorq.

The `CliCapabilities` registry added here is the right home for gating that: once this lands, #2117 should emit the relocate flag via `CliCapabilities.has_flag("build", "--no-relocate-reads")` rather than unconditionally, so an old inner xorq that defaults to non-relocating and lacks the flag does not error.

Heads-up: both PRs edit `PackagedBuilder._build` and will conflict; the one merging second should reconcile by folding the relocate flag into the capability probe.

PR #2013, opened 2026-05-30
fix(flake): use wheel on darwin

hussainsultan:


ghoersti (approved):
successfully ran `nix develop` 


```
❯ nix develop
evaluating derivation 'git+file:///Users/ghoersti/repos/xorq#devShells.aarch64-darwin.default'...
bad JSON log message from the derivation builder: [json.exception.parse_error.101] parse error at line 1, column 53: syntax error while parsing value - invalid string: missing closing quote; last read: '"pyprojectOutputSet'
(nix:nix-shell-env) bash-5.3$ 
```

PR #2057, opened 2026-06-09
feat(sinking): ADR-0014 deferred write as a side effect (TeeNode)

mesejo:
## Summary

Adds **ADR-0014** and its **Phase 1 implementation**: a deferred write performed as a side effect of execution.

`expr.sink(ParquetSink(path, mode))` wraps the table in a transparent **`TeeNode`**: rows pass through unchanged while an inline generator pushes each pulled batch to a generic RecordBatch consumer, then publishes atomically. It is the write-side counterpart to `deferred_read_*`.

## Design

- **`TeeNode`** (`expr/relations.py`): transparent op, stripped before hashing so a sinked expr hashes identically to its parent. Holds a generic consumer in its `tee` field; references no concrete writer type and no terminal node. `register_and_transform_tee_nodes` rewrites it to a backend table fed by `_drive_tee`, **after** cache resolution so a downstream cache hit prunes it and the write never fires.
- **Consumer contract**: `read(batch)` / `commit()` / `abort()`. `_drive_tee` commits on full exhaustion, aborts on early stop or error (all-or-nothing).
- **`ParquetSink`** (`sinking/`): `create` (replace) / `append` (new file per run); stages to a temp file on `read`, renames into place on `commit` (atomic publish).
- A terminal **`SinkNode`** (deferred write via `methodcaller`) is recorded in the ADR as a **deferred future alternative**, not implemented here.

## Checklist

- [x] **1. `sink()` is deferred** — write happens at `execute()`, not at `.sink()`.
- [x] **2. `sink()` is passthrough** — evaluates to the parent's rows; pipeline continues.
- [x] **3. The writer never pulls** — consumer is push-fed (`read(batch)`); the only puller is downstream.
- [x] **4. Write what's pulled; move tmp→dest only on exhaustion** — `read` stages, `commit` (rename) fires only when the stream is fully consumed; `abort` otherwise.
- [x] **5. Cacheable with a TeeNode; cache hit ⇒ no write** — TeeNode stripped before hashing (transparent key); a downstream cache hit prunes it.
- [x] **6. No incremental loading or accumulation; `create` and `append` only.**
- [x] **7. Fan-out of 2 (one backpressure-capable consumer); N via chaining N−1 TeeNodes.**
- [x] **8. Side-effect consumers expected to never give backpressure** (inline lock-step generator).
- [x] **9. `expr.sink().schema` works like a normal expr** (`schema == parent.schema`).

## Tests

`python/xorq/tests/test_sink.py` (17 passed, 1 skipped): passthrough, write-back, hash transparency, cache-hit-no-write, create/append, consumer commit/abort, cross-backend (datafusion, pandas), and mixed ops (`deferred_read`, `into_backend`, `cache`, sink-in-middle, chained fan-out).

## Known limitation

The streaming tee **deadlocks on single-connection engines like duckdb** (its connection is not re-entrant: the tee pulls the parent while the same connection serves the outer query). Datafusion and the pandas path work. duckdb is a documented `skip`; fixing it (parent pulled on a separate connection/thread) is deferred. Recorded in the ADR.

## Deferred / out of scope

- Terminal `SinkNode` (write via `methodcaller`).
- Decoupled write throughput via a buffered tee (batchcorder, ADR-0013, forthcoming).
- Single-connection-engine support; native N-way fan-out node.

## References

- ADR: `docs/adr/0014-sinknode-deferred-sinks.md`
- [XOR-252](https://linear.app/xorq-labs/issue/XOR-252)


docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,305 @@
...
+- **Deciders:** Daniel
+- **Context area:** `python/xorq/sinking/` (new), `python/xorq/expr/relations.py`, `python/xorq/expr/api.py`, `python/xorq/vendor/ibis/expr/types/relations.py`
+
+## Context
+
+Xorq's caching system solves **memoization**: "have I computed this exact expression
+before?" Each `execute()` of a given expression is keyed by a content hash; identical
+expressions hit a cached file, different expressions miss and compute.
+
+It does not solve **accumulation**: "append today's result to yesterday's." A daily
```
dlovell: the goal of sink node is not to solve accumulation: its to solve write-as-side-effect

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,305 @@
...
+pipeline run produces a different expression every day (a new `snapshot_date` parameter,
+new source data), so by construction each run is content-keyed to a different file. There
+is no mechanism to write to a named, durable target across runs. dbt provides this with
+`materialized='incremental'`.
+
+This ADR proposes `SinkNode`, a new DAG node that defers a **write** decision to
```
dlovell: sink node doesn't defer a write decision

docs/adr/0014-sinknode-deferred-sinks.md
mesejo: Reframed in 6d1304cb: accumulation/dbt dropped entirely; the ADR is now framed around write-as-a-side-effect (SinkNode terminal write + TeeNode passthrough).

docs/adr/0014-sinknode-deferred-sinks.md
mesejo: Fixed in 6d1304cb: no longer framed as deferring a decision. It's a deferred write (the write action deferred to execution time, the counterpart to deferred_read_*), not a read-or-compute decision like CachedNode.

python/xorq/sinking/__init__.py
dlovell, line 1: significant logic should not be in `__init__.py`

dlovell: https://github.com/xorq-labs/xorq/issues/2066

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,292 @@
...
+separate terminal write node is deferred (see Alternatives), so `TeeNode` is the only node
+this ADR ships.
+
+`TeeNode` does not reference any concrete writer type. It holds a `SinkNode` in its `sink`
+field (see the `SinkNode` contract below) and is otherwise oblivious to what the sink does
+with the batches it receives.
+
+### `TeeNode` is hash-neutral, like `Tag`
+
+`TeeNode` is modeled on `Tag` (`relations.py:102`). Its schema equals its parent's schema,
```
hussainsultan, line 46: I think we want the build hash to change when a TeeNode is added, since it necessarily implies a Sink.

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,292 @@
...
+as if the tee were not there.
+
+### The `SinkNode` contract
+
+A `SinkNode` is an abstract base class with a single method:
+
+```python
+class SinkNode(abc.ABC):
+    @abc.abstractmethod
+    def execute(self, batches: Iterable[pa.RecordBatch]) -> Iterator[pa.RecordBatch]:
```
hussainsultan, line 59: `execute` is too similar to the execute method on an expression. Since SinkNode is not an expr, we should differntiate the execution method with something like `.sink(...)` or something close.

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,292 @@
...
+
+`TeeNode` is modeled on `Tag` (`relations.py:102`). Its schema equals its parent's schema,
+and it is **stripped before hashing** by a resolution pass that replaces it with its parent,
+exactly as `_remove_non_hashing_tag_nodes` (`api.py:368`) does for `Tag`. So `expr.tee(s)`
+and `expr` produce the **same** content hash, and a `CachedNode` above or below the tee keys
+as if the tee were not there.
+
+### The `SinkNode` contract
+
+A `SinkNode` is an abstract base class with a single method:
```
hussainsultan, line 54: potentially, `SinkNode` should not be called a Node because its not an expr node.

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,292 @@
...
+The user-facing method is `.tee()`, matching the Unix `tee` command: data flows through and
+a copy is written as a side effect. The consumer abstraction is named `SinkNode`, and the
+field on `TeeNode` is `sink`. This split is deliberate: "tee" names the pass-through
+behavior the caller sees, while "sink" names what the consumer does with the data it
+receives (writes it somewhere). A future terminal write node (see Alternatives) would use
+`SinkNode` directly without wrapping it in a `TeeNode`.
+
+## Alternatives considered
+
+### A terminal `SinkNode` (deferred write via `methodcaller`)
```
hussainsultan, line 210: we may not need a terminal SinkNode since the build hash of a TeeNode + Sink is different.

docs/adr/0014-sinknode-deferred-sinks.md
```diff
@@ -0,0 +1,292 @@
...
+- The streaming tee requires a backend whose reader can be pulled concurrently with the outer
+  query. Datafusion and the pandas path work; a single-connection engine like duckdb
+  **deadlocks**, because the tee re-enters the same connection to pull the parent while that
+  connection is serving the outer query. Supporting such engines needs the parent pulled on a
+  separate connection or thread, which is deferred.
+- `BackendSink`'s per-batch path (for backends with `mode` support) commits each batch
+  individually, so a mid-stream failure leaves earlier batches written. This is not
+  all-or-nothing like `ParquetSink`; rollback requires the backend's own transactional support.
+
+## References
```
hussainsultan, line 283: whats the behavior for `..ls.untagged`

PR #2065, opened 2026-06-12
Transform scope

Barre:
`TransformScope` owns everything the transform materializes: upstream readers, StreamCaches, and the placeholder tables registered in target backends. Resources are adopted the moment they're acquired inside the replacer, and close() is idempotent and step-isolated, so a failed drop can't skip the remaining cleanups. Call sites reduce to:

```python
with transformed(expr, params=params) as texpr:
    return con.execute(texpr)
```

The streaming path keeps its deferred cleanup via `scope.bind(reader)`: generator finally on full drain, plus a weakref.finalize backstop for abandoned readers.

This fixes the leaks from the review thread: caches and placeholder tables stranded when planning or a deferred read fails after registration, the unguarded drop_view in clean_up skipping all cache closes on failure, and get_plans never dropping the tables it registers. One deliberate exception: prepare_create_table_from_expr closes with `drop_tables=False` because snowflake's create_table runs CTAS against the placeholders after it returns.

### DIFF
--- a/python/xorq/caching/storage.py
+++ b/python/xorq/caching/storage.py
@@ -203,15 +203,13 @@ class SourceStorage(CacheStorage):
 
         if is_remote(value):
             if is_single_backend(self, value):
-                from xorq.expr.api import _transform_expr  # noqa: PLC0415
+                from xorq.expr.api import transformed  # noqa: PLC0415
 
                 # must transform for Read ops: create_table expects a vanilla ibis expr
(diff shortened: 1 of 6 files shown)


dlovell:
can we change the name so its apparent what is being managed? perhaps one of
- remote_tables_materialized
- remote_table_scope
- materialized_remote_tables

mesejo (approved):

PR #2089, opened 2026-06-17
feat: add TeeNode for deferred sinks

mesejo:
## Summary

Adds **deferred writes** to xorq, symmetric to the existing deferred reads (`deferred_read_parquet`/`deferred_read_csv`/`Read`). An expression can now write its rows to a target *as a side effect of execution* and keep going, via a hash-neutral pass-through `TeeNode` that drives a `WriteThrough`.

See `docs/adr/0014-teenode-deferred-writes.md` for the full design rationale.

## What's in here

- **`python/xorq/writes/`** (new): `WriteThrough` ABC (`write_through(batches) -> Iterator[RecordBatch]` — pull, write each batch as a side effect, yield onward), concrete write-throughs (`ParquetWriteThrough`, `BackendWriteThrough`, `ThreadedBackendWriteThrough`, `WritePrimaryWriteThrough`), and `enums` (`WriteMode`).
- **`TeeNode`** (`expr/relations.py`): pass-through node modeled on `Tag`. Evaluates to its parent's rows unchanged; hands each batch to a `WriteThrough` as it streams through.
  - **Hash-neutral**: schema equals parent's, stripped before hashing (like `_remove_non_hashing_tag_nodes`). `expr.tee(s)` and `expr` produce the same content (cache) hash, so caching keys as if the tee weren't there. The build hash (`get_expr_hash`) does include the writer identity so different writers produce different build artifacts.
  - **Respects the cache**: if downstream is served from cache and data is never pulled, the write never fires. Cache *hits* are suppressed structurally — the cache pass prunes the `TeeNode` before the tee pass runs.
- **Drain-on-close** (`drain=True` on `.tee()`): `DrainingIterator` wraps the write-through generator so that when downstream stops early (e.g. `LIMIT`), a background thread drains the remaining parent stream to completion. Lock-guarded, thread-safe close/join lifecycle, errors surfaced through `join()`. `BaseExceptionGroup` used for multi-failure cleanup on Python 3.11+ (`raise_collected_errors` in `common/compat.py`).
- **`.tee()`** user-facing API (`expr/api.py`, vendored `ibis/expr/types/relations.py`), polymorphic on target: a `WriteThrough`, a backend connection (builds `ThreadedBackendWriteThrough`, requires `table_name`), or a path (builds `ParquetWriteThrough`).
- **Cross-platform file locking**: `flock_exclusive()` in `common/compat.py` (fcntl on POSIX, msvcrt on Windows).
- Supporting changes in `caching/strategy.py`, `dasher/_opaque.py`, `graph_utils.py`, `provenance_utils.py` for hashing/graph-transform handling. `register_and_transform_tee_nodes` uses `op.replace` (not `replace_nodes`) to avoid firing side effects inside opaque sub-expressions.
- **Hash version bump v3 → v4** (`dasher/_opaque.py`, `_recompute.py`): the `Tag`/`HashingTag` tokenization change plus the tee plumbing invalidate all previously computed build/cache hashes. Snapshot fixtures are updated accordingly.

## Tests

- `python/xorq/tests/test_write_through.py` (new, ~1,550 lines): covers `TeeNode`, `WriteThrough`, `DrainingIterator`, hashing neutrality, drain-on-close, temp table cleanup, concurrency, and graph transforms.
- Updates to `test_dasher.py`, `test_cli.py`, `test_ls_accessor.py`, and compiler snapshots.


python/xorq/expr/relations.py
```diff
@@ -647,12 +681,85 @@ def replacer(node, kwargs):
...
+            node = node.__recreate__(kwargs)
+        parent_expr = node.parent.to_expr()
+        con, _ = find_backend(node.parent, use_default=True)
+        if con.name in _NON_REENTRANT_TEE_BACKENDS:
+            warnings.warn(
+                f"tee() on a {con.name!r} backend is likely to deadlock: the "
+                "streaming write pulls the parent reader on the same single "
+                "connection that serves the outer query. Phase 1 targets "
+                "engines that allow concurrent reader pulls (e.g. datafusion). "
+                "See ADR-0014.",
```
mesejo, line 734: nit: not sure how to see ADR-0014 helps a user. Aren't ADR just for developers of the library?

python/xorq/caching/strategy.py
```diff
@@ -141,13 +141,9 @@ def _build_hasher(self, expr):
         return snapshot_hasher(*extra)
 
     def _replace_remote_table(self, expr: Expr, local_hasher: Hasher) -> Expr:
-        from xorq.common.utils.graph_utils import (  # noqa: PLC0415
-            replace_nodes,
-            walk_nodes,
-        )
         from xorq.expr.relations import RemoteTable  # noqa: PLC0415
 
-        if walk_nodes((RemoteTable,), expr):
```
mesejo, line 150: why was this changed?

python/xorq/expr/relations.py
dlovell: A user might read tests to understand expected outcomes. Similarly a user might read an ADR to understand desired and future properties.

python/xorq/caching/strategy.py
dlovell: `SnapshotStrategy._replace_remote_table` is vestigial from pre-`xorq-dasher` hashing. This particular change was probably a spurious change related to concerns about side-effect inducing walking.

I'm pushing a commit that removes this method entirely.

PR #2099, opened 2026-06-22
refactor(api): bundle execution cleanup into ExecutionResources

dlovell:
## Summary

Pure, behavior-preserving refactor that consolidates execution-cleanup state and logic into a single `@frozen` attrs class, `ExecutionResources`.

- Replace the loose `(created, drains)` tuple threaded out of `_transform_expr` and the three standalone helpers (`_close_and_join_drains`, `_drop_created_tables`, `_run_cleanup`) with `ExecutionResources` holding `created` + `drains`, exposing `cleanup()` (and the two private steps as methods).
- `_transform_expr` now returns `(expr, resources)` instead of `(expr, created, drains)`.
- `_pandas_execute` / `to_pyarrow_batches` call `resources.cleanup()`; `to_pyarrow_batches` drops its `clean_up` closure and hands the bound `resources.cleanup` method straight to `rbr_wrapper`.
- Non-executing / discarding call sites (`get_plans`, `caching/storage.py`, `relations.py:prepare_create_table_from_expr`, vendor `relations.py` view path) just unpack `(x, _)` — behavior unchanged.
- Cleanup ordering (close+join drains, then drop tables) and `raise_collected_errors` aggregation are preserved exactly.
- Fields are precisely typed: `created: dict[str, BaseBackend]`, `drains: list[DrainingIterator]` (matching the `register_and_transform_*` signatures; `DrainingIterator` imported under `TYPE_CHECKING`).

### Why

Two recently-landed features (TeeNode deferred writes, and incoming batchcorder StreamCache fan-out) both grow the "resources to release after execution" return value of `_transform_expr` and duplicate per-call-site cleanup. Bundling them into one object means the next addition is a single new field + a single new step in `cleanup()`, with no tuple-arity changes and no new `finally` blocks at call sites.

## Test plan

- [x] `ruff check` clean on all changed files; pre-commit hooks pass
- [x] cleanup-semantics tests in `test_write_through.py` migrated to construct `ExecutionResources` and pass (5/5)
- [x] `test_write_through.py` + `params/test_param_output_methods.py`: 116 passed, 1 skipped, 0 failed (6 repeated runs, serial; also clean under `-n auto`)
- [x] Re-checked the earlier "3 tee tests fail on clean `main`" note: could **not** reproduce. 12 serial + 8 xdist runs of `test_write_through.py` on `main` were all green, and ~40 runs total across branch/main never produced `generator already executing`. If those failures are real they appear to be a load/machine-dependent thread race outside this diff's surface — flagging for CI confirmation rather than asserting it as fact.

### DIFF
--- a/python/xorq/caching/storage.py
+++ b/python/xorq/caching/storage.py
@@ -215,7 +215,7 @@ class SourceStorage(CacheStorage):
                 from xorq.expr.api import _transform_expr  # noqa: PLC0415
 
                 # must transform for Read ops: create_table expects a vanilla ibis expr
-                (transformed, _, _) = _transform_expr(value.to_expr())
+                (transformed, _) = _transform_expr(value.to_expr())
                 self.source.create_table(key, transformed)
             else:
                 assert hasattr(self.source, "read_record_batches")
--- a/python/xorq/expr/api.py
+++ b/python/xorq/expr/api.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Mapping
 
 import pyarrow as pa
 import toolz
+from attr import field, frozen
 
 import xorq.vendor.ibis.expr.datatypes as dt
 import xorq.vendor.ibis.expr.operations as ops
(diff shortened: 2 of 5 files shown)


dlovell:
Closing as superseded.

This PR's sole justification was forward-looking: bundle `(created, drains)` into `ExecutionResources` so the incoming batchcorder StreamCache fan-out (#1977) would be a single new field + a single new `cleanup()` step. That bet was placed before #1977's code existed.

Now that #1977 is rebased onto current `main`, it independently introduced a strict superset of this idea — `RemoteTableScope` (in `python/xorq/expr/remote_table_exec.py`) — which:

- tracks readers, caches, tables, and drains as typed `_ScopeEntry` lists (not just `created: dict[str, BaseBackend]` + `drains`),
- has an ordered `close()` (drains → tables → caches → readers, reversed) with per-entry error aggregation,
- is a context manager with a `remote_table_scope()` wrapper and `bind_scope_to_reader()` for the lazy reader path,
- lives next to `register_and_transform_remote_tables`, which actually mints the resources, rather than in `api.py`.

Consequences:
- The motivation for this PR is moot — the "next addition" it was meant to ease already exists and didn't use it.
- `ExecutionResources`' `dict[str, BaseBackend]` shape can't even represent the StreamCache/raw-reader resources #1977 needs, so the abstraction is inadequate for the feature it cited.
- Landing this first would force #1977 to rip out `ExecutionResources` across all five files and replace it with `RemoteTableScope` — pure conflict + revert.

The valuable nugget (one closeable object owns post-execution cleanup) is already realized, more fully and in the right place, inside #1977. The shared dead-helper cleanup (`_close_and_join_drains` / `_drop_created_tables` / `_run_cleanup`) will be handled there.

PR #2113, opened 2026-06-24
 docs(adr): ADR-0017 — incremental WAP publish (upsert/merge/delete)

ghoersti:
Proposes extending the WAP library (ADR-0014) from append-only to incremental publish —
  insert/update/delete a keyed changeset into a target — across all xorq backends. Design doc only;
  no code in this PR (Status: Proposed).

  Decision
  - One user knob, PublishMode (APPEND | UPSERT | MERGE), mirroring the sink's WriteMode. APPEND ==
  today's behavior, so existing WAP is unified under it at zero new code. MERGE carries deletes via
  an _op column ('D' deletes, else upsert).
  - Mechanism is auto-routed, not user-chosen — PublishStrategy is resolved from the backend via
  lazy_singledispatch (register-by-string, the datafusion/pandas idiom): no backend imports at load,
  decision table in one file.
  - Five publish functions cover every backend (no per-backend logic): native MERGE INTO
  (duckdb/pg15+/snowflake/databricks/trino, one sqlglot statement), iceberg
  Transaction.upsert+delete (one snapshot), statement DML in a transaction (sqlite/pg<15), and an
  engine-agnostic anti-join + union-all rewrite for the immutable engines (datafusion/xorq/pandas).
  - Every tier is atomic via verified primitives (single MERGE; one iceberg Transaction;
  con.begin(); temp-rename / create_table(overwrite=True)).
  - WAP skeleton unchanged — reuses make_wap_expr's tee→audit→publish; a vacuous audit_fn keeps the
  aggregate pipeline-breaker (staging-commits-before-publish) when no audit is wanted.

### DIFF
--- a/docs/adr/0017-incremental-wap-publish.md
+++ b/docs/adr/0017-incremental-wap-publish.md
@@ -1,0 +1,525 @@
+# ADR-0017: Incremental WAP publish — one `PublishMode`, capability-routed `PublishStrategy`
+
+<!-- Provisional number: 0015 and 0016 are in flight on remote branches; bump if they land first. -->
+
+- **Status:** Proposed
+- **Date:** 2026-06-24
+- **Deciders:** George (george@xorq.dev)
+- **Context area:** `python/xorq/writes/` (`wap.py`, `enums.py`, new `incremental.py`), `python/xorq/backends/pyiceberg`, `python/xorq/backends/xorq_datafusion`
+
+## Context
+
+The WAP library (`python/xorq/writes/wap.py`, ADR-0014's TeeNode underneath) is **append-only**.
+`make_wap_expr` writes a staging artifact via a tee, audits the full stream with an aggregate
+pipeline-breaker, and the publish step *concatenates* staging into final — parquet merges row
+groups (`make_publish_with_parquet`), iceberg `add_files`/fast-forwards a branch
+(`publish_staging_table` / `publish_branch`). There is no way to publish a **changeset**:
+insert-or-update by key, or apply deletes.
+
+We want incremental publish — upsert and full merge (insert/update/delete) into a target —
+across xorq's backends. The backends are **heterogeneous** in merge capability, which is the crux:
+
+- **Native single-statement `MERGE INTO`:** duckdb, postgres (15+), snowflake, databricks, trino (ACID connectors).
+- **Row-level upsert + delete, no `MERGE`:** pyiceberg (`Backend.upsert`, `python/xorq/backends/pyiceberg/__init__.py:197`).
+- **`UPDATE`/`DELETE` but no portable `MERGE`:** sqlite.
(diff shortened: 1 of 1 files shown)


docs/adr/0017-incremental-wap-publish.md
```diff
@@ -0,0 +1,525 @@
...
+    @make_pandas_udf(schema=schema({STAGING: str, FINAL: str, PASSED: bool}),
+                     return_type=dt.boolean, name=f"incr_publish_{strategy.value}")
+    def publish(df):
+        if len(df) != 1:
+            raise ValueError(f"expected 1 row, got {len(df)}")
+        row = df.iloc[0]
+        if not row[PASSED]:
+            return [False]                              # audit failed -> staging retained
+        dispatch = {
+            PublishStrategy.APPEND:        _publish_append,
```
mesejo, line 180: I think in here we could use functools.single_dispatch or just a match statement? What do you think?

ghoersti: So I tried that but I uncovered an issue for us, initially on dev machine with all backends it was fine but on single backend installs it was revealed. 

I could have changed vendored ibis or switched to use routing from the `mro` which I went with. 

I pushed this example up to this branch. 

https://github.com/xorq-labs/xorq/blob/george/demo/lazy-singledispatch-single-backend/dispatch_demo.py

we could also have solved in vendored ibis like:
#python/xorq/vendor/ibis/common/dispatch.py,

```python
Before:
      package = klass.__module__.split(".", 1)[0]
            if lazy := self.lazy_lookup.get(package):
                # Import all lazy implementations first before registering
                # (which should never fail), to ensure an error anywhere
                # doesn't result in a half-registered state.
                new = {import_object(name): func for name, func in lazy.items()}
                self.lookup.update(new)
                # drop lazy implementations, idempotent for thread safety
                self.lazy_lookup.pop(package, None)
                return self.dispatch(typ)

After:
            # 2. Check lazy implementations
            package = klass.__module__.split(".", 1)[0]
            if lazy := self.lazy_lookup.get(package):
                # Import the package's lazy implementations, skipping any whose
                # module isn't importable — e.g. an optional backend registered
                # under this package whose driver isn't installed. Skipping is
                # safe: dispatch only ever needs the impl for a type we hold an
                # instance of, and that type's module is necessarily importable,
                # so a skipped registration is never one that would be selected.
                # (A missing *attribute* surfaces as plain ImportError from
                # import_object and still propagates — that's a real bug, not a
                # missing dependency.)
                new = {}
                for name, func in lazy.items():
                    try:
                        new[import_object(name)] = func
                    except ModuleNotFoundError:
                        continue
                self.lookup.update(new)
                # drop lazy implementations, idempotent for thread safety
                self.lazy_lookup.pop(package, None)
                return self.dispatch(typ)
```

ghoersti:
Closing because will be superseded by https://github.com/xorq-labs/xorq/pull/2129

PR #2117, opened 2026-06-25
feat(cli): expose cache pin/unpin for builds and catalog entries

dlovell:
## Summary

Exposes the `expr.ls.pin()` / `expr.ls.unpin()` transforms through the CLI, for both build directories and catalog entries.

- **Build-level** (`xorq pin` / `xorq unpin <build>`): load → pin/unpin → rewrite a build artifact. Options: `--builds-dir`, `--cache-dir`, `--ensure-materialized/-e` (populate caches before pinning), `--relocate-reads` (portable self-contained bundle).
- **Catalog-level** (`xorq catalog pin` / `unpin <entry>`): re-adds the result as a new content-named entry (pinning changes the build hash, so it's never in-place). Options: `--alias`, `--move-aliases` (move every alias from the source entry), `--relocate-reads`, `--cache-dir`, `--ensure-materialized/-e`.
- Pinning a cold (unmaterialized) cache surfaces a clear, actionable CLI message instead of writing a silently-broken entry ("Populate the caches first … or pass `--ensure-materialized/-e`").
- Shared option decorators (`relocate_reads_option`, `ensure_materialized_option`) keep the flag, default, and help text consistent across `build`, `pin`/`unpin`, and `catalog pin`/`unpin` — the help noun (build vs. entry) and the frozen-caches mention (pin only) are the only differences.
- Docs CLI-reference generator updated to group the four new commands.

## Engine support

This is no longer pure CLI plumbing — a few expr/engine-level changes back the CLI:

- **Relocated builds are load+rebuild hash-stable.** A single pre-hash pass (`_prepare_relocatable_reads`) marks local-file reads relocatable and bakes in the serialized `read_path` *before* the build is hashed; the write pass (`ExprDumper._prepare_relocatable_read`) derives the same string from `relocatable_read_path_str` — the single source of the `reads/<content-hash>` layout — so a fresh build and every later load+rebuild hash identically and the two passes cannot drift.
- **`--ensure-materialized/-e` populates cold caches before pinning.** `pin_cache(ensure_materialized=True)` checks each `CachedNode` and runs a `head(0)` execute per cold subtree (already-warm caches are skipped, downstream compute never runs), so `-e` is a real guarantee rather than a best-effort hint. `head(0)` fully writes each cache — cache population is a side effect of the execute pass, independent of the read-back row limit.
- **A cold-cache pin without `-e` raises a `ClickException`** with an actionable hint (via the shared `apply_pin_transform`) instead of writing a silently-broken entry.
- **A missing bundle source raises a clean `ClickException`** (via the shared `raise_for_missing_relocation_source`) telling you to restore the file or re-run with `--no-relocate-reads`, for both `xorq pin`/`unpin` and `xorq catalog pin`/`unpin`. Only a genuinely-missing *source* is translated; unrelated `FileNotFoundError`s (e.g. inside the build/cache dirs, or a loaded catalog entry's extract dir) surface raw.
- **The inner-xorq option probe fails loudly on a broken environment.** The packager probes the bundled ("inner") xorq once via `xorq build --help` to learn which options it accepts (`--relocate-reads`, `--emit-build-path-to`) before passing them. `xorq build --help` exits 0 on any xorq that has the `build` subcommand, so a non-zero exit means a broken inner env, not a missing feature — it now raises a clear `RuntimeError` instead of silently omitting `--relocate-reads` and producing a machine-local build behind a misleading "does not support --relocate-reads" warning. The version-fallback path is reached only when help genuinely succeeds but the option is absent.

## ⚠️ Behavior change: `xorq build` now defaults to `--relocate-reads`

`xorq build` previously left local-file reads in place by default; it now bundles them into the build artifact (`--relocate-reads` defaults on). Pass `--no-relocate-reads` to restore the previous machine-local behavior.

**Rationale:** a build exists to be added to a catalog, and a catalog entry exists to run from anywhere — so a build should be self-contained by default. Remote reads (`s3://`, `gs://`, …) are already location-independent and are left in place regardless.

**Migration note:** relocation changes build content and therefore the build hash, so pipelines that pin or compare build hashes will see new hashes after this change. Add `--no-relocate-reads` to any build whose hash must stay stable.

⚠️ **`--no-relocate-reads` is not a blanket hash-preserver.** The pre-hash pass now bakes a read's bundled `read_path` in whenever the read is *already* relocatable — including reads constructed via `deferred_read_parquet(..., relocatable=True)` — regardless of the flag (this is what makes relocated builds load+rebuild hash-stable). So a pipeline that uses the relocatable read API will see its build hash change on upgrade **even with `--no-relocate-reads`**; the flag only prevents relocation of reads that are still machine-local. Pipelines that must hold a hash constant across this upgrade should re-pin/re-record against the new hash rather than assuming `--no-relocate-reads` reproduces the old one.

**Relocation is one-way.** Bundling replaces a read's original path with a content hash, so `--no-relocate-reads` only prevents *new* bundling — it cannot un-bundle an already-relocated build. To keep a build lean, pass `--no-relocate-reads` at `xorq build` time; you can't recover a machine-local build by unpinning one that was already relocated.

**API defaults:** the programmatic `build_command(...)` matches this CLI default (`relocate_reads=True`). The lower-level `build_expr(...)` and `Catalog.add(...)` intentionally keep the lean default (`False`) — relocated reads aren't yet resolved in the fuse/bind execute path, so `Catalog.add(expr, relocate_reads=True)` entries can't be fused today. Aligning those defaults with the CLI is tracked in #2133.

⚠️ **`catalog pin`/`unpin` are subject to the same gap.** They already re-add with `relocate_reads=True`, so a pinned catalog entry that contains a relocated read (e.g. its bundled cache parquet) currently returns an **empty result when consumed via fuse/bind** — the same #2133 root cause. Guarded by `test_fuse_pinned_entry_with_relocated_read` (xfail, strict), which flips to a failure the moment #2133's fuse-path fix lands and prompts us to turn it into a normal passing regression test.

## Dependency

Stacked on **#2111** (`CacheTag` + public `.pin()`/`.unpin()`). Base is `feat/cache-pinning`; retarget to `main` once #2111 merges.

Implements `plans/cli-pin-unpin.md`.

## Test plan

- [x] `python/xorq/tests/test_cli_pin.py` (16 tests) — build round-trip, cold-cache→error + `-e`, unpin inverts, build-hash round-trip invariant, partial materialization, `--relocate-reads`, full-cache contract across backends, no-op equivalence, and a real-subprocess `xorq pin`/`unpin --help` cold-start smoke test
- [x] `python/xorq/catalog/tests/test_cli_pin.py` (14 tests) — new-entry, `--alias`, `--move-aliases`, no-op, unpin inverts, default-relocate round-trip, cold-cache hint, `--no-relocate-reads` lean entry, `--relocate-reads` bundles source read, and a real-subprocess `xorq catalog pin`/`unpin --help` cold-start smoke test
- [x] `python/xorq/ibis_yaml/tests/test_compiler.py` — relocated-build load+rebuild hash stability; `ArtifactStore.copy_file` same-path no-op (self-copy skip)
- [x] `python/xorq/catalog/tests/test_bind.py::test_fuse_pinned_entry_with_relocated_read` — xfail(strict) reproducing #2133 for the pin path (fused pinned entry with a relocated read yields an empty result / `schema names don't match input data columns`)
- [x] `docs/generate_cli_reference.py` runs clean

### DIFF
--- a/docs/generate_cli_reference.py
+++ b/docs/generate_cli_reference.py
@@ -78,6 +78,10 @@ MAIN_GROUPS = (
         "Serve",
         ("serve-flight-udxf", "serve-unbound"),
     ),
+    (
+        "Cache",
+        ("pin", "unpin"),
+    ),
 )
 
 CATALOG_GROUPS = (
@@ -102,6 +106,10 @@ CATALOG_GROUPS = (
     (
         "Composition and execution",
         ("compose", "run", "run-cached", "serve-unbound"),
+    ),
+    (
+        "Cache",
+        ("pin", "unpin"),
     ),
     (
         "Sync, audit, replay",
--- a/examples/pin_unpin.sh
+++ b/examples/pin_unpin.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Freeze a build's caches with `xorq pin` (rewrites each cache into a direct
+# read of its file) and thaw them with `xorq unpin`. --relocate-reads bundles
+# those files into the build for a self-contained, portable artifact.
+
+script="${1:?Usage: ./pin_unpin.sh <script.py> [-e expr_name]}"
+shift || true
+
+expr_name="expr"
+cache_dir="./pin-unpin-cache"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -e) expr_name="$2"; shift 2 ;;
+        --cache-dir) cache_dir="$2"; shift 2 ;;
+        *) echo "Unknown option: $1" >&2; exit 1 ;;
+    esac
+done
+
+echo "==> Building '${expr_name}' from ${script}"
+build_path=$(xorq build "$script" -e "$expr_name" --cache-dir "$cache_dir")
+echo "    build: ${build_path}"
+
+# Pin requires materialized caches; run once to populate them (or use `pin -e`).
+echo "==> Running build to materialize caches"
+# `head` closes the pipe early, so under `set -o pipefail` xorq's SIGPIPE (141)
+# would fail the script; `|| true` keeps the demo going. The caches are still
+# fully written -- materialization is a side effect of the run, not of how many
+# rows are read back.
+xorq run "$build_path" -f json -o - --cache-dir "$cache_dir" | head -n 5 || true
+
+echo "==> Pinning (freeze caches into direct reads, bundle them in)"
+# --relocate-reads is on by default; shown here for clarity.
+pinned_path=$(xorq pin "$build_path" --cache-dir "$cache_dir" --relocate-reads)
+echo "    pinned: ${pinned_path}"
+# Pinning intentionally changes the build hash: a distinct build, not an edit.
+
+echo "==> Unpinning (thaw frozen reads back into recomputable caches)"
+unpinned_path=$(xorq unpin "$pinned_path" --cache-dir "$cache_dir")
+echo "    unpinned: ${unpinned_path}"
+# Unpin reverses pin's cache transform. The hash only returns to the original
+# build's when reads are not relocated on either leg (relocate rewrites read
+# identity, so a relocated build is not load+rebuild hash-stable); see
+# `xorq pin/unpin --no-relocate-reads` for the hash-stable round-trip.
--- a/python/xorq/catalog/bind.py
+++ b/python/xorq/catalog/bind.py
@@ -3,8 +3,9 @@ from __future__ import annotations
 from functools import reduce
 
 from xorq.catalog.enums import CatalogTag
+from xorq.common.exceptions import UnsupportedOperationError
 from xorq.common.utils.graph_utils import replace_nodes, replace_unbound, walk_nodes
-from xorq.expr.relations import HashingTag, RemoteTable, gen_name
+from xorq.expr.relations import HashingTag, Read, RemoteTable, gen_name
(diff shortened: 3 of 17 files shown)


dlovell:
Follow-up: #2132 fixes the empty-commit behavior at the root (guards `commit_context` and unifies the two definitions). It's based on `main` and stands alone, so it's independent of this PR — the two are alternatives to the lightweight call-site guard discussed here.

dlovell:
Follow-up / cross-PR dependency with #2010.

This PR threads `--relocate-reads/--no-relocate-reads` through `xorq uv build` into the inner `xorq build` subprocess (`PackagedBuilder._build`), appending the flag **unconditionally**. `--no-relocate-reads` therefore requires an inner xorq new enough to accept it — the same version gate the existing `--emit-build-path-to` fallback deals with (noted in a code comment for now).

#2010 introduces the `CliCapabilities` registry (replacing `_inner_xorq_supports_emit_option`) with `has_flag()` / `resolve_flag()` — exactly the probe needed to gate this cleanly, so an old inner xorq (which defaults to non-relocating and lacks the flag) does not error.

Reconciliation needed: both PRs edit `PackagedBuilder._build`, so they will conflict. Whichever merges second should route the relocate flag through `CliCapabilities.has_flag("build", "--no-relocate-reads")` instead of the unconditional append.

hussainsultan (approved):

PR #2123, opened 2026-06-29
refactor(backends): extract xorq_datafusion and databricks impls into submodules (#2066 phase 2)

mesejo:
## Summary

Phase 2 of #2066. Both `xorq_datafusion/__init__.py` (980 lines) and `databricks/__init__.py` (670 lines) held full backend implementations — SQL compilation, UDF registration, type converters, connection management — rather than re-exporting from submodules.

This moves each implementation into a `backend.py` submodule and reduces the `__init__.py` to a thin re-export of `Backend` and `connect` (the only names external code imports).

## Changes

- `xorq_datafusion/__init__.py` → `xorq_datafusion/backend.py` + 9-line re-export
- `databricks/__init__.py` → `databricks/backend.py` + 11-line re-export
- `test_register.py`: repoint the private `_select_and_cast` import to `.backend`

## Why no breaking change

External code only imports `Backend` / `connect` from these packages (verified via grep across the repo). Entry points target the package (`xorq.backends.xorq_datafusion`), and both names are re-exported, so backend resolution is unchanged.

## Verification

- Import smoke: `api`, `flight`, `databricks_utils`, `config` all resolve `Backend`/`connect`
- `connect()` + `read_record_batches` runs
- `ruff check` clean
- `952 passed` (datafusion backend + `test_api`); databricks tests collect (21); benchmark-imports pass

## Notes

- Stacked on phase 1 (PR #2116). Until that merges, this PR's diff includes the phase-1 commits; it reduces to phase-2-only once #2116 lands. Merge #2116 first.
- Phase 3 (`@public` + lazy-import conventions) out of scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2129, opened 2026-06-30
feat(writes): incremental WAP publish across backends (XOR-444)

ghoersti:
## Summary

Extends the WAP (Write-Audit-Publish) library from append-only to **incremental publish** — upsert / merge / delete of a keyed changeset — across backends. One user-facing `PublishMode` (APPEND / UPSERT / MERGE) routes to a capability-matched `PublishStrategy` **declared by each backend** (`publish_strategy(self, mode)` on the backend class — capability is a backend fact, looked up on `type(con)`, importing nothing), dispatching to one of five publish mechanisms. APPEND is each strategy's no-key case, so existing append-only WAP becomes the `mode=APPEND` path of the same machinery (no behavior change).

Implements **XOR-444** and extends to most of **XOR-445** (postgres, gizmosql, databricks); only trino remains deferred — see Follow-up.

> **Reconciled with the parallel review on `dan/explore/incremental-wap`** — adopts that branch's reviewed ADR-0017 and folds its findings into the code: corrected dep pins, the REWRITE table-target non-atomicity note, and the duplicate-key contract (§3).

## 1. Design — ADR-0017

`docs/adr/0017-incremental-wap-publish.md` (Status: **Accepted**, reconciled with the landed implementation):

- **`PublishMode`** (APPEND / UPSERT / MERGE) — user-facing intent, a `StrEnum` like `WriteMode`.
- **`PublishStrategy`** — the capability-routed mechanism. Each backend that can do more than a full rewrite declares `publish_strategy(self, mode)` (postgres's is version-aware: `NATIVE_MERGE` on 15+, else `STATEMENT_DML`); a backend with no declaration falls back to `REWRITE` (with a debug log). Looked up on `type(con)` — never trips table-access `__getattr__`, and routing one backend imports none of the others, so a single-backend install (`xorq[duckdb]`) stays lean.
- **`StagingStrategy`** (TABLE / BRANCH) — see §4.
- The changeset contract — including **unique keys per delta** (§3) and the REWRITE table-target non-atomicity caveat.

## 2. Feature — `writes/publish.py`

`resolve_strategy(con, mode)` maps the backend to one of five mechanisms:

| backends | strategy | mechanism |
|---|---|---|
| duckdb / snowflake / postgres 15+ / databricks / gizmosql | `NATIVE_MERGE` | one `MERGE INTO`, rendered per-dialect (sqlglot) |
| pyiceberg | `UPSERT_DELETE` | `Transaction` upsert + delete, one snapshot |
| datafusion / xorq / pandas | `REWRITE` | anti-join + union-all |
| sqlite / postgres < 15 | `STATEMENT_DML` | UPDATE + INSERT + DELETE in one txn |
| parquet | (files) | pyarrow concat (APPEND) / pandas anti-join rewrite |

**All statements are authored once as strings** (`_q`-quoted identifiers) and rendered per dialect via `parse_one(...).sql(dialect=con.dialect)` — both the merge tier **and** statement DML. String authoring over the sqlglot builder API is deliberate: `exp.merge` needs sqlglot ≥26 while co-installation with ibis 9.5 (the `ci-test-ibis-compatibility` matrix) caps sqlglot <25.21 — a builder-API version briefly shipped on this branch and was reverted when that cell failed (tech debt tracked as **XOR-449**, recorded in ADR-0017 Alternatives). Strings render identically across sqlglot 24.0–28.6 (verified per version); the pin stays `>=23.4,!=26.32.0,<28.7.0`. One finding from the attempt is kept and pinned by test: sqlglot does **not** parenthesize an OR condition after the implicit AND in `WHEN MATCHED AND <cond>` — the not-delete predicate carries explicit parens, or merge semantics silently flip.

Correctness: `_op` is data for APPEND and a control column only for MERGE (`'D'` deletes, else/NULL upserts); key-only tables omit the empty `SET`; staging is consumed after publish; empty-input fails fast. `REWRITE`'s **table** target is `create_table(overwrite=True)` — drop→create, non-atomic (documented in the ADR); the **parquet** target uses an atomic temp+rename.

**Public API** (`xorq.writes`): `PublishMode`, `StagingStrategy`, the WAP builders `make_backend_wap_expr(con, *, key, mode, staging_strategy)` / `make_parquet_wap_expr(*, key, mode)`, and the **non-WAP** primitives `publish(con, staging, final, *, key, mode)` / `publish_parquet(...)` — the same reconciliation without the tee + audit gate. The layering: publish is the primitive; WAP is one consumer that adds write + audit in front of it (`wap` imports `publish`, never the reverse).

## 3. Duplicate-key gate (resolves ADR-0017 open Q5)

Key uniqueness within a changeset is a **data-level** invariant `_validate` cannot check (it sees only `mode`/`key`/`columns`, not rows), and the mechanisms **diverge** on duplicates: native `MERGE` and iceberg `upsert` raise a cardinality error; `REWRITE` and `STATEMENT_DML` resolve silently *and differently*. So the WAP builders **default `UPSERT`/`MERGE` to a no-duplicate-keys `audit_fn`** — the audit breaker already drains the full stream, so the check is essentially free — failing the publish before any tier runs. `APPEND` keeps always-True; the standalone primitives leave uniqueness a caller responsibility.

## 4. Iceberg branch WAP folds into the unified factory (`StagingStrategy`)

Iceberg's branch mechanism is no longer a separate code path: `make_backend_wap_expr(ice, mode=APPEND, staging_strategy=StagingStrategy.BRANCH)` stages on a branch **of `final` itself** and publishes by metadata-only fast-forward. Factory-time validated: APPEND-only (snapshot promotion is all-or-nothing — no keyed merge is meaningful), no key, and `type(con)` must declare `publish_branch` (same capability-lookup idiom as `publish_strategy`). The core sink hook widened to `make_sink(staging, final)`, which removes the redundant `table_name` binding — and fixes a latent bug where `table_name != final` staged on one table and fast-forwarded another. `make_iceberg_wap_expr` survives as a thin APPEND alias; existing examples/tests pass through it unchanged.

## Test plan

- `tests/test_incremental.py` — routing + `_validate` contract (parametrized over duckdb/snowflake/sqlite/iceberg/datafusion/pandas/databricks/gizmosql), SQL generation (identifier quoting/transpilation, the WHEN-condition paren regression, key-only MERGE), BRANCH factory-time validation, public-API exports.
- `tests/test_incremental_publish.py` — upsert / merge-with-deletes / append / first-run / NULL-op / key-only / identifier-quoting round-trips per strategy; end-to-end through the tee; the standalone primitives; the duplicate-key gate.
- `backends/pyiceberg/tests/test_wap.py` — both staging strategies incl. branch-via-unified-factory, retained-staging retry semantics, empty-input fail-fast.
- Gated per-backend live round-trips in `backends/<b>/tests/test_wap_incremental.py`: snowflake, postgres, gizmosql, databricks.

Local: the incremental suites + pyiceberg/parquet WAP pass (**83**), including a full run with `sqlglot==25.20.2` installed (the version the ibis-9.5 compat cell resolves) — green on both ends of the sqlglot range. All three WAP examples run (`wap_incremental_duckdb.py`, `publish_incremental_duckdb.py`, `wap_audit_iceberg.py`). CI: **postgres + gizmosql green** in their single-backend jobs; snowflake verified live; **databricks verified live on the current head** (`ci-test-databricks` run 28796214804 @ 3433cd38, post-revert). Rebased on `main` @ 5b72cf2f (post #2123 refactor — databricks `publish_strategy` moved to `backends/databricks/backend.py`).

## Split out

- **#2134** — snowflake unencrypted-keypair support (empty passphrase → `password=None`) + the loosened test-creds gate. Was enabling infra for the snowflake integration tests here; moved to its own PR.

## Follow-up — XOR-445

This PR covers the XOR-444 minimal set **plus** postgres, gizmosql, and databricks. The only backend left for [**XOR-445**](https://linear.app/xorq-labs/issue/XOR-445/incremental-publish-remaining-backends-adr-0017-follow-up):

- **trino** → `NATIVE_MERGE`, blocked on two real issues: it has no `read_record_batches` (so it can't be a tee sink for staging), and trino `MERGE` only works on ACID connectors (iceberg/delta/hive-ACID), needing connector detection + a guard/fallback. Until then trino declares no strategy and falls back to `REWRITE` (debug-logged).


ghoersti:
**Backend coverage in this PR**

Verified (APPEND / UPSERT / MERGE): duckdb, sqlite, datafusion, xorq_datafusion, pandas, parquet, pyiceberg, snowflake (XOR-444); **postgres** (`NATIVE_MERGE` pg 15+ / `STATEMENT_DML` fallback — CI postgres job); **gizmosql** (`NATIVE_MERGE`, duckdb dialect — local + CI).

**Two backends not landed:**

- **databricks — soft-blocked (a solution exists).** Routing is implemented + unit-tested and a gated round-trip test exists; it's unverified only because the workspace token is expired — `403 Invalid access token` in both `.envrc.secrets` and the `DATABRICKS_TOKEN` CI secret (a well-formed but stale `dapi…` PAT). A fresh PAT unblocks it; a real run may still need minor debugging (the `con` fixture's Volume + parquet `PUT` setup, the cross-backend staging write, Delta `MERGE` / `create_table`). Gated behind `@pytest.mark.databricks` (manual `ci-test-databricks`), so it never blocks CI.
- **trino — hard-blocked, no solution yet.** Deferred to XOR-445. Two blockers with no quick path: (1) the trino backend has no `read_record_batches`, so it can't be a tee sink for staging; (2) trino `MERGE` only works on ACID connectors (iceberg / delta / hive-ACID), needing connector detection + a guard/fallback.

docs/adr/0017-incremental-wap-publish.md
```diff
@@ -0,0 +1,595 @@
...
+```
+
+### Tier 3 — `STATEMENT_DML` (sqlite; postgres < 15 fallback)
+
+Genuine insert/**update**/delete in one transaction — *not* delete-then-reinsert, so in-place
+`UPDATE` semantics (triggers, identity columns, FKs) are preserved:
+
+```python
+def _publish_statement_dml(con, staging, final, key, columns, mode):
+    if not hasattr(con, "begin"):
```
mesejo: Did you check if we could use sqlglot for building the statements?

ghoersti:
**Note** : `make_iceberg_wap_expr` is kept as a thin alias, not removed. Branch publishing is now a first-class strategy on the unified factory (`make_backend_wap_expr(con, mode=APPEND, staging_strategy=StagingStrategy.BRANCH)`

 **factory-time validated**: APPEND-only, no key, backend must declare publish_branch). The alias stays because it's public API used by existing examples/tests; table_name now only selects the strategy (the branch's table is final at the pipe, which also fixes the latent table_name != final mismatch). New code should prefer make_backend_wap_expr; we can deprecate the alias later.

docs/adr/0017-incremental-wap-publish.md
ghoersti: `PublishStrategy.NATIVE_MERGE` was using it for the backends that support native merge. 
but `PublishStrategy.STATEMENT_DML` was just using `f-strings` when I can actually use `sqlglot` so I will update those backends to also use `sqlglot` ,  along with updating the ADR

python/xorq/backends/databricks/backend.py
```diff
@@ -123,6 +123,12 @@ class Backend(SQLBackend, CanCreateDatabase, UrlFromPath):
     name = "databricks"
     compiler = sc.databricks.compiler
 
+    def publish_strategy(self, mode):
+        """Incremental WAP publish mechanism (ADR-0017): Delta MERGE INTO."""
+        from xorq.writes.enums import PublishStrategy  # noqa: PLC0415
```
mesejo: Why is this import deferred?

python/xorq/backends/duckdb/__init__.py
```diff
@@ -16,6 +16,12 @@
 
 
 class Backend(IbisDuckDBBackend):
+    def publish_strategy(self, mode):
+        """Incremental WAP publish mechanism for this backend (ADR-0017)."""
```
mesejo: same here

python/xorq/backends/duckdb/__init__.py
```diff
@@ -16,6 +16,12 @@
 
 
 class Backend(IbisDuckDBBackend):
+    def publish_strategy(self, mode):
```
mesejo: Is the mode parameter used?

ghoersti: it is used in the publish and it was here as a way to protect against the for some reason a merge or upsert for some reason did not have the same capabilties. I think its better to remove it because IMO merge, upsert & append should behave the same on different backends.

python/xorq/backends/databricks/backend.py
ghoersti: This is ok not to be deferred, originally I was worried about cyclic imports but its ok.

ghoersti: Moved  import

PR #2139, opened 2026-07-05
refactor(transform): declarative pass driver for _transform_expr (R2)

dlovell:
## Summary

Turns the hand-run tier-1 transform sequence in `_transform_expr` into a **declarative, ordered pass table** so the correctness rules that previously lived only in prose comments become *data* the driver enforces. This is **R2** of the `to_pyarrow_batches` transform-robustness plan.

- **P2 killed** — traversal kind (`replace_nodes` descend vs `op.replace` boundary) is a `Traversal` field on each pass, applied by one driver (`apply_pass`). No call site can pick the wrong walk.
- **P3 killed** — pass ordering is an `after` dependency chain (`bind → tags → cache → tee → remote → deferred_reads`) asserted by `run_transform_passes`. Reordering the table now raises immediately instead of silently mis-transforming.
- Resource ownership: the shared `RemoteTableScope` is threaded to every pass via `TransformCtx` (the R1 mechanism); the tee and remote passes (flagged `produces_resources`) are the ones that adopt resources into it.

## What changed

- New leaf module `xorq/expr/transform.py`: `TransformPass` record, `apply_pass` (selects the walk from the record), `run_transform_passes` (folds + asserts `after`). `Traversal{DESCEND|BOUNDARY}` in new `xorq/expr/enums.py`.
- Each pass is a **replacer-builder** (`_make_*_replacer`) invoked by the driver. The public entry points external callers still use are kept: `bind_params` and `_remove_tag_nodes` (used by `ibis_yaml/compiler.py` and `to_sql`). Every per-pass adapter was removed — the old tuple-returning standalone wrappers first, then (in R2 review follow-up) the tee/remote `*_into` scope-threading adapters. All callers, tests included, now apply a single pass through the driver primitive `apply_pass(PASS, expr, TransformCtx(scope=...))`, so each pass has exactly one entry point (see Observability below).
- `replace_nodes` is imported lazily inside `apply_pass` to keep `transform` out of the `graph_utils → relations` import cycle.

## Design notes

- The plan's `kind={PURE|BOUNDARY}` is split into **two orthogonal axes**: `traversal` (the walk) and `produces_resources` (needs the scope). `deferred_reads` is pure yet must be BOUNDARY; `cache` is effectful yet owns nothing scope-tracked — one field couldn't express both.
- **R2 review follow-up:** `TransformCtx.scope` is optional (pure passes skip constructing a throwaway scope), `Replacer` is typed `Callable[[Node, dict|None], Node]`, and each imported pass echoes its `after` at the `_PASSES` assembly point so the whole ordering chain reads in one place (the driver still validates each record's real `after`).

## Observability (OTel) — behavioral change

The per-pass tracing moved from decorated per-function spans to one span per pass emitted by `apply_pass`. **This renames spans** — dashboards/alerts keyed on the old names must be updated:

- `register_and_transform_remote_tables` → `transform.remote`
- `register_and_transform_tee_nodes` → `transform.tee`
- `_register_and_transform_cache_tables` → `transform.cache`
- `_remove_tag_nodes` (on the transform path) → `transform.remove_tags`
- `_transform_deferred_reads` → `transform.deferred_reads`

On the `_transform_expr` path this is a net **increase** in coverage: all six passes now get a span, whereas `bind_params` previously had none (→ `transform.bind_params`). `_transform_expr` and `to_pyarrow_batches` spans are unchanged, and the `replace_read` event still attaches to the (now renamed) deferred-reads span.

Two now-orphaned decorated functions — `_register_and_transform_cache_tables` and `_transform_deferred_reads` — were deleted: once the driver invoked their builders directly they had zero callers, so their `@tracer` spans never fired anyway.

The aggregate `remote_table.replace` OTel event was also dropped (no consumers).

## Tests

- New `test_transform_driver.py` pins the driver's own guarantees: `run_transform_passes` raises `InternalError` on a violated `after` (P3 anchor), ordered / skipped-but-positioned cases, `apply_pass` DESCEND-vs-BOUNDARY selection (P2), and `when=False` skipping.
- Green: `test_transform_scope.py`, `test_write_through.py`, `test_stream_cache_contract.py`, `test_graph_utils.py`, datafusion `test_into_backend.py`.

## Stacking

Built on **#2138** (R1 — unified resource scope), now **merged**. This branch is rebased directly onto `main`, so the diff is just the R2 declarative-pass-driver work. Behaviour-preserving refactor otherwise (aside from the observability changes noted above).

Still open in the plan: **P4** (fuse the pure DESCEND passes; the `count_remote_table_readers` extra walk) and **R3/P5** (route the Flight early-return through the driver) — both intended as separate follow-ups.

### DIFF
--- a/python/xorq/expr/remote_table_exec.py
+++ b/python/xorq/expr/remote_table_exec.py
@@ -12,7 +12,7 @@ from xorq.common.compat import raise_collected_errors
 from xorq.common.utils.logging_utils import get_logger
 from xorq.expr.enums import Traversal
 from xorq.expr.relations import RemoteTable, gen_name
-from xorq.expr.transform import Replacer, TransformCtx, TransformPass, apply_pass
+from xorq.expr.transform import Replacer, TransformPass
 from xorq.vendor.ibis import Expr
 from xorq.vendor.ibis.backends import BaseBackend
 from xorq.vendor.ibis.expr import operations as ops
@@ -381,25 +381,6 @@ REMOTE_PASS = TransformPass(
     produces_resources=True,
     after=("tee",),
 )
-
-
-def register_and_transform_remote_tables_into(
-    expr: Expr, scope: RemoteTableScope, **kwargs: Any
-) -> Expr:
-    """Apply :data:`REMOTE_PASS` against the caller-owned ``scope``.
-
-    Thin adapter for callers that apply the remote pass on its own (tests,
-    one-off use): it runs the same :data:`REMOTE_PASS` record through the shared
-    driver that ``_transform_expr`` folds over its whole pass table, so both
-    apply identical traversal and scope discipline. The scope is threaded
-    explicitly by the caller; this never closes it -- ownership and teardown
-    stay with the caller that created it.
-    """
-    return apply_pass(
-        REMOTE_PASS,
-        expr,
-        TransformCtx(scope=scope, read_record_batches_kwargs=kwargs),
-    )
 
 
 def prepare_create_table_from_expr(
--- a/python/xorq/expr/remote_table_exec.py
+++ b/python/xorq/expr/remote_table_exec.py
@@ -7,12 +7,12 @@ from typing import Any, Callable, NamedTuple
 
 import pyarrow as pa
 from batchcorder import StreamCache
-from opentelemetry import trace
 
 from xorq.common.compat import raise_collected_errors
 from xorq.common.utils.logging_utils import get_logger
-from xorq.common.utils.otel_utils import tracer
+from xorq.expr.enums import Traversal
 from xorq.expr.relations import RemoteTable, gen_name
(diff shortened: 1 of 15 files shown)


python/xorq/expr/remote_table_exec.py
```diff
@@ -360,24 +354,33 @@ def replacer(node, kwargs):
...
             node = node.__recreate__(kwargs)
 
         return node
 
-    # Intentionally op.replace, not replace_nodes: mark_remote_table has side effects
-    # that must not descend into opaque sub-exprs (e.g. ExprScalarUDF.computed_kwargs_expr).
-    # The caller owns scope teardown, so a failure here just propagates.
-    expr = op.replace(replacer).to_expr()
-    if scope.table_count:
-        trace.get_current_span().add_event(
```
mesejo, line 377: I think this is missing from the new implementation

mesejo (approved):

Issue #2066, opened 2026-06-12
Standardize `__init__.py` conventions across the codebase


dlovell:
## Problem

The repo's `__init__.py` files follow several divergent patterns with no clear governing convention. This makes the public API ambiguous, hinders IDE tooling, and creates inconsistency across sibling packages.

### 1. Missing `__all__` in large implementation modules

8 files contain 200–980 lines of public classes/functions but never declare `__all__`:

- `backends/xorq_datafusion/__init__.py` (980 lines)
- `backends/databricks/__init__.py` (670 lines)
- `backends/pandas/__init__.py` (389 lines)
- `backends/pyiceberg/__init__.py` (319 lines)
- `backends/snowflake/__init__.py` (317 lines)
- `expr/builders/__init__.py` (282 lines)
- `sinking/__init__.py` (212 lines)
- `backends/datafusion/__init__.py` (209 lines)

Meanwhile, `caching/__init__.py`, `flight/__init__.py`, `expr/ml/__init__.py`, `vendor/ibis/__init__.py`, and `common/utils/dasher/__init__.py` *do* declare `__all__`. There is no discernible rule for which modules get it and which don't.

**Impact:** Without `__all__`, `from module import *` exports every name including internal helpers (e.g. `_select_and_cast`, `_databricks_type_to_ibis`). IDE autocompletion cannot distinguish public from private API.

### 2. Overloaded `__init__.py` files

`backends/xorq_datafusion/__init__.py` (980 lines) and `backends/databricks/__init__.py` (670 lines) contain full backend implementations — SQL compilation, UDF registration, type converters, connection management — rather than re-exporting from submodules.

Conventional expectation is that `__init__.py` is either empty, re-exports from submodules, or contains minimal glue. Putting the entire implementation here makes the package harder to navigate since the directory listing gives no indication of where logic lives.

### 3. Inconsistent `@public` decorator

`caching/__init__.py` uses a `@public` decorator to mark exported symbols. No other module does this. It's either a useful convention that should be adopted broadly, or an outlier that should be removed for consistency.

### 4. Mixed lazy/eager import strategy

Backend modules vary without a documented rule:
- `backends/pandas/__init__.py` eagerly imports at module level
- `backends/databricks/__init__.py` lazy-imports inside methods
- Most backends use a mix of both

The *intent* seems to be "lazy-import optional dependencies," which is sound, but it isn't codified or consistent.

## Proposed Conventions

[...]

PR #2140, opened 2026-07-06
refactor(backends): standardize __all__, drop @public (#2066 phase 3)

mesejo:
## Summary

Phase 3 of #2066 — standardize decorator & import conventions.

First-party code used the `@public` decorator in only two modules
(`caching/__init__.py`, `backends/pandas/rewrites.py`); the rest of the package
declares its public surface with an explicit `__all__`. `@public` mutates
`__all__` at runtime, hiding the public API from static tooling and forcing
`# noqa: PLE0604`. This resolves the inconsistency in favor of explicit
`__all__`.

## Changes

- **`caching/__init__.py`** — drop `@public` (7) + `from public import public`,
  replace with explicit `__all__`. **Behavior-preserving**: `@public` had been
  adding `ParquetDummySnapshotCache` to the effective public set (which flows to
  `xorq.api` via `*caching.__all__`), so it is now listed explicitly.
- **`backends/pandas/rewrites.py`** — drop `@public` from the 9 op classes, add
  explicit `__all__`.
- **`CONTRIBUTING.md`** — new "Module and import conventions" section:
  - `__all__` is the single source of truth for the public API; no `@public` in
    first-party code.
  - eager-vs-lazy import rule (optional deps / heavy imports / cycle-breaking),
    enforced by ruff `PLC0415`.

`@public` remains only in vendored ibis (`python/xorq/vendor/`), which follows
upstream conventions and is exempt.

## Verification

- `ruff check` + `ruff format` clean (pre-commit hooks passed).
- Public surface preserved: `caching.__all__` = all 7 classes; `xorq.api`
  exposes all 7.
- Tests: 687 passed (pandas backend), 12 passed (cache).

Refs #2066

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,43 @@ just up postgres # some of the tests use postgres
 python -m pytest # or pytest
 ```
 
+## Module and import conventions
+
+These rules keep the public API explicit and import time predictable across the
+package. They apply to first-party code under `python/xorq/`; vendored code
+under `python/xorq/vendor/` follows upstream conventions and is exempt.
+
+### Declaring the public API with `__all__`
+
+- Every `__init__.py` (and any module meant to be imported with `*`) declares
+  an explicit `__all__` listing the names it exports. This is the single source
+  of truth for the public surface and what IDEs and doc tooling read.
+- Do **not** use the `@public` decorator in first-party code. It mutates
+  `__all__` at runtime, which hides the public surface from static tools and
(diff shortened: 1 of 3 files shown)


dlovell (approved):
Clean, behavior-preserving refactor. Verified the `public()` decorator only mutates `__all__`, both explicit lists exactly match the previously `@public`-decorated classes, and CONTRIBUTING.md claims (PLC0415 enforced, no remaining first-party @public) are accurate.

Issue #1580, opened 2026-02-13
CI hang: correlated subqueries with UDAFs deadlock on low-core runners


ghoersti:
## Summary

The `sklearn_deferred_auc_from_curve` example hangs indefinitely during `execute()` on GitHub Actions CI (2-core `ubuntu-latest` runner). It passes on machines with more cores (both macOS and Linux, Python 3.10 and 3.11).

## Reproduction

The hang occurs when composing multiple metric subqueries via `.as_scalar().mutate()` without caching the intermediate predictions:

```python
proba_expr = fitted.predict_proba(test_expr, name="scores")

# 4 correlated subqueries, each re-executing the full predict_proba pipeline
deferred_roc = deferred_sklearn_metric(expr=proba_expr, ..., metric=roc_curve)
deferred_pr = deferred_sklearn_metric(expr=proba_expr, ..., metric=precision_recall_curve)
deferred_det = deferred_sklearn_metric(expr=proba_expr, ..., metric=det_curve)
deferred_roc_auc_score = deferred_sklearn_metric(expr=proba_expr, ..., metric=roc_auc_score)

# Each .as_scalar() creates a correlated subquery containing a UDAF
auc_metrics = (
    deferred_auc_from_curve(deferred_roc).as_scalar().name("roc_auc").as_table()
    .mutate(pr_auc=deferred_auc_from_curve(deferred_pr).as_scalar())
    .mutate(det_auc=deferred_auc_from_curve(deferred_det).as_scalar())
    .mutate(roc_auc_score=deferred_roc_auc_score.as_scalar())
)

auc_metrics.execute()  # <-- hangs on 2-core CI, works on higher core counts
```

## CI log evidence

```
before execute main
##[error]The operation was canceled.    <-- 8 minutes later, no output
```

All imports succeed. The hang is during query execution only.

## Environment

- **Hangs on**: `ubuntu-latest` (2 vCPU, 7 GB RAM)
- **Works on**: machines with more cores (Linux Python 3.11, macOS Python 3.10)
- **Nondeterministic**: identical code passed once and hung on subsequent runs on the same commit

## Analysis

The expression tree contains:
- 1 `RemoteTable` (with nested `Tag` + sklearn UDFs)
- 4 `ScalarSubquery` nodes, each containing a UDAF (pandas_df aggregate)
- pyarrow scalar UDFs (`deferred_auc_from_curve`)

[...]

PR #2141, opened 2026-07-06
chore(deps): bump xorq-datafusion to 0.2.10

mesejo:
## Summary
- Bump `xorq-datafusion` pin `0.2.9` → `0.2.10` (`pyproject.toml`, `uv.lock`)
- Keep same-backend DataFusion exprs compiling to a **native DataFrame** in `_resolve_expr_for_register`; only genuine cross-backend exprs register via `IbisTableProvider`. (An earlier revision of this PR routed same-backend exprs through the provider — see the note below on why that was reverted.)
- Add tests covering the `register()` dispatch arms that lacked coverage, plus a core-pinned regression test for the reentrant deadlock.

## ⚠️ Note: reverted the same-backend → `IbisTableProvider` routing (issue #1580)
An earlier revision of this PR dropped the same-backend→native-DataFrame path and routed same-backend exprs through `IbisTableProvider`, on the assumption that 0.2.10's runtime-within-runtime made `scan()` re-entering the same connection safe.

Testing showed that holds for a **single** level but not for **reentrant** exprs. `IbisTableProvider.scan()` re-enters the source connection **synchronously**, blocking the tokio worker it runs on; a ≥3-level reentrant same-backend expr then blocks workers on inner scans that themselves need workers, starving the pool and **deadlocking on core-limited (2-core CI) runners** — the exact #1580 pattern. Reproduced deterministically: pinned to 2 cores, reentrant depth ≥3 hangs (3/3); depth 1–2 and any depth at ≥48 cores pass; `main`'s native-DataFrame path never hangs at any depth. It only surfaces on 2-core runners, so it would slip through local/high-core CI.

The fix here effectively **reverts the "route same-backend exprs through `IbisTableProvider`" decision** (keeping the phase-1 signature cleanup). If that routing was intended to solve a specific problem beyond the coverage cleanup, **that goal is not addressed** — the provider path simply is not safe for same-connection reentrancy, and same-backend exprs should stay on native materialization.

## Coverage
`backend.py` improved **87.56% → ~89%** (54 → 46 missed lines locally).

New tests exercise the previously-uncovered `register()` dispatch arms:
- native DataFusion `DataFrame` arm (also the target of the same-backend materialization path)
- pyarrow `Dataset` arm
- bound non-table column `ir.Expr` arm
- unknown-source-type `ValueError`
- `IbisTableProvider.scan(filters)` via predicate push-down (cross-backend)

Plus a core-pinned, timeout-guarded regression test (`test_reentrant_same_backend_register_no_deadlock_on_two_cores`) that reproduces the #1580 shape on two cores and asserts it now completes.

Remaining uncovered lines are structural, not gaps:
- `read_delta` — requires the `deltalake` extra (absent in the test env)
- native `Table()` arm — `xorq.internal.Table` is FFI-only, no public constructor
- `provider.py` scan body — runs on a DataFusion worker thread that `coverage.py` does not trace (proven to execute via a scan trace)
- scattered validation/error branches

## Testing
- `xorq_datafusion` + `datafusion` backend suites: **all passing** (`test_register.py`: 32 passed)
- `test_into_backend` (core, non-postgres/trino): **80 passed**
- Reentrant deadlock repro/regression: verified 2-core hang on the provider route and completion on the native-DataFrame path (depths 3–6)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "typing-extensions>=4.3.0",
     "py-yaml12",
     "cloudpickle>=3.1.1",
-    "xorq-datafusion==0.2.9",
+    "xorq-datafusion==0.2.10",
     "opentelemetry-sdk>=1.32.1",
     "opentelemetry-exporter-otlp>=1.32.1",
     "opentelemetry-exporter-otlp-proto-grpc>=1.32.1",
--- a/python/xorq/backends/xorq_datafusion/backend.py
+++ b/python/xorq/backends/xorq_datafusion/backend.py
@@ -487,7 +487,9 @@ class Backend(SQLBackend, CanCreateCatalog, CanCreateDatabase, CanCreateSchema, 
         table_name
             The name of the table
         kwargs
-            Datafusion-specific keyword arguments
+            DataFusion-specific keyword arguments. Only applied to path and
+            record-batch-reader sources; ignored for expr, DataFrame, and
+            in-memory table sources.
 
         """
         import pandas as pd  # noqa: PLC0415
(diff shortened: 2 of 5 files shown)


dlovell:
we claim re-entrancy is possible, should we explicitly test that here?

dlovell (approved):

PR #2158, opened 2026-07-13
feat(bigquery): add lightweight xorq bigquery backend

mesejo:
## Summary

Adds a lightweight BigQuery backend to xorq, plus the Arrow ingestion path, cache-key support, tests, and CI needed to make it usable end-to-end.

### Backend
- **`xorq.backends.bigquery`** — a thin `Backend` subclassing the vendored Ibis BigQuery backend, registered as the `bigquery` entry point and exposed as `xo.bigquery`. Compilation works offline (no credentials/connection needed).
- **`read_record_batches`** — ingests a `pa.RecordBatchReader`/`pa.Table` into a BigQuery table, landing it in the connection's dataset. Implemented over the BigQuery **ADBC driver** (`BigQueryADBC`), mirroring the snowflake/databricks backends: `adbc-driver-manager` connects with `driver="bigquery"` and calls `adbc_ingest`, reusing the backend's credentials (user auth when available, else Application Default Credentials).
  - Bulk ingest is only in the BigQuery ADBC driver ≥ 1.12 (the PyPI `adbc-driver-bigquery` 1.11.0 has none), so the driver is installed out-of-band via `dbc install bigquery` rather than as a Python dependency.

### Cache-key normalizer (dasher)
- **`_normalize_bigquery_databasetable_xorq`** — the vendored `xorq_dasher` bigquery normalizer unpacks its result frame by column label (`((x,),) = df.to_dataframe()`) and raises `ValueError: too many values to unpack` for every BigQuery table. Added a repo-local normalizer (same pattern as the existing duckdb/datafusion overrides) that reads `last_modified_time` out of the frame correctly, casts the numpy scalar to a native `int`, and catalog-qualifies `__TABLES__`; bigquery is dispatched to it. This makes `tokenize`/caching work for BigQuery-backed expressions.

### Tests
- `test_backend.py` — offline: registration, API surface, compile (project/aggregate/join), `has_operation`.
- `test_client.py` — live: connect, `raw_sql`, dataset listing, table create/drop, create-from-expr, `get_schema`.
- `test_basic.py` — live: filter/aggregate/join, collect variants (`to_pyarrow`/`to_pyarrow_batches`/`execute`), `read_record_batches`, `sql`, `train_test_split`, `into_backend` (BigQuery → DuckDB) + `tokenize`.
- `conftest.py` — session fixtures for credentials/project/dataset (resolves project from `GOOGLE_BIGQUERY_PROJECT_ID` or the ADC/service-account default), `batting`/`awards_players`. Live tests skip cleanly without credentials.
- `backends/conftest.py` — skip bigquery test collection when the `bigquery` extra is absent.

### Packaging & CI
- `pyproject.toml` — `bigquery` extra (`google-cloud-bigquery`, `-storage`, `db-dtypes`, `pydata-google-auth`, `adbc-driver-manager`) and the `bigquery` backend entry point.
- `.github/workflows/ci-test-bigquery.yml` — installs the project + `dbc install bigquery` for the ADBC driver (mirroring the databricks workflow), authenticates with the `GCP_SA_KEY` service account, and runs `-m bigquery`. Project is taken from the `GOOGLE_BIGQUERY_PROJECT_ID` variable, falling back to the service-account key's embedded project.

## Notes
- Live tests require GCP credentials (service account in CI, ADC locally) and run only under `-m bigquery`; offline tests run everywhere.
- The BigQuery ADBC driver is not a PyPI dependency by design — it ships via the `dbc` channel (v1.12.1, which supports bulk ingest).


ghoersti:
Tested `bcb210b` end-to-end on macOS (ADC auth, dataset create/drop, `SELECT 1`, `read_record_batches` ingest + read-back) against a real project — works. `dbc install bigquery` (Foundry driver 1.12.1) is required and correct.

One UX note: the PyPI `adbc-driver-bigquery` wheel (1.11.0, old apache/arrow-adbc lineage) is a trap. It resolves and loads, but its `SetOption` has no `adbc.ingest.*` support, so pip-installers get `unknown statement string type option 'adbc.ingest.target_table'` at ingest time with no pointer back to `dbc`. Suggest catching that error in `adbc_ingest` (or extending the `get_conn` message) to say the PyPI wheel can't bulk-ingest and `dbc` is the supported install path. @mesejo

ghoersti:
@mesejo   You can replicate the issue from the pip install here. 

Minimal repro of the PyPI-wheel trap — no credentials/network needed, the stale driver rejects the ingest option locally in its own option parser (the first call `cursor.adbc_ingest()` makes):

```python
# pip install adbc-driver-bigquery adbc-driver-manager
from importlib.resources import files
from adbc_driver_manager import _lib

db = _lib.AdbcDatabase(
    driver=str(files("adbc_driver_bigquery") / "libadbc_driver_bigquery.so"),
    **{"adbc.bigquery.sql.project_id": "demo"},
)
stmt = _lib.AdbcStatement(_lib.AdbcConnection(db))
stmt.set_options(**{"adbc.ingest.target_table": "t"})
# ProgrammingError: INVALID_ARGUMENT: unknown statement string type option `adbc.ingest.target_table`
```

Binary-level confirmation the feature was never compiled in:

```bash
$ strings -a .../site-packages/adbc_driver_bigquery/libadbc_driver_bigquery.so | grep -c adbc.ingest
0   # PyPI wheel 1.11.0 (apache lineage)
$ strings -a ~/Library/Application\ Support/ADBC/Drivers/bigquery_*/libadbc_driver_bigquery.dylib | grep -c adbc.ingest
5   # dbc-installed Foundry 1.12.1
```

ghoersti (changes requested):
Need to address about pip install for driver. 
 https://github.com/xorq-labs/xorq/pull/2158#issuecomment-5024220970

ghoersti (approved):

PR #2182, opened 2026-07-24
feat(backends): mixpanel backend with profile-carried auth; backend-declared secret keys

dlovell:
## Summary

Phase 1 of API-as-Backend (`plans/udxf-source-api-backend.md`): make a REST API (Mixpanel) look like a xorq Backend whose auth/config is a saved Profile.

- **Backend-declared secret keys**: `check_for_exposed_secrets` now consults `Backend._secret_keys` (entry-point load via new `get_declared_secret_keys`), falling back to the legacy dict when the backend module is unimportable; postgres and snowflake declare their keys.
- **New read-only `mixpanel` backend**: `events`/`engage` resources exposed as deferred `flight_udxf` relations. `MixpanelClient` fields hold env var references that resolve per request — expressions and build artifacts never capture credential values. Raw secrets are rejected at `Profile.save` **and** at expression construction (`_expr_client`), since cloudpickled closures in `expr.yaml` are base64-encoded and not auditable by grep.
- **ADR-0016** (lands with this implementation): build artifacts are credential-free; Profiles are the sole credential carrier.
- `page_size` passthrough for engage pagination (server clamps to >= 100), which made multi-page pagination testable.

## Verification

- 11 unit tests (no network, fake env creds) + 1 live test (skipif without creds).
- Live against a real Mixpanel project: profile save→load→reconnect round-trip; `read_events` → 350 rows; `xorq build` leak-grep clean with fake creds in env; `xorq run` of the artifact re-fetches; engage multi-page (120 profiles, page_size=100 → 2 pages, session_id continuation).
- Untested: the 429/backoff branch (needs a genuine rate-limit hit).

## Known limitation (Phase 2 scope)

The mixpanel profile rides inside the pickled exchanger rather than the build's `profiles.yaml`; Phase 2's `Read`-op integration makes the backend/profile a first-class build participant.

### DIFF
--- a/docs/adr/0016-build-artifacts-are-credential-free.md
+++ b/docs/adr/0016-build-artifacts-are-credential-free.md
@@ -1,0 +1,141 @@
+# ADR-0016: Build artifacts are credential-free; Profiles are the sole credential carrier
+
+- **Status:** Accepted
+- **Date:** 2026-07-23
+- **Deciders:** Dan Lovell
+
+## Context
+
+A xorq expression that reads from an authenticated source (postgres, snowflake,
+and now REST APIs like Mixpanel) must be serializable into a build artifact
+that is shareable, cacheable, and cataloged by content hash. Credentials create
+two hazards there:
+
+1. **Leakage.** Anything a build captures — `profiles.yaml`, SQL source YAML,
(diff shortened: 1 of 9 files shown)


dlovell:
Withdrawn.

PR #2185, opened 2026-07-28
feat(lineage): compact, boundary-aware lineage in expr metadata (XOR-363)

mesejo:
> **Stacked on #2177.** Based on `main`, so the diff includes #2177's four `graph_utils` refactor commits (already under review there — `d160a240`, `e1e31381`, `73fab4bb`, `d64e247c`). Review and merge #2177 first; this diff shrinks to the lineage work once it lands. The only `graph_utils` change *owned by this PR* is the `gen_children_flight_leaf` visibility flip described below.

## What

Enrich the sidecar `LineageDAG` with a semantic boundary taxonomy, and make the compact and nested views **derived** rather than stored.

## Why

XOR-363. The sidecar recorded a flat node list with `" -> ".join(labels)` rendering: no notion of which nodes are semantically meaningful, no way to expand or collapse, and Flight `input_expr` nodes flattened into the outer graph so they appeared twice with wrong edges.

## How

- **Boundary taxonomy** via singledispatch — `_boundary_kind` (`ingestion`, `cache`, `pin`, `engine_crossing`, `flight_udxf`, `flight_expr`, `udf`, `table`, `unbound`, `join`, `tag`/`tag:<value>`) and `_boundary_extras` for per-kind intrinsic facts. The dispatch registry is the single source of truth; a test asserts every `opaque_ops` type registers. `base_kind()` strips a tag's qualifier so kind queries and registry lookups match either form.
- **Node identity from `content_hash`** for relations, so a relation keys identically in `expr.yaml` and the sidecar (`snapshot_hash`); `@op_N` stays the doc-local readable label. Value ops (`Field`, `SortKey`, `Literal`, …) use a memoized structural token instead — `content_hash` snapshot-normalizes and compiles to SQL, which fails outright on `SortKey` and is wasted work for nodes that never reach `expr.yaml`.
- **Flight-leaf outer walk** — `bfs(root, children=gen_children_flight_leaf)`, the policy landed in #2177. A Flight node's `input_expr` is no longer flattened into the outer graph; it is recursed as a *nested* sub-DAG whose nodes live in the one shared node table under a `scope: <flight label>` tag. BFS dedup guards nested-of-nested recursion.
- Structured `{from, to, scope}` edges, `version: 2`, an `overlays` scaffold for future dimensions (columns, engines), and a code-side `CAPABILITY_REGISTRY` keyed by `boundary_kind`.
- **Query API** on `LineageDAG`: `resolve` (hash | label | tag | kind | predicate), `boundaries`, `capabilities`, `available`, `expand`, `scope`, `compact`, `by_id`.
- `compact()` is an adjacency BFS over the **id-graph** (stored labels + edges, not the op-graph — at render time there are no ops), collapsing non-boundary runs into `via: [types]` and keeping one edge per boundary pair (the most direct run).
- `from_dict` tolerates legacy sidecars (integer ids, 2-tuple edges, no version).
- **TUI**: `lineage_text` renders the compact tree instead of `" -> ".join(labels)`. `FlightUDXF` shows as `UDXF[class] : N->M cols` with its schema delta and nested input sub-tree; `FlightExpr`'s label no longer stringifies `input_expr`.
- **TUI panel** (`25adf622`): the Info panel becomes a `VerticalScroll` and joins `FOCUS_CYCLE` -- the old single-line lineage fit in `max-height: 6`, a tree of unbounded depth does not, and a plain `Vertical` outside the focus cycle had no scrollbar, no keys and no wheel target. Same treatment `#sql-panel` already has; `j`/`k` needed no change because `action_cursor_down`/`_up` already dispatch to a focused `VerticalScroll`.
- **TUI colours** (`802fdc81`): `BOUNDARY_STYLES` gives each base kind an icon and a theme colour, loudest for the Flight boundaries (the only hop that leaves the process); tree glyphs dim, collapsed `via [...]` runs dim italic.

```
  Join(2 inputs) (6 cols) [xorq_datafusion]
  |-- Cache[ParquetCache] (4 cols) [xorq_datafusion]   via [Field, JoinReference]
  |   `-- -> xorq_datafusion (4 cols)
  |       `-- UDXF[OrderEnricher] : 3->4 cols [xorq_datafusion]   (+order_score)
  |           `-- xorq_datafusion:raw_orders (3 cols)     <- nested input scope
  `-- RemoteTable[duckdb -> xorq_datafusion] (3 cols)   via [Field, JoinReference]
      `-- duckdb:raw_customers (3 cols)
```

  Styling reads `LineageRow.kind`, never the rendered label: `Cache[...]`, `UDXF[...]` and `via [...]` all contain brackets, so Rich markup would misparse them. `lineage_rich`/`info_rich` are the renders and `lineage_text`/`info_text` are their `.plain`, so the two forms cannot drift. Icons live in the renderer, so `expr_metadata.json` is untouched by this.
- **Label widths + backend tag** (`7e07cd41`): every boundary label is now `<what> (<n> cols) [<backend>]`, matching the ticket's target render. Pieces drop out when the node cannot supply them (a legacy sidecar with no schema renders bare); Flight kinds keep their schema *transition* instead of one width; the tag is appended only for kinds whose label does not already name a backend, so `duckdb:orders`, `Read[duckdb]` and `RemoteTable[duckdb -> xorq]` are untouched. Two extras were needed: `backend` on the Flight nodes (they have a `source` and never reported it) and `backend` on `JoinChain`, which has none -- a join's is its *left input's* (`first.parent`), because walking to the leaves reaches the remote side of an `into_backend` and would report a single-backend join as multi-backend.
- **`LineageRow`** (`fed211c2`): `compact_lineage_rows(dag)` returns the compact tree with the glyph prefix, label, boundary kind and `via` kept apart, which is what makes the styling above possible without label parsing. `TextTree._lines` is derived from the same rows; the plain-text output is unchanged.
- **`scope()` fix** (`fed211c2`): the view derived its id set from the scope's edges alone, so a Flight node whose `input_expr` is a bare table (one node, zero edges) came back with `nodes: ()` while `root` still pointed at that node. Found by exercising the sidecar-inspection recipes on a real build; the existing tests missed it because their UDXF input was a filter over an `into_backend`.
- One `graph_utils` change: `_gen_children_flight_leaf` → `gen_children_flight_leaf`. #2177 left it private with a "private until XOR-363 lands its caller" note; this branch is that caller.

## Per-kind boundary facts (`31013717`)

Every field in the ticket's per-kind table is stored:

- **tags** — `tag:<tag-value>` for `HashingTag`, full sanitized `tag_metadata`.
- **RemoteTable** — `remote_name`, `source_backend`, resolved with `find_all_sources(remote_expr, execution_only=True)` rather than off the node: the remote expr's root is usually a `Project`/`Filter` with no `source` of its own. A multi-source or memtable-rooted input has no single origin, so the field is omitted rather than guessed.
- **Read** — `format`, `path_or_uri`, `read_kwargs`, with credential-looking keys redacted to `***` and URI userinfo collapsed to `scheme://***@host`. The sidecar is committed next to the build, so a `postgres://user:pw@host/db` must not travel with it.
- **DatabaseTable / UnboundTable** — `table_name`, `database`, `catalog`.
- **JoinChain** — `n_inputs`, `predicates_summary` (column refs and operators only, never a literal's value).
- **ExprScalarUDF** — `udf_name`, `udf_signature`.
- **FlightUDXF** — `udxf_class`, `udxf_command`, `schema_in_required`, `schema_in_observed`, `schema_out`, `process_boundary`, `server_factory`, `do_instrument_reader`, and a stored `schema_diff` (ticket open question #3, resolved in favour of the sidecar so external readers get the transition for free).
- **FlightExpr** — `flight_command`, `input_schema`, `output_schema`, `server_factory`. The command is built from `UnboundExprExchanger`: `FlightExpr.to_rbr` registers one and exchanges on *its* `command`, so the exchanger is the only authority and recomputing the format string here would rot silently. Construction is pure (walk + rename + tokenize — no server, no connection) and a failure degrades to no field.
- **CachedNode / CacheTag** — `cache_kind`, `cache_key_prefix`, `cache_key` (pins), and `cache_path` as the storage's **relative** path only. The resolved directory is `base_path / relative_path` with `base_path` defaulting to the user's cache dir — an absolute per-machine path that must not enter a committed sidecar. The per-entry parquet path is also excluded: it needs `cache.calc_key(expr)`, which stats the source data under `ModificationTimeStrategy`. `SourceCache` has no path at all; its location is the `backend`.

No `server_kind` flavour is inferred: the `FlightExpr` default factory is the bare `FlightServer` class and the `FlightUDXF` default is a `make_mtls_server` closure, so name-sniffing yields a class name as often as a flavour. `server_factory` alone carries it.

## Deviations from the ticket text

- **No `BOUNDARY_TYPES` constant.** The `_boundary_kind` dispatch registry is the single source of truth; a parallel tuple would drift. A test asserts every `opaque_ops` type has a registration.
- **`read_options` is stored as `read_kwargs`** — the op's own attribute name.
- **Nested lineage is not a per-node `nested_input_lineage` sub-document.** One shared node table (deduped by `snapshot_hash`) plus `scope`-tagged edges and a `nested_root` pointer; `scope(label)` returns the nested view by filtering. This keeps global hash-addressing and lets a future overlay span a Flight boundary.
- **`via` lives on the derived `compact()` edges, not on a stored `compact_edges`** (ticket open question #1). Nothing about the compact view is serialized.
- **TUI is a plain-text tree** — no icons or colour; `Cache[...]`, `UDXF[...]`, `RemoteTable[a -> b]`, `↳`, `↻` carry the signal.

## Build hash

Build-stability snapshots move **only** in `expr_metadata.json`. `build_dir_name` and `expr.yaml` are unchanged, so the richer sidecar does not affect the build hash.

## Verification

`test_lineage_utils.py` + `test_graph_utils.py` + `test_content_hash.py` + `test_tui.py` + `test_compiler.py` — **526 passed, 1 skipped**; the full `test_tui.py` again after the Info-panel change — **355 passed**; `test_lineage_utils.py` + `test_tui.py` + `test_compiler.py` + `test_graph_utils.py` after the colour and `scope()` work — **530 passed, 1 skipped** with no snapshot movement; and again after the label widths — **543 passed, 1 skipped**, where two `expr_metadata.json` hashes move (the new `backend` field) and nothing else. The three `expected.json` stability snapshots move as described above (verified field-by-field against the failure diff before regenerating).

Coverage added for every new field, plus the behaviours that are easy to get wrong: kwarg/URI redaction, `schema_diff` add/drop/retype/unambiguous-rename/ambiguous/no-delta cases, `flight_command` compared against the exchanger itself rather than a hard-coded string, no resolved cache dir or `$HOME` in a cache node, `source_backend` present for duckdb → xorq and omitted when unresolvable, and a catalog entry containing a `FlightUDXF` rendering `UDXF[AddB] : 1->2 cols` with `(+b)` and its nested input source. Further tests assert the Info panel is a `VerticalScroll` that `tab` lands on and `j`/`k` scroll; that `LineageRow`s concatenate back to exactly the plain tree and keep `via` off the label; that a single-node nested scope keeps its root; that each boundary kind renders with its own icon and colour while a legacy sidecar falls back to a neutral one; that labels carry the width and the backend exactly once; and that a join's backend is its left input's rather than the remote side of an `into_backend`.

## Scope

Draft. Third and last link of the XOR-363 chain: Prep #1 `content_hash` → Prep #2 graph refactor (#2177) → **this**.

Refs XOR-363

### DIFF
--- a/python/xorq/catalog/tests/test_tui.py
+++ b/python/xorq/catalog/tests/test_tui.py
@@ -38,9 +38,11 @@ from xorq.catalog.tests.testing import (
     wait_until,
 )
 from xorq.catalog.tui import (
+    BOUNDARY_STYLES,
     GIT_LOG_COLUMNS,
     KIND_ORDER,
     KIND_STYLES,
+    UNKNOWN_BOUNDARY_STYLE,
     AddAliasScreen,
     AddEntryScreen,
     CatalogRowData,
@@ -61,6 +63,7 @@ from xorq.catalog.tui import (
     _list_revisions_cached,
     _pygments_to_text,
     _pygments_tokens,
+    _render_lineage_rows,
     _render_sql_dag,
     _render_sql_text,
     _styled_branch_label,
@@ -72,6 +75,7 @@ from xorq.common.utils.env_utils import (
     EnvConfigable,
     env_templates_dir,
 )
+from xorq.common.utils.lineage_utils import LineageDAG, compact_lineage_rows
 from xorq.config import TUI, options
 from xorq.ibis_yaml.enums import ExprKind
 
@@ -254,6 +258,55 @@ def test_lineage_text_renders_compact_boundary_tree(entry_a):
 def test_lineage_text_shows_cache_boundary(entry_cached: CatalogEntry) -> None:
     row = CatalogRowData(entry=entry_cached)
     assert "Cache[" in row.lineage_text, row.lineage_text
+
+
+def test_lineage_rich_styles_each_boundary_kind(entry_udxf: CatalogEntry) -> None:
+    """Boundary kinds get an icon and a colour; the tree glyphs and the collapsed
+    `via` runs stay dim.  `lineage_text` is the same render's plain text."""
+    row = CatalogRowData(entry=entry_udxf)
+    rich = row.lineage_rich
+
+    assert row.lineage_text == rich.plain
+    assert BOUNDARY_STYLES["flight_udxf"].icon in rich.plain
+    assert BOUNDARY_STYLES["table"].icon in rich.plain
+
+    styles = {str(span.style) for span in rich.spans}
+    assert any(BOUNDARY_STYLES["flight_udxf"].color in s for s in styles), styles
+    assert any("dim" in s for s in styles), styles
+
+    # a kind we did not style would render with the unknown-boundary icon
+    assert UNKNOWN_BOUNDARY_STYLE.icon not in rich.plain
+
+
+def test_info_rich_labels_and_indents_the_lineage(entry_udxf: CatalogEntry) -> None:
+    row = CatalogRowData(entry=entry_udxf)
+    lines = row.info_rich.plain.splitlines()
+
+    assert lines[0] == "Lineage:"
+    assert all(line.startswith("  ") for line in lines[1:-2])
+    assert lines[-2].startswith("Cache: ")
+    assert lines[-1].startswith("Hash: ")
+    assert row.info_text == row.info_rich.plain
+
+
+def test_lineage_rich_of_a_legacy_sidecar_uses_the_unknown_style() -> None:
+    """A pre-boundary sidecar annotates nothing, so every row falls back to the
+    unknown-boundary style instead of raising or missing a style lookup."""
+    legacy = LineageDAG.from_dict(
+        {
+            "nodes": [
+                {"id": "0", "type": "Filter", "label": "Filter"},
+                {"id": "1", "type": "InMemoryTable", "label": "InMemoryTable"},
+            ],
+            "edges": [["0", "1"]],
+            "root": "0",
+        }
+    )
+
+    rendered = _render_lineage_rows(compact_lineage_rows(legacy))
+
+    assert rendered.plain == f"{UNKNOWN_BOUNDARY_STYLE.icon} Filter"
+    assert any(UNKNOWN_BOUNDARY_STYLE.color in str(s.style) for s in rendered.spans)
 
 
 def test_lineage_text_renders_udxf_identity_and_nested_input(
--- a/python/xorq/catalog/tests/test_tui.py
+++ b/python/xorq/catalog/tests/test_tui.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from unittest.mock import patch
 
 import pytest
+from textual.containers import VerticalScroll
 from textual.widgets import DataTable, Input, Select, Static, Tree
 
 import xorq.api as xo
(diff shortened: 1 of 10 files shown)


dlovell (approved):
Approved by repository owner (relayed via Claude Code session).

PR #2188, opened 2026-07-29
feat(catalog): add `xorq catalog lineage` (--level, --node, --format)

mesejo:
> **Stacked on #2185** (`mesejo/xor-363-compact-semantically-meaningful-lineage-in-expr-metadata`), which is itself stacked on #2177. Review and merge those first; this diff is one command, renderer additions to `lineage_utils`, a TUI rework of the right panel, their tests, and the docs-nav entry.

## What

`xorq catalog lineage <entry> [--level compact|boundaries|raw] [--node HANDLE] [--format text|mermaid] [--expand HANDLE]`.

Four independent dials: `--level` how much detail, `--node` how much of the graph, `--format` the rendering, `--expand` one node opened up to its columns.

The TUI keeps pace: the right panel becomes a 3-way Lineage/SQL/Data switch, and the Lineage panel gains a row cursor with the same column expansion the CLI's `--expand` does.

## Why

XOR-363 gave the sidecar a boundary-annotated `LineageDAG` and the TUI a compact tree to render it, but left no way to read either outside the TUI. Inspecting a build meant hand-writing `jq` over `expr_metadata.json` or a `python -c` against `LineageDAG`. The ticket never asked for a CLI -- this is the follow-up.

## Levels

Every level is **read or derived from the stored sidecar**. None re-walks the expression or loads the build's expr, so the command is cheap and works on a build whose backends are unreachable.

```
$ xorq catalog lineage lineage-demo                        # compact (default)
Join(2 inputs) (6 cols) [xorq_datafusion]
|-- Cache[ParquetCache] (4 cols) [xorq_datafusion]   via [Field, JoinReference]
|   `-- -> xorq_datafusion (4 cols)
|       `-- UDXF[OrderEnricher] : 3->4 cols [xorq_datafusion]   (+order_score)
|           `-- xorq_datafusion:raw_orders (3 cols)
`-- RemoteTable[duckdb -> xorq_datafusion] (3 cols)   via [Field, JoinReference]
    `-- duckdb:raw_customers (3 cols)

$ xorq catalog lineage lineage-demo --level boundaries     # tab-separated
@join_chain_1       join             Join(2 inputs) (6 cols) [xorq_datafusion]
@remote_table_10    engine_crossing  RemoteTable[duckdb -> xorq_datafusion] (3 cols)
@database_table_13  table            duckdb:raw_customers (3 cols)
@cached_node_14     cache            Cache[ParquetCache] (4 cols) [xorq_datafusion]
@flight_u_d_x_f_17  flight_udxf      UDXF[OrderEnricher] : 3->4 cols [...]   (+order_score)
# nested  @flight_u_d_x_f_17   1 node   root=@database_table_18

$ xorq catalog lineage lineage-demo -l raw | jq '.nodes[0]'
```

- **compact** is exactly what the TUI's Lineage panel shows (same `format_compact_lineage`), minus the colour.
- **boundaries** is tab-separated with no tree glyphs, for `grep`/`cut`. A `flight_udxf`/`flight_expr` boundary also reports its nested scope on a `# nested` line, so a UDXF's input lineage is discoverable without switching levels.
- **raw** is `LineageDAG.to_dict()` as JSON: every node, not just the boundaries the compact view keeps, plus the `{from,to,scope}` edges.

## Narrowing to one node (`--node`, `0d7abe79`)

The plan's Q3 ("stable expand-by-hash") and Q9 (`resolve(handle)` over `hash | @label | tag | kind | predicate`) were only reachable from Python. `--node` surfaces them; `resolve` needed no change -- it already matches a hash, an id, a kind (qualified or base), then a bare tag value.

```
$ xorq catalog lineage demo --node 2997f938a27d      # content hash (durable)
$ xorq catalog lineage demo --node @cached_node_14   # doc-local label
$ xorq catalog lineage demo --node catalog-source    # tag value
$ xorq catalog lineage demo --node flight_udxf       # boundary kind

$ xorq catalog lineage demo --node flight_udxf
UDXF[OrderEnricher] : 3->4 cols [xorq_datafusion]   (+order_score)
`-- xorq_datafusion:raw_orders (3 cols)              <- nested input scope

$ xorq catalog lineage demo --node flight_udxf --level boundaries
@flight_u_d_x_f_17  flight_udxf  UDXF[OrderEnricher] : 3->4 cols [...]   (+order_score)
# capabilities  @flight_u_d_x_f_17  schema_in_required,schema_in_observed,schema_out,schema_diff,nested
# nested  @flight_u_d_x_f_17  1 node  root=@database_table_18
```

- **compact** renders the subtree *feeding* the node, including a Flight boundary's nested input lineage.
- **boundaries** adds `# capabilities` (the code registry for that kind) and `# nested` scope lines. Capability lines are suppressed on the full listing, where they would double the output with what the kind column already implies.
- **raw** prints a JSON array of the matched node dicts instead of the whole document.
- A handle matching **several** nodes (a kind, a tag) prints every match, blank-line separated -- `--node table` listing all tables is useful. A handle matching **none** fails pointing at `--level boundaries`.

`format_compact_lineage`/`compact_lineage_rows` gain `root=`, with a `_view_containing` helper: a node reached only inside a Flight's nested lineage is absent from the root view, so rooting a subtree at it means resolving its scope first.

## Mermaid output (`--format`, `d4beaff2`)

`--format` defaults to `text`, so nothing above changes. `--format mermaid` is a second renderer over the same compact views:

```
$ xorq catalog lineage demo -f mermaid
flowchart TD
  n__join_chain_1["Join(2 inputs) (6 cols) [xorq_datafusion]"]
  n__flight_u_d_x_f_17["UDXF[OrderEnricher] : 3->4 cols [...]   (+order_score)"]
  subgraph n__flight_u_d_x_f_17_scope["input of UDXF[OrderEnricher]"]
    n__database_table_18["xorq_datafusion:raw_orders (3 cols)"]
  end
  n__database_table_18 --> n__flight_u_d_x_f_17
  n__cached_node_14 -->|via Field, JoinReference| n__join_chain_1
  n__remote_table_10 -->|via Field, JoinReference| n__join_chain_1
  classDef flight_udxf fill:#ffe0f0,stroke:#c2185b,color:#111
  class n__flight_u_d_x_f_17 flight_udxf
```

- **Edges are reversed.** The stored direction is depends-on; a diagram wants data flow, so `TD` puts sources at the top and the final expression at the bottom. A collapsed run becomes the edge label.
- **It keeps the graph**, where the text tree cannot: a node with two consumers is one mermaid node with two inbound edges instead of a repeat marked `↻`.
- **A Flight scope becomes a `subgraph`** -- which is what the stored `scope` tag already means.
- **Its own palette**, not the TUI's theme colours: a diagram gets pasted into docs and issues with their own background.
- Ids are sanitized (`@` is not legal in mermaid), labels are HTML-escaped so one cannot break out of its quoted node, and node order comes from a BFS rather than set iteration -- byte-identical under three `PYTHONHASHSEED`s, so a committed diagram does not churn.
- Composes with `--node`: `--node flight_udxf -f mermaid` diagrams just that node's upstream. A handle matching several nodes emits one `flowchart` block per match, to paste one at a time.

### The un-collapsed graph (`-l raw -f mermaid`, `9efffab6`)

`-l raw` is the stored graph -- every node, including the `Field`/`JoinReference`/`Literal` plumbing that `compact()` folds into `via`. Rendering *that* as mermaid is a third diagram, not a re-spelling of the compact one:

```
$ xorq catalog lineage demo -l raw -f mermaid
  n__join_reference_2["JoinReference (3 cols)"]
  n__field_4["Field:id"]
  n__equals_12["Equals"]
  n__field_4 --> n__equals_12
  classDef unknown fill:#fafafa,stroke:#bdbdbd,stroke-dasharray:3 3,color:#555
```

Non-boundary nodes get a dashed grey class so the boundaries still read as the structure.

`-l boundaries -f mermaid` stays a **UsageError**: keeping its edges would just re-spell `-l compact -f mermaid`, and dropping them draws an edgeless pile of nodes -- a worse listing than the text one. Erroring says the flag did nothing rather than silently picking one. (An earlier revision rejected `raw` on the same grounds; that was wrong, since `raw` really is a different graph.)

### Columns of one node (`--expand HANDLE`, `58b2b084`)

`--node` *narrows* -- it roots the view at the match and drops everything above it. `--expand` keeps the whole view and opens one node up to its columns:

```
$ xorq catalog lineage demo --expand @flight_u_d_x_f_17
...
|       `-- UDXF[OrderEnricher] : 3->4 cols [xorq_datafusion]   (+order_score)
|           |-- in id            int64
|           |-- in amount        float64
|           |-- out + order_score  float64      <- `+`/`~`/`-` from the stored schema_diff
|           `-- ...
```

Every relational node already stores its schema in the sidecar, so `expand_columns` lists those columns as child rows in the text tree, and inside the node's label (`<br/>`-joined) in mermaid, leaving the graph's shape alone. A Flight boundary is the one kind whose schema changes across it, so both sides are drawn (`in `/`out `) and the stored `schema_diff` marks what it added (`+`), retyped (`~`) or dropped (`-`).

Same handle forms as `--node`, and every match expands. `--expand` needs a rendered tree or diagram, so `--level boundaries|raw` with `--format text` is a UsageError (raw JSON already carries every stored schema).

> An earlier revision of this branch made `--expand` switch one *region* of the mermaid diagram to the stored op graph. That reading does not survive in a tree -- it repeats every shared op per consumer, so one expansion of a 681-node graph came to 1207 rows -- and `--level raw` already draws every op, so the mixed-detail dial and its `_stored_expansion` machinery were dropped in `58b2b084` (`format_mermaid_lineage`'s `expand`/`expand_from` args became `raw`/`expand_columns`).

## TUI (`21df8291`, `58b2b084`)

- **3-way right panel.** The Info panel was always on screen, capped at six rows, so a lineage tree deeper than that scrolled inside a sliver while SQL held the tall slot. It becomes the first of three swappable views -- `1` Lineage, `2` SQL, `3` Data -- one full-height slot above the always-visible Schema panel, with Lineage as the startup view. SQL and data previews cost a worker, so only the active view pays for them.
- **Row cursor and column folds.** While the Lineage panel holds focus, `j`/`k` move a row cursor (the panel scrolls to follow), `]` lists the cursor node's columns and `[` folds them back -- the same expansion as the CLI's `--expand`. A two-column gutter (`▸`/`▾`) marks expandable rows; fold state is kept per entry, so browsing to a neighbour and back leaves it as you left it. Rows are cached per fold state, so a keypress re-render does no tree walk.

## Behaviour

- Entry lookup goes through the existing `_get_catalog_entry`, so an **alias works** and a typo points at `xorq catalog list` / `list-aliases`.
- A sidecar with **no lineage** (written before XOR-363) fails with how to get one rather than printing an empty tree.
- The whole body runs inside `click_context_catalog`, as the sibling commands do, so a malformed sidecar reports instead of tracebacking.
- `--level` and `--format` are `click.Choice`s, so an unknown value is rejected by Click with the valid set.

## Verification

`test_cli_lineage.py` (new; 25 tests parameterized across the three catalog backends) + `test_lineage_utils.py` (77) + `test_tui.py` (386) + `test_compiler.py` + `test_graph_utils.py` + `docs/test_cli_contract.py` -- all green locally.

Covers: compact is the default and matches `--level compact`; alias resolves to the same output; `boundaries` rows are 3 tab-separated fields with `@`-prefixed ids, carry the expected kinds and no glyphs, and report the nested scope; `raw` round-trips against the entry's stored `LineageDAG`; unknown level, unknown entry, and a lineage-less sidecar all fail with a useful message. For `--node`: hash and `@label` agree, a tag resolves by value as well as by kind, a kind handle prints one subtree per match, `raw` narrows to the matched nodes, `boundaries` reports capabilities and nested scope (and does not on the full listing), and an unknown handle points at `--level boundaries`. For `--format`: text is the default and matches `--format text`, mermaid emits a `flowchart TD` with a subgraph, edges and classDefs and no tree glyphs, composes with `--node`, and is rejected on the boundaries level; `-l raw -f mermaid` draws strictly more nodes than compact and no `via` labels. For `--expand`: it lists a node's columns in the tree and in the diagram, shows both sides of a Flight boundary's transition with the diff marks, accepts hash/label/kind handles interchangeably, and errors on the listing levels and on an unknown handle. In `test_lineage_utils.py`: `root=` renders the subtree feeding a node, resolves a node that lives only in a nested scope, and degrades to `(empty)` for an unknown id; the mermaid renderer declares every node it classes, declares each node once, points its edges downstream, escapes labels, and is byte-stable across processes. TUI tests cover the view switch, the cursor, fold/unfold, per-entry fold state, and the no-op cases.

## Not changed

The **sidecar format** and `compiler.py`. Nothing new is stored and nothing is re-extracted: this only adds readers (and the TUI presentation of them).

`lineage_utils` does grow, in ways the CLI and TUI share:

- `format_mermaid_lineage` -- the second renderer, with `root=`, `raw=` and `expand_columns=`.
- `root=` and `expand_columns=` on `format_compact_lineage` / `compact_lineage_rows`, plus `node_columns`, `_view_containing`, and `node_id` on `LineageRow` so a renderer can address the node behind a row.
- `_compact_node_label` → public `compact_node_label` (rename only).

`LineageDAG` itself -- fields, serialization, `resolve`/`compact`/`scope` semantics -- is untouched.

## Docs

`docs/generate_cli_reference.py` fails when a visible command is in no nav group, so the new command is added to *Entries and aliases* with See-also links to `show`, `schema` and `tui` (and back-links from `show`/`schema`). The page body comes from the command docstring, so `--help` and the reference site cannot disagree. Docstring lists use the repo's `` - `name`—description `` bullet form (as in `xorq init -t`), which survives both Click's `\b` and markdown; the aligned two-column form collapsed in the generated page.

Refs XOR-363

https://claude.ai/code/session_019xTFQFkPvzLa4cNEc689pg

### DIFF
--- a/docs/generate_cli_reference.py
+++ b/docs/generate_cli_reference.py
@@ -97,6 +97,7 @@ CATALOG_GROUPS = (
             "list",
             "show",
             "schema",
+            "lineage",
             "get",
             "add-alias",
             "remove-alias",
@@ -212,11 +213,18 @@ SEE_ALSO = {
     ),
     "catalog/show": (
         "[`catalog schema`](schema.qmd)—schema-only view",
+        "[`catalog lineage`](lineage.qmd)—lineage-only view",
         "[`catalog list`](list.qmd)—list entries",
     ),
     "catalog/schema": (
         "[`catalog show`](show.qmd)—full metadata for an entry",
+        "[`catalog lineage`](lineage.qmd)—where an entry's data comes from",
         "[`catalog list`](list.qmd)—list available entries",
+    ),
+    "catalog/lineage": (
+        "[`catalog show`](show.qmd)—the rest of an entry's metadata",
+        "[`catalog schema`](schema.qmd)—input and output schemas only",
+        "[`catalog tui`](tui.qmd)—browse the same lineage tree interactively",
     ),
     "catalog/get": (
         "[`catalog add`](add.qmd)—re-add an exported archive to another catalog",
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -23,6 +23,7 @@ from xorq.catalog import constants as catalog_constants
 
 if TYPE_CHECKING:
     from xorq.catalog.content_store import ContentStoreConfig
+    from xorq.common.utils.lineage_utils import LineageDAG
 
 from xorq.cli import (
     _get_cache_dir,
@@ -1018,6 +1019,88 @@ def schema(ctx, name, as_json):
                     click.echo(f"  {col:<24} {dtype}")
 
 
+def _lineage_boundaries_lines(dag: LineageDAG) -> Iterator[str]:
+    """One line per boundary: id, kind, and the compact label's facts."""
+    from xorq.common.utils.lineage_utils import compact_node_label  # noqa: PLC0415
+
+    for node in dag.boundaries():
+        yield f"{node['id']}\t{node['boundary_kind']}\t{compact_node_label(node)}"
+
+
+def _lineage_scope_lines(dag: LineageDAG) -> Iterator[str]:
+    """Nested Flight scopes, rendered under the boundary that owns them."""
+    for node in dag.boundaries():
+        if (nested_root := node.get("nested_root")) is None:
+            continue
+        n_nodes = len(dag.compact(scope=node["id"])["nodes"])
+        plural = "" if n_nodes == 1 else "s"
+        yield f"{node['id']}\t{n_nodes} node{plural}\troot={nested_root}"
+
+
+@cli.command()
+@click.argument("name", shell_complete=_complete_entry_or_alias_names)
+@click.option(
+    "-l",
+    "--level",
+    type=click.Choice(("compact", "boundaries", "raw")),
+    default="compact",
+    show_default=True,
+    help="How much lineage detail to print.",
+)
+@click.pass_context
+def lineage(ctx: click.Context, name: str, level: str) -> None:
+    """Show the lineage recorded in an entry's metadata sidecar.
+
+    The sidecar stores one node table with scope-tagged edges; the compact
+    boundary tree is derived from it, so no level re-walks the expression.
+
+    \b
+    Levels:
+      compact     Boundary-only tree, as the TUI renders it (default).
+      boundaries  One tab-separated line per boundary: id, kind, label.
+      raw         The stored lineage as JSON — pipe it to `jq`.
+
+    \b
+    Arguments:
+      NAME  Entry name or alias.
+
+    \b
+    Examples:
+      # The tree the TUI shows
+      xorq catalog lineage penguins-prod
+      # Grep-able boundary list
+      xorq catalog lineage penguins-prod --level boundaries
+      # Every stored field, for jq
+      xorq catalog lineage penguins-prod --level raw | jq '.nodes[0]'
+    """
+    from xorq.common.utils.lineage_utils import (  # noqa: PLC0415
+        format_compact_lineage,
+    )
+
+    with click_context_catalog(ctx):
+        catalog = ctx.obj.make_catalog(init=False)
+        entry = _get_catalog_entry(catalog, name)
+
+    dag = entry.metadata.lineage
+    if dag is None or not dag.nodes:
+        # A sidecar written before lineage existed, or an expression with none.
+        raise click.ClickException(
+            f"Entry {name!r} has no lineage in its metadata sidecar — "
+            f"rebuild it with a current xorq to record one."
+        )
+
+    match level:
+        case "raw":
+            click.echo(json.dumps(dag.to_dict(), indent=2, default=str))
+        case "boundaries":
+            for line in _lineage_boundaries_lines(dag):
+                click.echo(line)
+            for line in _lineage_scope_lines(dag):
+                click.echo(f"# nested\t{line}")
+        case _:
+            click.echo(format_compact_lineage(dag))
+
+
 @cli.command()
 @click.argument("name", shell_complete=_complete_entry_or_alias_names)
 @json_option
@@ -1029,7 +1112,7 @@ def schema(ctx, name, as_json):
     help="Print the metadata sidecar file as-is (YAML).",
 )
 @click.pass_context
-def show(ctx, name, as_json, as_raw):
+def show(ctx: click.Context, name: str, as_json: bool, as_raw: bool) -> None:
     """Show full metadata for a catalog entry.
 
     Prints name, aliases, kind, backends, schemas, parameters, builders,
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -1064,8 +1064,19 @@ def _lineage_boundaries_lines(
     help="Expand one node: a snapshot hash, an @label, a tag value, or a "
     "boundary kind. Scopes every level to the match(es).",
 )
+@click.option(
+    "-f",
+    "--format",
+    "fmt",
+    type=click.Choice(("text", "mermaid")),
+    default="text",
+    show_default=True,
+    help="Render the compact level as a text tree or a mermaid flowchart.",
+)
 @click.pass_context
-def lineage(ctx: click.Context, name: str, level: str, handle: str | None) -> None:
+def lineage(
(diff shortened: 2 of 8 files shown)


dlovell (approved):
Approved by repository owner (relayed via Claude Code session). Rebased onto main post-#2185; tree object identical to the previously CI-green head 3c222feb.

PR #2189, opened 2026-07-29
chore(deps): bump adbc-driver-gizmosql to >=2.0.0 in gizmosql extra

prmoore77:
## Summary

Bumps the `gizmosql` extra's `adbc-driver-gizmosql` pin from `>=1.1.3` to `>=2.0.0`, and updates `uv.lock` accordingly.

[adbc-driver-gizmosql 2.0.0](https://pypi.org/project/adbc-driver-gizmosql/2.0.0/) is a Go-backed rewrite of the driver with a byte-compatible API — no changes needed in xorq's gizmosql backend. The 2.0 driver no longer depends on `adbc-driver-flightsql`, which drops out of `uv.lock`, and its pyarrow floor was deliberately relaxed to `>=14.0.0` so it resolves cleanly with xorq's `pyarrow<22` cap.

## Test evidence

- Fresh venv `pip install -e ".[gizmosql]"` resolves without conflicts: `adbc-driver-gizmosql==2.0.0` from PyPI, `pyarrow==21.0.0` (satisfies xorq's `<22` pin).
- `uv lock --check` passes.
- Full gizmosql backend test suite against the GizmoSQL Docker container:
  ```
  python -m pytest python/xorq/backends/gizmosql/tests -q
  73 passed
  ```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,7 +132,7 @@ sqlite = [
     "adbc-driver-sqlite>=1.4.0",
 ]
 gizmosql = [
-    "adbc-driver-gizmosql>=1.1.5",
+    "adbc-driver-gizmosql>=2.0.0",
     "duckdb>=1.5",  # see duckdb extra above
     "gizmosql>=1.26.0,<2",
     "cryptography>=42",
--- a/uv.lock
+++ b/uv.lock
@@ -9,69 +9,51 @@ resolution-markers = [
 ]
 
 [[package]]
-name = "adbc-driver-flightsql"
-version = "1.11.0"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "adbc-driver-manager" },
-    { name = "importlib-resources" },
-]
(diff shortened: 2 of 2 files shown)


prmoore77:
A couple of notes on the red ❌'s here:

**1. The `ci-test-lowest-direct` failures are pre-existing on `main`, not caused by this PR.** All four Python versions fail on the same two snapshot tests:

- `python/xorq/ibis_yaml/tests/test_compiler.py::test_build_file_stability_and_relocatability`
- `python/xorq/ibis_yaml/tests/test_compiler.py::test_generated_name_sanitization_memtable`

The most recent `ci-test-lowest-direct` run on `main` (commit `d1afd75`) fails on exactly the same two tests: https://github.com/xorq-labs/xorq/actions/runs/30352948156 — so it looks like a lowest-resolved dependency drifted and the checked-in snapshots no longer match under that resolution. Everything else on this PR is green, including the gizmosql-specific `ci-test` job (`--extra gizmosql`), which passes with the 2.0 driver.

**2. The roborev finding doesn't match the actual diff.** It reports `uv.lock` still pinning `adbc-driver-gizmosql==1.1.7` with the `>=1.1.5` specifier, but this PR's commit updates the lockfile: version `1.1.7` → `2.0.1`, specifier `>=1.1.5` → `>=2.0.0`, and `adbc-driver-flightsql` dropped (the 2.0 driver no longer depends on it). `uv lock --check` passes locally. The bot appears to have read the base-branch lockfile rather than the head commit.

Happy to look into the lowest-direct snapshot drift separately if useful, but I believe this PR is good to go as-is.

PR #2190, opened 2026-07-29
test(rbr_utils): pin copy_rbr_batches as arrow-rs#6471 alignment guard

dlovell:
## What

Adds direct tests for `copy_rbr_batches` and `make_filtered_reader` in `python/xorq/common/utils/rbr_utils.py`, which previously had none.

## Why

Started as an investigation into whether `copy_rbr_batches` — a HAK for [apache/arrow-rs#6471](https://github.com/apache/arrow-rs/issues/6471) buffer misalignment — is still needed, since the upstream bug is marked fixed ([arrow-rs#6472](https://github.com/apache/arrow-rs/pull/6472), Oct 2024) and we now run `xorq-datafusion==0.2.10` / `datafusion>=47`.

**It is still load-bearing, and there's a real (non-synthetic) way to hit it.**

### The natural trigger: decimals over Flight

Ordinary `decimal128` data fetched over Arrow Flight reproduces the bug with no buffer surgery:

- The PyArrow/C++ Flight client hands back record-batch buffers aligned to only **8 bytes**.
- `decimal128` requires **16-byte** alignment (`decimal256` needs 32) — so every decimal data buffer arrives misaligned.
- Iterating the raw Flight chunks (exactly what `make_filtered_reader` does — not `read_all()`, which secretly re-consolidates into aligned buffers) and feeding them into datafusion's `read_record_batches` (a `RecordBatchReader` C-stream FFI import) makes the Rust side panic and **silently drop every batch → 0 rows**.
- `copy_rbr_batches` reallocates each batch through `batch.copy_to(default_cpu_memory_manager())`, producing aligned buffers → all rows ingest.

This is precisely the three live call sites (`relations.py`, `exchanger.py`, `server.py`): data crossing the PyArrow Flight boundary into a Rust consumer. Any `decimal`/`decimal256` column through a Flight exchange is a live landmine without the copy.

### Why the upstream fix doesn't cover us

I read [arrow-rs#6472](https://github.com/apache/arrow-rs/pull/6472) in full. It calls `align_buffers()` in **`RecordBatch::from_pyarrow_bound` only** — the single-batch Python binding path. Reviewer `@tustvold` argued to keep the strict "misalignment = error" contract at the general FFI boundary, and `@alamb` scoped the fix accordingly: *"this PR is very specific to pyarrow ... but not other arbitrary FFI boundaries."* The general `ffi::from_ffi` import — the C Stream Interface path that `RecordBatchReader`/`read_record_batches` uses — was **intentionally left strict**. That's the path xorq relies on, so the workaround is still required.

(PyArrow later added `IpcReadOptions.ensure_alignment` in [apache/arrow#44279](https://github.com/apache/arrow/issues/44279), but it defaults to no realignment and only covers IPC reads, not Flight.)

## Conclusion

The opposite of "safe to remove" — this PR locks the workaround in with a regression test that fails (0 rows) if `copy_rbr_batches` is ever turned into a passthrough.

## Tests (all pass locally)

- `test_copy_rbr_batches_realigns_misaligned_flight_decimals` — the natural guard: real Flight round-trip of `decimal128`, raw-chunk ingest into datafusion. Asserts a precondition that the buffers really are misaligned (so it can't pass vacuously); verified it fails with 0 rows when `copy_rbr_batches` is neutered to identity.
- `test_copy_rbr_batches_preserves_values_and_schema` — the copy is a pure realignment.
- `test_make_filtered_reader_unwraps_flight_chunks_and_drops_empty` — pins its role: unwrap `FlightStreamChunk.data`, drop metadata-only chunks.
- `test_make_filtered_reader_schema_survives_all_empty` — an all-metadata stream still yields a schema-correct, zero-row table.

### DIFF
--- a/python/xorq/common/utils/tests/test_rbr_utils.py
+++ b/python/xorq/common/utils/tests/test_rbr_utils.py
@@ -1,0 +1,165 @@
+from __future__ import annotations
+
+import decimal
+import threading
+from contextlib import closing
+
+import pyarrow as pa
+import pyarrow.flight as flight
+
+import xorq.api as xo
+from xorq.common.utils.rbr_utils import (
+    copy_rbr_batches,
+    make_filtered_reader,
+)
+
+
+class _BatchServer(flight.FlightServerBase):
+    """Minimal vanilla PyArrow Flight server that streams pre-built batches."""
+
+    def __init__(self, location: str, batches: list[pa.RecordBatch]) -> None:
+        super().__init__(location)
+        self._batches = batches
+        self._schema = batches[0].schema
+
+    def do_get(
+        self,
+        context: flight.ServerCallContext,
+        ticket: flight.Ticket,
+    ) -> flight.GeneratorStream:
+        return flight.GeneratorStream(self._schema, iter(self._batches))
+
+
+def _decimal_batches(n_batches: int = 16, rows: int = 7) -> list[pa.RecordBatch]:
+    """Ordinary ``decimal128`` batches.
+
+    ``decimal128`` requires 16-byte buffer alignment, but the PyArrow (C++)
+    Flight client only hands back 8-byte-aligned buffers, so every decimal
+    data buffer arrives misaligned -- no buffer surgery required.
+    """
+    return [
+        pa.record_batch(
+            {
+                # a tiny bool column shifts the following buffer's offset
+                "pad": pa.array([True] * rows),
+                "amount": pa.array(
+                    [decimal.Decimal(f"{k}.0{i}") for i in range(rows)],
+                    pa.decimal128(18, 2),
+                ),
+            }
+        )
+        for k in range(n_batches)
+    ]
+
+
+def _fetch_raw_flight_batches(
+    batches: list[pa.RecordBatch],
+) -> list[pa.RecordBatch]:
+    """Round-trip ``batches`` through a real Flight server/client and return the
+    RAW received chunks (the buffers as the C++ Flight client laid them out).
+
+    We iterate ``reader`` chunk-by-chunk rather than calling ``read_all()`` on
+    purpose: ``read_all()`` re-consolidates into freshly aligned buffers and
+    would hide the misalignment. Iterating ``chunk.data`` is exactly what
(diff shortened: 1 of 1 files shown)


dlovell:
### The `lowest-direct` failures are pre-existing and unrelated to this PR (corrected)

> Correction to an earlier version of this comment: `renovate/pin-dependencies` does **not** pin pyarrow — its diff only pins GitHub Actions in `.github/workflows/*.yml`. That was my error.

The 4 `lowest-direct` (3.10–3.13) failures are two `ibis_yaml` snapshot tests
(`test_compiler.py::test_build_file_stability_and_relocatability`,
`::test_generated_name_sanitization_memtable`). This PR adds only one test file under
`common/utils/tests/`; it cannot affect `ibis_yaml` build hashes.

**Confirmed environmental, with a proven root cause:**

- Re-ran `lowest-direct` on `main` itself (`d1afd759`, this branch's exact base, no test file) → same two failures.
- Same commit `d1afd759` resolved **pyarrow==21.0.0** on 07-28 (green) and **pyarrow==20.0.0** today (red) — verified via the run's attempt-1 log. A transitive constraint change flipped uv's `--resolution lowest-direct` pick 21→20.
- Reproduced locally: building the test's expr under **pyarrow 21** gives the committed snapshot (`build_dir_name 222f26324a6e`); under **pyarrow 20** it gives exactly the CI-failing hashes (`9d8f46b5e7ac`, `expr.yaml e8b4516a…`, `expr_metadata.json 2b382443…`). `profiles.yaml` is identical in both.

**Why version-sensitive:** the expr embeds a `memtable`, which is serialized to `memtables/<hash>.parquet` by pyarrow's writer and gets an `ibis_pandas_memtable_<hash>` node name tokenized from its data — both differ across pyarrow versions. (Deferred parquet reads, normalized by path md5, are not affected.)

**Fix** (separate from this PR): pin pyarrow in the `lowest-direct` job so resolution is deterministic, and/or regenerate the two snapshots. Longer term this looks like a real reproducibility gap worth its own issue — content-addressed build hashes shifting with the pyarrow version.

PR #2197, opened 2026-08-03
feat(skills): add /review-retro for post-review structural retrospectives

dlovell:
## What

The `/review-retro` skill plus the trial review policy that gives it repo residency: CONTRIBUTING gains a "Review policy: shared-infrastructure changes" section (≥2 sequential independent cold reviews; structural retrospective before merge when rounds produce findings; findings-per-round convergence curve as the stop rule), and the skill is its executable half. AGENTS.md carries the agent-facing summary.

## Why residency is by coupling, not location

Community practice scopes project skills to workflows the repo *depends on*, not workflows that merely originated here. A generic retro skill alone didn't meet that bar — the policy paragraph is what creates the dependency (same structure as #2198's blast-radius skill, which the removal convention cites). Skill and policy land together as one priced, revertable unit.

The skill's method is generic; its taxonomy prior is grounded in this repo's history (28 findings across the #2177/#2196 review rounds; ADR-0016 records the code-side outcome) and its routing step names this repo's homes (`docs/adr/`, the CONTRIBUTING sections, AGENTS.md, memory).

## Relationship to #2198

Base is `main`, but the branch is built on `dlovell/process-norms`: the policy paragraph belongs textually adjacent to the process-conventions section that PR introduces, so this PR's diff **includes #2198's commit** until #2198 merges (after which a rebase drops the duplicate and only the review-policy delta remains). Merge #2198 first. Basing on `main` keeps this PR safe from the branch auto-delete that fires when #2198 merges.

## Notes for reviewers

- The policy is deliberately **trial-framed and scoped** (shared traversal/hashing/serialization infrastructure, not all PRs) — it is adopted from n=2 uses; the ask is consent to trial it, with a revisit after a few uses.
- `SKILL.md` doubles as a manual checklist for contributors not using agents.

### DIFF
--- a/.claude/skills/blast-radius/SKILL.md
+++ b/.claude/skills/blast-radius/SKILL.md
@@ -1,0 +1,36 @@
+---
+name: blast-radius
+description: Before removing or renaming behavior, compute the repo-wide blast radius — grep for removed symbols, error strings, and config keys across code, tests, and docs; report every hit that must be updated. Use when deleting code paths (even dead ones), renaming public names, or changing error messages.
+---
+
+# Blast radius of a removal or rename
+
+Given the working diff (or a described removal/rename):
+
+1. Extract from the removed or renamed lines:
+   - function/class/variable names (public AND private — tests import private names),
+   - string literals: error messages, log strings, config keys,
+   - CLI flags, dotted module paths, entry-point names.
+2. For each token, `grep -rn` the whole repo — including `**/tests/**`, `docs/`,
+   and `.github/` — for the exact token AND for distinctive substrings of string
+   literals (error messages are often asserted via `pytest.raises(match=...)`
+   with only a fragment of the message).
+3. Also search for the *module name itself* in other modules' test files:
+   a test that does `monkeypatch.setattr(other_module, ...)` is in the blast
+   radius even when no removed symbol matches textually.
+4. Classify every hit:
+   - (a) already updated by this diff,
+   - (b) must be updated — **blocker**, fix before proceeding,
+   - (c) safe to leave (changelog, historical notes).
+5. Report the classification before making the removal. If there are zero hits,
+   say exactly what was searched so the absence is verifiable.
+
+Rules of thumb:
+
+- Dead code can still have live tests. "Unreachable by construction" does not
+  mean "unreferenced".
+- When retiring a test that guarded a now-impossible drift, replace it with a
+  structural invariant test co-located with the owning module, and leave a
+  pointer comment at the old site.
+- After the removal, re-run the test modules of every file that appeared in
+  category (b), not just the module you edited.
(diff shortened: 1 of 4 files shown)


dlovell:
### Data point from #2192 (canonical memtable hashing), 4 cold rounds

Moving this out of #2192's description, where it was crowding out the change itself, into the issue that actually consumes it.

**Curve:** 4 medium → blocker + 2 major → minors only → 1 medium + 2 minor + 2 nit.

**The stopping rule "one clean round" would have stopped too early.** Round 3 ended with minors only and declared convergence. Round 4 then found a medium: the module's stated layout-invariance contract was not achieved (null-slot payload bytes reach the digest, so `.equals()`-identical columns from `from_pandas` vs literals hash differently), and, more importantly, that the refusal criterion was incoherent — union and run-end-encoded types were refused for "equal data can hash unequally", which describes the accepted residual equally well.

**The pattern across rounds 2–4 is the useful part.** Each landed a finding no predecessor saw, and all three were in the *stated contract* rather than the mechanism:

| Round | Finding | Location |
|---|---|---|
| 2 | extension type behind a dictionary bypassed the refusal walk (`num_fields == 0`) | walk logic, but surfaced by reading the refusal's stated guarantee |
| 3 | `keys_sorted` exclusion applied non-uniformly vs the other constraint flags | stated exclusion policy |
| 4 | invariance contract unmet; refusal criterion not the real criterion | docstrings + ADR |

The mechanism was verified early and stayed correct. What kept breaking was the description of what the mechanism guarantees. Round 4 re-ran the invariants independently (10 type families for slice/chunk, nested dictionary decode through struct/map/fixed-size-list, all four refusals, empty/zero-chunk/zero-column/null-type/duplicate-name edges) and found all of them sound.

**Suggested stopping signal, for the skill:** stop when a round finds only defects in *documentation of intent* and none in behavior — not when severity merely drops. Round 3 satisfied "severity dropped"; it did not satisfy "no behavior-adjacent contract defects", because the null-slot gap was a live mismatch between the contract and the code.

**Two process notes worth encoding in `/review-retro`:**

1. **Corpus construction can hide the defect class the corpus exists to catch.** The null-slot residual was invisible because every nullable fixture (`int64_nulls`, `bool_nulls`) was built from Python literals, which zero the null slots. A checklist item like "for each invariant, is any fixture built by a *different producer* than the others?" would have caught it in round 1.
2. **Reviews should be required to try to withdraw findings.** Round 4 raised a `TypeError` crash on `Path`-constructed sqlite connections, chased it, and withdrew it: `xorq_dasher`'s own sqlite rule already calls `is_in_memory()`, so it is pre-existing and out of scope. A round that reports only confirmations is not obviously a round that verified.

One caveat on reading this curve as evidence: rounds 2–4 reviewed a body of work that was already heavily documented, which biases findings toward the documentation. A sparser PR would likely invert the ratio.

PR #2225, opened 2026-08-12
fix(tui): record query relations in build metadata and order the SQL DAG from them

hussainsultan:
## Problem

The SQL view's DAG renderer (`_render_sql_dag`) guessed dependencies by scanning SQL text with `FROM "([a-f0-9]{20,})"`, which matches no real query name:

- deferred reads are named `ibis_xorq-read_parquet_<hash>` (non-hex prefix)
- remote tables are named `ibis_rbr-placeholder_<token>` (non-hex token)
- `JOIN "..."` references were not scanned at all

With zero edges detected, Kahn's sort degenerated to dict insertion order, rendering `main` first with arrows pointing down into its own inputs. On a real catalog entry:

```
-- [main] (xorq_datafusion)
  ↓
-- [ibis_xorq-re] (xorq_datafusion)
  ↓
-- [ibis_xorq-re] (xorq_datafusion)
  ↓
-- [ibis_xorq-re] (xorq_datafusion)
```

The `name[:12]` label truncation also collapsed every sibling read to the same `ibis_xorq-re`.

## Fix

Record the dependency information in the build metadata instead of parsing SQL text. `generate_sql_plans` already computes each query's referenced `relations`, but `_extract_sql_queries` discarded them:

- `sql_queries` entries in `expr_metadata.json` are now `[name, engine, sql, relations]`
- the TUI builds the DAG from the recorded relations, filtered to known query names; `"main"` is never treated as a dependency, so a source table literally named `main` (e.g. duckdb's default schema) cannot introduce a cycle
- metadata written before relations were recorded (3-element entries) parses with empty relations and keeps the previous stored-order rendering; rebuilding an entry upgrades it
- `_dag_label` truncates only a trailing content hash, so sibling reads stay distinguishable (`ibis_xorq-read_parquet_780c01253639`)

After, on a rebuilt join over two deferred reads:

```
-- [ibis_xorq-read_parquet_566gt4jwgrgavmdlz7bayyy7yi] (xorq_datafusion)
  ↓
-- [ibis_xorq-read_parquet_itgbzmul2vhzrmhz4so3io3l3q] (xorq_datafusion)
  ↓
-- [main] (xorq_datafusion)
```

The RFC (`docs/rfcs/build-time-metadata-caching.md`) is updated to document the new entry shape.

## Tests

- `_extract_sql_queries` records each read's relations and main's relations name the reads
- old 3-element `sql_queries` metadata parses with empty relations and round-trips
- DAG ordering from relations, self/unknown-relation filtering, the `main`-named-table cycle guard, and label uniqueness
- existing SQL render and pilot-driven SQL preview tests pass unchanged

### DIFF
--- a/python/xorq/catalog/cli.py
+++ b/python/xorq/catalog/cli.py
@@ -1276,7 +1276,12 @@ def show(ctx: click.Context, name: str, as_json: bool, as_raw: bool) -> None:
             return
 
         if as_json:
-            click.echo(json.dumps(entry.sidecar_metadata, indent=2, default=str))
+            data = dict(entry.sidecar_metadata)
+            if "expr_metadata" in data:
+                # normalize through the parser so legacy sidecars present the
+                # same sql_queries shape as `schema --json`; --raw stays verbatim
+                data["expr_metadata"] = entry.metadata.to_dict()
+            click.echo(json.dumps(data, indent=2, default=str))
             return
 
         meta = entry.metadata
--- a/python/xorq/catalog/tests/test_cli_inspect_git.py
+++ b/python/xorq/catalog/tests/test_cli_inspect_git.py
@@ -4,6 +4,7 @@ from pathlib import Path
 
 import pytest
 import yaml12
+from click.testing import CliRunner
 from git import Repo as GitRepo
 
 from xorq.catalog.catalog import (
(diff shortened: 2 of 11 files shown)


dlovell:
`_dag_label` no longer truncates names outside the three recognized generated shapes (was always `name[:12]`). Docstring says it's deliberate, to avoid collisions from greedy truncation — but it means a long user-chosen name can now blow out the DAG panel width. Intentional, or worth a bounded truncation with a disambiguating suffix?

(`python/xorq/catalog/tui.py:683`)

dlovell (approved):

Issue #2227, opened 2026-08-13
A failing Flight UDXF hangs the caller, or silently returns zero rows


dlovell:
## Summary

An exception inside a `flight_udxf` exchange does not reach the caller.
Depending on its type and where it is raised, the caller gets one of three
things, none of which is an exception:

1. **an indefinite hang** -- no traceback, and Ctrl-C cannot be delivered
2. **zero rows, silently** -- written into a `ParquetCache` as the answer
3. **a stream that never terminates**, plus a leaked pool thread

> **Status, 2026-08-14.** Modes 1 and 2 are **fixed on `main`** for the two
> single-shot exchangers (#2231). What remains open is listed in
> [Where this stands](#where-this-stands) below. **Three claims in the original
> text were wrong and have been corrected in place** -- if you are implementing
> from this issue, read the status section first; the audit trail for the
> corrections is
> [this comment](https://github.com/xorq-labs/xorq/issues/2227#issuecomment-5292218746)
> and the starvation update is
> [this one](https://github.com/xorq-labs/xorq/issues/2227#issuecomment-5292480290).

Line references are against `main` @ `0b92f31b` (2026-08-14). **Prefer the symbol
names over the numbers**: this issue has already accumulated three mutually
inconsistent sets of `rbr_utils.py` line numbers across its body and comments
(cf. #2203, #2237).

## Where this stands

| Item | State |
|---|---|
| `streaming_exchange` / `streaming_expr_exchange` swallow (modes 1, 2) | **fixed**, #2231 on `main` |
| client drops the error instead of surfacing it | **fixed**, #2231 on `main` |
| `log_excepts` re-swallow at the RPC boundary in debug mode | **fixed**, #2231 on `main` |
| `streaming_split_exchange` (the split path) | **open** -- #2242 (draft): abort is the policy |
| thread-pool starvation | **moved to #2243**, verified and reproduced there |
| metadata-only chunk truncates the stream | **open** |
| `maybe_log_excepts` is inert on `do_action` (a generator) | **open** |
| `instrument_reader` PEP-479, bare `assert` on the schema check | **open** |

## Reproducer (verified; describes pre-#2231 behavior)

Hung forever before #2231:

```python
def boom(df):
    raise SystemExit("missing credential")

udxf = xo.expr.relations.flight_udxf(
    process_df=boom,
    maybe_schema_in=xo.schema({"unit": "int64"}),
    maybe_schema_out=xo.schema({"unit": "int64"}),
[...]

Issue #2155, opened 2026-07-10
Catalog.load() SIGSEGVs on a cross-version entry instead of raising a clear version/format error


dlovell:
## Summary

`Catalog.load(alias)` **SIGSEGVs** (native crash, no Python traceback) when an entry's serialized `expr.yaml` was produced by a *different* xorq version than the one loading it. On an older runtime the same entry instead raises `ModuleNotFoundError`. In both cases the failure mode is wrong: a cross-version / format-incompatible entry should raise a **clear, catchable version/format error**, never crash the interpreter.

This is easy to hit silently because the CLI `xorq catalog run` path (a different code path) works fine on the same entry — so a catalog can look healthy end-to-end while `Catalog.load()` hard-crashes.

## Environment

- `xorq==0.3.33` (also observed on `0.3.28`), uv-managed CPython 3.12, Linux (x86_64), clean container image (reproduced outside any IDE/desktop context).

## Root cause (as far as I traced it)

Each catalog entry serializes its `normalize_method` as a **cloudpickled callable** in `expr.yaml`. `Catalog.load()` reconstructs it via `deserialize_callable → cloudpickle.loads`. The pickled reference points at `xorq.common.utils.file_utils`, whose presence/layout differs by xorq version:

| loading runtime | `xorq.common.utils.file_utils` | `Catalog.load()` result |
|---|---|---|
| 0.3.28 | absent | `ModuleNotFoundError` |
| 0.3.33 | present (moved/changed) | **SIGSEGV** (native crash) |

The entry under test was serialized by an intermediate build (~0.3.30-era). So `cloudpickle.loads` of a callable bound to a module path that has shifted between versions can drive the interpreter into a native crash rather than a clean import error.

Narrowing observations:
- `Catalog.from_repo_path(path)` and `.list_aliases()` **succeed** — only `.load(alias)` crashes.
- CLI `xorq catalog run <alias>` **works** on the same repo on both 0.3.28 and 0.3.33 (it does not go through the cloudpickle `load()` path).
- Authoring the *same* logical entry **natively** under 0.3.33 (data baked as an `xo.memtable`, no cloudpickled callable) `load()`s cleanly — the control showing this is version/format skew, not `load()` being broken in general.

## Expected vs actual

- **Expected:** `Catalog.load()` raises a clear, catchable error identifying a version/format mismatch (or a missing module), so callers can handle it.
[...]

PR #2233, opened 2026-08-13
fix(ml): drop MetricComputation.__module__ property so builds can load

dlovell:
## Problem

Any build embedding a `MetricComputation` is unloadable. `xorq build` succeeds, then the load fails in `xorq/ibis_yaml/common.py:deserialize_callable`:

```
TypeError: can only assign string to MetricComputation.__name__, not 'property'
```

## Root cause

`MetricComputation` defined both `__name__` and `__module__` as properties. The error names `__name__`, but that one is a red herring — the trigger is `__module__`.

- `type.__name__` is a data descriptor on the metaclass, so it shadows the class dict entry: `MC.__name__` still returns the real `str`.
- `type.__module__` has no such protection — its getter returns `cls.__dict__["__module__"]` verbatim, so the property object leaks out at class level:

```python
>>> MetricComputation.__module__
<property object at 0x73fecf19bec0>   # not a str
```

cloudpickle's `_lookup_module_and_qualname` therefore fails, so it pickles the class **by value** instead of by reference, and reconstruction dies in `_class_setstate` doing `setattr(cls, "__name__", <property>)`.

Confirmed with a two-class control at module scope: a class with only `__name__` round-trips under both `pickle` and `cloudpickle`; adding `__module__` breaks both.

## Fix

Remove the `__module__` property, with a comment on `__name__` recording why it must not come back.

The only consumer is `udf.agg.pandas_df`, which copies `fn.__module__` onto the generated node type. That now reads `xorq.expr.ml.metrics` instead of `sklearn.metrics` — metadata only:

- **Expr hash is unchanged**: `2d88eb20b648` before and after.
- A full `build_expr` → `load_expr` round-trip executes and returns the correct result; it raises on `main`.

No other class in the codebase defines `__module__` or `__qualname__` as a property.

## Tests

Three regression tests in `test_metrics.py`, all of which **fail on `main` and pass here**:

- `test_metric_computation_module_is_a_str` — the cheap invariant; fails immediately and legibly if the property is ever re-added
- `test_metric_computation_cloudpickle_round_trip`
- `test_metric_computation_survives_build_serde` — drives the real `serialize_callable` → `deserialize_callable` frames a catalog load hits

`test_metrics.py`: 89 passed. Whole `expr/ml/tests/` dir: 481 passed. `test_split_lib.py` has 7 failures that reproduce on unmodified `main` and are unrelated to this change.

## ⚠️ Does not rescue existing artifacts

The pickle payload in already-written builds embeds the broken class *by value*, so upgrading does not make them loadable — I verified that building on `main` and loading with this fix still raises. **Builds produced by 0.3.39 or earlier that use metrics must be rebuilt.** Since `build` succeeded silently, anyone using metrics has been accumulating unloadable artifacts.

https://claude.ai/code/session_017ozUTdLxjMyUfa9wzAYbbS

### DIFF
--- a/python/xorq/expr/ml/metrics.py
+++ b/python/xorq/expr/ml/metrics.py
@@ -341,14 +341,19 @@ class MetricComputation:
         return dict(self.metric_kwargs_tuple)
 
     @property
-    def __name__(self):
-        """Return the name of the metric function for UDF registration."""
-        return self.metric_fn.__name__
+    def __name__(self) -> str:
+        """Return the name of the metric function for UDF registration.
 
-    @property
-    def __module__(self):
-        """Return the module of the metric function for UDF registration."""
-        return self.metric_fn.__module__
+        Deliberately no matching ``__module__`` property: ``type.__module__``
+        reads ``cls.__dict__["__module__"]`` raw, so a property there leaks out
+        at the class level as a non-str. cloudpickle's module/qualname lookup
+        then fails, it falls back to pickling the class *by value*, and
+        reconstruction dies on ``setattr(cls, "__name__", <property>)`` --
+        making every build embedding a MetricComputation unloadable (#2227).
+        ``__name__`` is safe because ``type.__name__`` is a data descriptor on
+        the metaclass and shadows the class dict entry.
+        """
+        return self.metric_fn.__name__
 
     @staticmethod
     def _normalize_columns(value, name):
--- a/python/xorq/expr/ml/metrics.py
+++ b/python/xorq/expr/ml/metrics.py
@@ -342,16 +342,28 @@ class MetricComputation:
 
     @property
     def __name__(self) -> str:
-        """Return the name of the metric function for UDF registration.
+        """Name the metric function, for naming the generated UDF node type.
+
+        Required, not cosmetic: ``agg.pandas_df`` reads ``fn.__name__``
+        unconditionally to name the node class it builds
+        (``xorq/expr/udf.py``, ``_make_udf_name(fn.__name__)``). Deleting this
+        property raises ``AttributeError`` there -- the ``name=`` argument
+        ``on_expr`` passes covers only the UDF name, not the class name.
 
         Deliberately no matching ``__module__`` property: ``type.__module__``
         reads ``cls.__dict__["__module__"]`` raw, so a property there leaks out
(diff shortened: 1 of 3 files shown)


dlovell:
Review follow-ups pushed in b4eb925b:

- **Citation corrected.** `#2227` is the Flight UDXF error-propagation issue; no issue was ever filed for this bug, so the comment and tests now cite `#2233` — the PR body carries the analysis. (3 sites.)
- **Comment accuracy.** Records that `__name__` is *required*, not cosmetic: `agg.pandas_df` reads `fn.__name__` unconditionally via `_make_udf_name(fn.__name__)` to name the generated node type, and the `name=` that `on_expr` passes does not cover it — deleting the property raises `AttributeError` there (verified). Also records that `__name__` is safe only *given by-reference pickling*: `cloudpickle.register_pickle_by_value(xorq.expr.ml.metrics)` reproduces the identical `TypeError`, so this removed the trigger, not the fragility.
- **End-to-end test added** in `test_build_pipeline.py`, reusing that file's existing round-trip shape. The three original tests all stop at the pickle layer, so nothing covered the actual symptom — `build` succeeding while the artifact it wrote is unloadable.

All four regression tests fail with the `__module__` property restored and pass without it. `test_build_pipeline.py` + `test_metrics.py`: 93 passed.

Independently confirmed the two claims the fix rests on:
- **Build hash unchanged** — `df15652bb91d` pre- and post-fix, on a different expr than the one in the description.
- The yaml `__module__` field does change (`sklearn.metrics._regression` → `xorq.expr.ml.metrics`); the hash does not commit to it, which is what makes this safe.

Three follow-ups filed:
- #2234 — `xorq build` does not verify its own output. Load-back on by default; measured at **1.3 ms** with `lazy=True, read_only_parquet_metadata=True`. `lazy=True` is required — the default `lazy=False` hydrates connections eagerly, so a naive default-on guard would make `build` open remote connections.
- #2235 — no invariant prevents a class from shadowing a pickle-critical dunder. A package-wide walk was run: **1436 classes, zero other violations**, so this fix closed the only instance.
- #2236 — identifying poisoned builds in the wild. The hash can't distinguish them, but the raw pickle bytes can: broken payloads contain `_make_skeleton_class` alongside a `xorq.` module string (14,207 B vs 1,260 B), so a scanner can find them without unpickling — which matters given #2155.

PR #2253, opened 2026-08-15
fix(catalog): source tag records the alias the caller named, not the entry's first

dlovell:
## What

The catalog SOURCE tag recorded the wrong alias: it fell back to `entry.aliases[0]` instead of the alias the caller actually addressed the entry by. An entry addressed as `prod` recorded whichever alias happened to sort first in `catalog.yaml` — provenance for a name the caller never mentioned.

`Catalog.load` already had the right conditional. This lifts it to `Catalog.maybe_alias` and uses it from every path that reaches `_make_source_tag`.

## The two paths fixed

`_resolve_and_execute` sends one entry to `_resolve_single_entry` and two or more to `_compose_expr`. Both were wrong, so both are fixed:

| Path | Command | Before | After |
|---|---|---|---|
| `_compose_expr` fast path | `xorq catalog compose prod trn` | `src` | `prod` |
| `_eval_entry` | `xorq catalog run prod` | `src` | `prod` |

`_eval_entry` needed the alias threaded in explicitly — it receives an already resolved `CatalogEntry`, so the string the caller typed was lost before `_make_source_expr` was reached. `_resolve_single_entry` still has it.

## Not fixed here

`_compose_expr`'s **slow path** (`--rename-params`, `--cache-dir`) emits no source tag at all — it pre-loads the entry, so `_resolve_source` hits its untagged `case Expr()` branch. That is a different defect: adding a tag makes an expression newly eligible for `fuse_catalog_source`, which changes the executed plan rather than just recorded metadata. Tracked in #2255.

Addressing an entry **by hash** still records the entry's first alias. `test_compose_by_hash_falls_back_to_an_arbitrary_alias` pins that so it is a decision on the record rather than an accident — a recorded alias does not tell you the caller asked to follow one, which matters to anything built on this metadata.

## ⚠️ Build hash change

The alias is `HashingTag` metadata and so is tokenized. This shifts the build hash of any CLI compose **or** single-entry run addressed by a non-first alias.

## Tests

Both fixes are parametrized over `first-alias` / `second-alias` on the same entry, so the old fallback gets exactly one of each pair wrong whichever way the alias list is ordered — the pairs pin the behavior without depending on that order.

Each was verified to fail against unfixed code (`second-alias` fails on all three backend parametrizations, `first-alias` passes) rather than merely to pass against the fix. `test_single_entry_records_the_alias_the_caller_named` goes through `_resolve_single_entry` rather than calling `_eval_entry` with an explicit alias, which would have passed either way.

Local: `test_cli_run_compose.py` + `test_bind.py` 244 passed; `test_replay_rebuild*.py` 40 passed.

https://claude.ai/code/session_015Gn9aHyPL8yNnoZsdMepEt

### DIFF
--- a/python/xorq/catalog/catalog.py
+++ b/python/xorq/catalog/catalog.py
@@ -78,6 +78,7 @@ from xorq.ibis_yaml.enums import DumpFiles, ExprKind
 
 if TYPE_CHECKING:
     from xorq.api import Expr
+    from xorq.vendor.ibis.backends import BaseBackend
 
 
 logger = get_logger(__name__)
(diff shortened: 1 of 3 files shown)


dlovell:
Design context: #2261 (late-bound catalog references) lists this PR as a prerequisite. Alias-following reads `(entry_name, alias)` off the SOURCE tag, so which alias gets recorded decides which alias a refresh would follow — that makes this a correctness prerequisite rather than a provenance nicety.

Two things from #2261 that bear on this PR:

- The `test_compose_by_hash_falls_back_to_an_arbitrary_alias` characterization test here is load-bearing for that design. Because `_make_source_tag` falls back to `entry.aliases[0]`, a recorded alias does *not* mean the caller asked to follow one. #2261 argues provenance ("how it was addressed") and intent ("follow this on refresh") need separating before any refresh command ships, and points at this test as the record of the current behavior.
- Provenance is still lost entirely on `_compose_expr`'s other path: with `--rename-params` or `--cache-dir`, the composed expr carries zero catalog tags (verified), because `_resolve_source`'s `Expr` branch does not tag. Out of scope here, but it caps how far this metadata reaches — see #2261 prerequisite 4.

Issue #2263, opened 2026-08-25
ci: unformatted code merges — ci-lint runs ruff check but not ruff format --check


dlovell:
## Summary

`ci-lint.yml` runs `ruff check` but never `ruff format --check`, so unformatted code merges without CI noticing. Five files are currently drifted.

## The gap

`.github/workflows/ci-lint.yml:54-55` is the entire Ruff step:

```yaml
- name: Run Ruff
  run: uv run --no-sync ruff check --output-format=github python examples docs scripts
```

`ruff check` is the linter (rule violations); `ruff format` is the formatter. Different commands, different failure conditions. `.pre-commit-config.yaml` runs both hooks — CI runs only the first, so the formatter is enforced solely by whoever has pre-commit installed locally.

A file can be lint-clean and format-dirty at the same time. Against `origin/main`'s copy of `python/xorq/catalog/tests/test_cli_run_compose.py`:

```
ruff check          → All checks passed!  (exit 0)
ruff format --check → Would reformat
```

## Why "Verify no uncommitted changes" doesn't catch it

`ci-lint.yml:57-65` looks like a backstop but cannot be one here: `ruff check` without `--fix` never modifies files, so `git diff` is empty regardless of formatting. That step guards a different failure — dependency/generated-file churn, per its own hint text.

## Current drift

`ruff format --check python examples docs scripts` → 4 files (a 5th is fixed in #2253):

| File | Landed in |
|---|---|
| `python/xorq/cli.py` | #2258 |
| `python/xorq/ibis_yaml/combine.py` | #2258 |
| `python/xorq/ibis_yaml/tests/test_combine.py` | #2258 |
| `python/xorq/ibis_yaml/tests/test_compiler.py` | #2225 |
| `python/xorq/catalog/tests/test_cli_run_compose.py` | #2258 — reformatted in #2253 |

690 files are already formatted, and every drifted file traces to one of the two most recent merges touching these areas. This is a recent leak, not accumulated debt — which is why the cleanup is cheap and worth doing now.

## Proposed fix

1. **Reformat the 4 remaining files** (`ruff format python examples docs scripts`) as its own commit. Required first — the CI step below lands red otherwise.
2. **Add the check** to `ci-lint.yml`, after the existing Ruff step:

```yaml
- name: Run Ruff format check
  run: uv run --no-sync ruff format --check python examples docs scripts
```

Same paths as the `ruff check` step so the two stay in sync.

## Alternative worth considering

[...]

PR #2274, opened 2026-09-02
chore: format with ruff

ray1097:
Five files had drifted from `ruff format` output: four landed via #2258 and #2225, one via #2253. `ci-lint` runs `ruff check` but never `ruff format --check`, so nothing on the merge path caught them.

Reformatted with the ruff pinned in .pre-commit-config.yaml (v0.15.12).

Kept standalone so the reformat stays reviewable and blame-ignorable apart from the CI change that prevents a recurrence.

refs #2263

### DIFF
--- a/python/xorq/catalog/tests/test_cli_run_compose.py
+++ b/python/xorq/catalog/tests/test_cli_run_compose.py
@@ -1149,7 +1149,8 @@ def test_catalog_join_command_rebinds_same_profile_file_backed_entries(
     assert result.exit_code == 0, result.output
 
     result = runner.invoke(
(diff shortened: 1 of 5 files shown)


dlovell:
see https://github.com/xorq-labs/xorq/tree/dan/chore/ruff-format-drift
~adds an entry in `.git-blame-ignore-revs`, though thinking about it now, perhaps~ this needs to come after the squash merge
also changes `python/xorq/ibis_yaml/tests/test_compiler.py`: protected access pragma is not necessary in test files. this was spurious / unnecessary before you got there (https://github.com/xorq-labs/xorq-style/issues/29)

mesejo (approved):

PR #2278, opened 2026-09-08
fix: reattach the one dead style pragma and gate style on the merge path

### LINKED ISSUE #2276: Remove the 12 dead `xorq-style: disable=` pragmas and add a changed-lines style gate
Every `# xorq-style: disable=` pragma in the tree is currently ineffective, and one of them is masking a live violation. Separately, `xorq-check-style` is not on the merge path at all, which is why nobody noticed.

## The pragmas

Stripping each pragma from a path-preserving copy of its file and re-running the checker shows which ones actually suppress something. Result at `fff2526b`: **12 pragmas, 0 live.**

| File | Line | Rule | Why it does nothing |
|---|---|---|---|
| `python/xorq/expr/remote_table_exec.py` | 102 | `protected-access` | **Detached** — violation reports at line 100 and is live |
| `python/xorq/expr/remote_table_exec.py` | 435 | `protected-access` | Rule does not fire (`expr` is a parameter) |
| `python/xorq/catalog/expr_utils.py` | 85 | `protected-access` | Rule does not fire |
| `python/xorq/catalog/tui.py` | 1927 | `protected-access` | Rule does not fire |
| `python/xorq/catalog/catalog.py` | 1715 | `protected-access` | Rule does not fire |
| `python/xorq/catalog/content_store.py` | 521 | `os-path` | Rule does not fire |
| `python/xorq/ibis_yaml/sql.py` | 42 | `protected-access` | Rule does not fire |
| `python/xorq/ibis_yaml/combine.py` | 203, 243, 281 | `protected-access` | Rule does not fire (`backend` is a parameter) |
| `python/xorq/ibis_yaml/translate.py` | 571 | `protected-access` | Rule does not fire |
| `python/xorq/ibis_yaml/compiler.py` | 439 | `protected-access` | Rule does not fire |

The detached one:

```python
# python/xorq/expr/remote_table_exec.py:98-102
provider = placeholder._find_backend(
    use_default=True
)  # xorq-style: disable=protected-access
```

Suppressions are keyed by the comment's line (102); `ProtectedAccessRule` reports at the `ast.Attribute` node's line (100). Wrapping the call across lines breaks the match, and the violation is live:

```
python/xorq/expr/remote_table_exec.py:100: [protected-access] protected member access `._find_backend` on external object
```

The same wrapping broke four pragmas in `python/xorq/ibis_yaml/tests/test_compiler.py` during the `ruff format` sweep in #2274; those are removed there, since they were doubly dead (`ProtectedAccessRule` skips test files). This issue covers the 12 remaining in non-test code.

**Proposed change:** delete all 12. For `remote_table_exec.py:102`, deleting it exposes the real violation at line 100 — either restructure that call so the receiver and the pragma land on the same line, or leave the violation visible and let it be handled with the rest.

## Why nothing caught it

`xorq-check-style` has exactly one wiring in this repo — a Claude Code `PostToolUse` hook (`.claude/settings.json:9`) matched on `Edit|Write`. It is absent from `.pre-commit-config.yaml` and from every workflow in `.github/workflows/`. Consequences:

- It runs only when Claude edits a file. A human edit, a `sed`, or the `ruff format` commit that detached these pragmas never triggers it.
- In hook mode it scopes to changed lines (`only_lines`, check.py:1285), so a violation stranded on a line nobody is editing stays invisible indefinitely.

## Scale, and why a whole-repo gate is off the table

At `fff2526b`, expanding the 696 tracked Python files under `python examples docs scripts` (the checker skips the 146 under `python/xorq/vendor/` internally) reports **7,275 violations**, dominated by `type-annotations` (5,156) and `pytest-param-id` (1,044). To regenerate:

```console
$ git ls-files 'python/*.py' 'examples/*.py' 'docs/*.py' 'scripts/*.py' \
  | xargs -n 40 uv run --no-sync xorq-check-style | wc -l
```

Note the `xargs`: passing the directories directly reports nothing and exits 0 (xorq-labs/xorq-style#30), so the obvious one-liner understates this by four orders of magnitude.

A gate that has to go from 7,275 to zero is not a gate anyone will land. That leaves a changed-lines gate.

## The changed-lines gate, and its one real limitation

**Stated plainly: a `--diff` gate would not have caught the bug in this issue.** The detached pragma's violation sits on line 100, which no subsequent diff touches — so the gate closes the hole for newly written code and delegates pragma rot entirely to xorq-labs/xorq-style#29. It is only a complete answer if that lands.

With that understood, `--diff` mode itself is sound for CI. Probed against a scratch repo (`scripts/probe_style_diff_mode.sh`):

| Case | Result |
|---|---|
| Violation on an added line | reported, exit 2 |
| Violation on a modified line | reported, exit 2 |
| Violation on an unchanged line of a changed file | not reported, exit 0 — by design, and the limitation above |
| Deletion-only diff | exit 0, no spurious failure |
| Pure rename | exit 0 |
| Empty diff | exit 0 |
| Invoked from a subdirectory | **exit 2 with an actionable usage error** |

The last row matters: unlike the positional-path handling in xorq-labs/xorq-style#30, `--diff` refuses to pass silently when it cannot resolve the diff's files. So #30 does not compromise this gate.

There is one pipeline hazard that does. `git diff` writes to stdin through a pipe, and an empty diff is a legitimate exit-0 pass, so *any* failure of the left-hand side becomes a silent green:

```console
$ git diff origin/main...HEAD | xorq-check-style --diff   # origin/main not fetched
$ echo $?
0
$ set -o pipefail
$ git diff origin/main...HEAD | xorq-check-style --diff
$ echo $?
128
```

Under `actions/checkout` defaults (shallow, single branch) `origin/main` is frequently absent, which is exactly this case. So the gate needs both:

- `set -o pipefail` (or an intermediate file with a checked exit status), and
- `fetch-depth: 0` on `actions/checkout`, since a shallow clone has no merge base to diff against.

Without those two, the job is green regardless of what the diff contains — the same failure class the gate exists to prevent.

## Proposed gate

```yaml
- name: Style (changed lines)
  shell: bash
  run: |
    set -euo pipefail
    # requires actions/checkout with fetch-depth: 0 — a shallow clone has no
    # merge base, and `git merge-base` failing must not be swallowed by the pipe
    base=$(git merge-base HEAD "origin/$GITHUB_BASE_REF")
    git diff "$base" HEAD | uv run --no-sync xorq-check-style --diff
```

Whether this belongs in `ci-lint.yml` alongside `ruff check`, or in `.pre-commit-config.yaml` where the diff is not available in the same form, is the open question. Ratcheting toward a whole-file gate — enabling only rules already at zero and tightening over time — is the direction of travel but is a separate decision.

## Related

- #2274 — the `ruff format` drift cleanup that surfaced this
- #2263 — adds a `ruff format --check` gate; different hole, same root cause of "no gate on the merge path"
- xorq-labs/xorq-style#29 — upstream `unused-suppression` rule; prerequisite for this gate to cover pragma rot
- xorq-labs/xorq-style#30 — silent exit-0 on directory arguments; affects whole-file invocation, not `--diff`

mesejo:
Closes #2276, but not as specified: the issue's premise does not hold, and the change it proposes would delete eleven live suppressions. The correction is in "Why".

## What

`xorq-check-style` was installed but never ran on the merge path, so every rule it implements was advisory. This puts it there as two gates, and reattaches the one pragma that genuinely suppresses nothing.

| File | Change |
| --- | --- |
| `.github/workflows/ci-lint.yml` | Whole-repo gate on the rules already at zero; changed-lines gate, pull requests only; `fetch-depth: 0`; `LINT_PATHS` as the tree list |
| `scripts/check-style-diff.sh` | The changed-lines gate itself. CI and the pre-commit hook both run this one script, so they cannot drift |
| `.pre-commit-config.yaml` | `xorq-check-style` hook over the staged diff, calling that script |
| `scripts/style_tests/` | One deliberately broken file per rule, plus six tests that read the gate's own config rather than restate it |
| `pyproject.toml` | `[tool.xorq-style.unlisted-import] src-roots`. The package lives under `python/`, which the `("src", ".")` default does not cover, so the rule resolved no module and passed every import |
| `python/xorq/expr/remote_table_exec.py:100` | Pragma moved back onto the line it suppresses |
| `CONTRIBUTING.md` | How to run each gate locally, how to suppress a line, and why a comment must not claim parity that nothing enforces |

Two rule sets, not one. The whole-repo gate runs on every build over the 5 rules with zero violations anywhere, unchanged lines included, which the diff form structurally cannot reach. The changed-lines gate runs every rule except the four bulk ones, so pre-existing violations in a file you edit stay out of your way. The `--disable` lists are the ratchet: a rule comes off when its count hits zero, and a test fails to ask for it.

## Why

Nothing stopped a violation from landing. Worse, a suppression can rot in place: `ruff format` re-wrapped the call at `remote_table_exec.py:100` and stranded its pragma two lines below, leaving a live violation covered by nothing.

#2276 reads that as twelve dead pragmas and proposes deleting all of them. Stripping each from a path-preserving copy and re-running the checker gives 11 of 12 live. The cause of the bad table is that violations go to stderr (`check.py:1346`) while `--json` and `--list` go to stdout, so a harness capturing stdout sees an empty result and reads every pragma as dead. The issue's own baseline recipe has the same defect and returns 0, not 7,275. Only the four pragmas in `ibis_yaml/tests/test_compiler.py` are dead, since the rule skips test files, and #2274 already removed them.

What was rejected:

- **Every rule on changed lines**, as #2276 specifies. Measured over the last five commits on main: 18 violations, every one from the bulk backlog (14 `type-annotations`, 3 `pytest-param-id`, 1 `future-annotations`). That reddens nearly every pull request touching a signature or a parametrize list. Those four rules are 6,794 of the 7,360 total and stay off.
- **Restructuring the call so receiver and pragma share a line**, also #2276's suggestion. At 102 characters `ruff format` re-wraps it straight back into the shape it started in.
- **The ratchet in `[tool.xorq-style] disable`** rather than a CI flag. That applies to every invocation, the editor hook included, so those rules would go unreported on new code too and their counts could never fall.

One correction to the proposed gate: `ci-lint.yml` also runs on push to main and on tags, where `GITHUB_BASE_REF` is empty. The snippet in #2276 would resolve `origin/`, fail, and with `set -euo pipefail` correctly present, redden every push to main. Guarded with `if: github.event_name == 'pull_request'`.

## Verification

| Check | Result |
| --- | --- |
| `pytest scripts/style_tests -q` | 32 passed |
| Whole-repo gate on this branch | exit 0, 707 files, 0 violations |
| Changed-lines gate on this branch | exit 0 |
| Enforced set (26 rules, 21 ratcheted) | `dataclasses`, `attrs-mutable-default`, `pytest-mark-qualify`, `strenum-compat`, `leaf-enum-import`, all at 0 across 707 files |
| Ratcheted rules still carry violations | 21 of 21, asserted by `test_ratcheted_rules_still_have_violations` |

The positive controls exist because `xorq-check-style` exits 0 both on a clean file and on one it never examined, so a broken rule looks exactly like a satisfied one. `test_rule_fires_on_its_fixture` gives all 26 rules a file each must flag. `unlisted-import` is the reason: before `src-roots` was set it sat on the enforced list resolving nothing, reporting a clean tree it never opened. It now sits on the ratchet with 71 real violations.

The remaining tests read configuration instead of duplicating it. `test_lint_paths_agree` compares the three copies of the linted-tree list (workflow env, script `paths`, pre-commit ruff args) and names the first that disagrees. `test_ratchets_name_real_rules` checks both `--disable` lists against `--list`, so a typo or a rule renamed upstream fails here. `test_changed_lines_gate_is_the_stricter_one` checks the whole-repo ratchet contains the diff ratchet. `test_every_rule_has_a_fixture` fails when a dependency bump adds a rule, rather than letting it join the enforced set unannounced.

## Not addressed

- No public API change. The only source edit is a moved comment.
- Neither gate catches pragma rot on a line nobody touched, which is exactly the bug found here. That needs xorq-labs/xorq-style#29 (unused-suppression).
- The other eleven pragmas stay. They suppress live violations.
- The 7,360-violation backlog is untouched. The ratchet is the mechanism for draining it, one rule at a time, not a drain.
- Violations still go to stderr while `--json` goes to stdout. Worth filing upstream, since that split is what made #2276 wrong, but nothing here depends on it.

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -26,7 +26,11 @@ jobs:
       matrix:
         python-version: [ "3.13" ]
     steps:
+      # fetch-depth: 0 for the changed-lines style gate below --- a shallow
+      # clone has no merge base with the target branch.
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       # Prose is listed explicitly rather than scanning the tree, so a
       # generated or vendored file cannot fail the build over wording nobody
@@ -53,6 +57,37 @@ jobs:
       # Update output format to enable automatic inline annotations.
       - name: Run Ruff
         run: uv run --no-sync ruff check --output-format=github python examples docs scripts
+
+      # Files are expanded with `git ls-files` because xorq-check-style reports
+      # nothing and exits 0 when handed a directory (xorq-labs/xorq-style#30).
+      #
+      # --disable is the ratchet: every rule listed has a nonzero count today.
+      # Move a rule off the list once it reaches zero and the gate tightens.
+      - name: Style (rules at zero, whole repo)
+        shell: bash
+        run: |
+          set -euo pipefail
+          git ls-files 'python/*.py' 'examples/*.py' 'docs/*.py' 'scripts/*.py' \
+            | xargs -n 40 uv run --no-sync xorq-check-style --disable \
+              type-annotations,pytest-param-id,future-annotations,print,pytest-tmp-path,protected-access,deferred-import-test,redundant-import,os-environ,init-reexport,deferred-stdlib,import-aliasing,stdlib-logging,os-path,test-class,exception-placement,relative-import,exception-hierarchy,enum-placement,cache-method
+
+      # Pull requests only: GITHUB_BASE_REF is empty on push and tag builds, so
+      # `git merge-base HEAD origin/` would fail and redden every push to main.
+      #
+      # pipefail is load-bearing --- an empty diff is a legitimate exit-0 pass,
+      # so a failure left of the pipe otherwise reads as clean.
+      #
+      # The four disabled rules are 93% of the backlog; with them on, touching
+      # any signature or parametrize list reddens the PR.
+      - name: Style (changed lines)
+        if: github.event_name == 'pull_request'
+        shell: bash
+        run: |
+          set -euo pipefail
+          base=$(git merge-base HEAD "origin/$GITHUB_BASE_REF")
+          git diff "$base" HEAD \
+            | uv run --no-sync xorq-check-style --diff --disable \
+              type-annotations,pytest-param-id,future-annotations,print
 
       - name: Verify no uncommitted changes
         run: |
--- a/python/xorq/expr/remote_table_exec.py
+++ b/python/xorq/expr/remote_table_exec.py
@@ -97,9 +97,9 @@ def count_remote_table_readers(expr: Expr) -> dict[RemoteTable, int]:
     if not sentinels:
         return {}
     try:
-        provider = placeholder._find_backend(
+        provider = placeholder._find_backend(  # xorq-style: disable=protected-access
             use_default=True
-        )  # xorq-style: disable=protected-access
+        )
         compiler = getattr(provider, "compiler", None)
         if compiler is None:
             return {}


dlovell:
an unset `src-roots` defaults to `('src', '.')`, we need to set it in `pyproject.toml`

PR #2282, opened 2026-09-11
feat: add sidecar semantic metadata via Catalog.add(metadata)

dheerapat:
`Catalog.add(..., metadata={...})` stores catalog-level metadata in the entry sidecar, exposed as `entry.semantic_metadata` and `show --json`.

Sidecar metadata (domain/owner/tags) shouldn't change the content hash like BSL/expr metadata does. Sidecar keeps it queryable + git-tracked.

 Verify:

 ```
   uv run examples/catalog_semantic_metadata_demo.py
   uv run xorq catalog --path /tmp/sales-catalog show sales-by-region --json
   uv run --with pytest pytest python/xorq/catalog/tests/test_catalog.py -k semantic_metadata -q
 ```

### DIFF
--- a/docs/tutorials/core_tutorials/build_a_semantic_catalog.qmd
+++ b/docs/tutorials/core_tutorials/build_a_semantic_catalog.qmd
@@ -236,7 +236,16 @@ catalog_dir = Path("flights-catalog")
 catalog = Catalog.from_repo_path(catalog_dir, init=True)
 
 # <3>
-catalog.add(flights_model_expr, aliases=("flights-model",), sync=False)
+catalog.add(
+    flights_model_expr,
+    aliases=("flights-model",),
+    metadata={
+        "domain": "flights",
+        "owner": "analytics",
+        "tags": ["semantic-model", "tutorial"],
+    },
+    sync=False,
+)
 
 print("Catalog at:", catalog_dir.absolute())
 print("Aliases:   ", catalog.list_aliases())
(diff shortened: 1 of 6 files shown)


dheerapat:
> ## roborev: Fail
> Review findings
> Good, that import is fine. I have enough to finalize the review.
> 
> ## Review Findings
> * **Severity**: Medium
> * **Location**: `python/xorq/catalog/catalog.py:1379` (`CatalogAddition._add`)
> * **Problem**: When `catalog.add(..., metadata=..., exist_ok=True)` targets content that already exists in the catalog (same content hash), `_add` takes the early-return branch that only re-adds aliases and never writes `self.metadata` (including the new `semantic_metadata`). Any metadata passed on a call that hits an existing entry is silently discarded with no error or warning, even though the caller explicitly supplied it.
> * **Fix**: Either raise/warn when `semantic_metadata` is provided but the entry already exists and metadata differs, or update the sidecar's `semantic_metadata` on the existing entry in the `exist_ok` path.
> 
> * **Severity**: Low
> * **Location**: `python/xorq/catalog/catalog.py:86-97` (`_normalize_semantic_metadata`)
> * **Problem**: The function is documented/used as validating a metadata _mapping_, but it accepts any JSON-like value at the top level (e.g. `catalog.add(expr, metadata=["a", "b"])` normalizes successfully instead of raising), which is inconsistent with the docs ("Use JSON/YAML-compatible values with string mapping keys") and with `CatalogEntry.semantic_metadata` being described/consumed elsewhere as a dict (e.g. CLI JSON output).
> * **Fix**: Enforce that the top-level `metadata` value is a `dict` (or `None`) before recursing, raising `TypeError` otherwise.
> 
> * **Severity**: Low
> * **Location**: `python/xorq/catalog/catalog.py:402-422` (`Catalog.add` docstring)
> * **Problem**: The docstring for `add()` documents `obj`, `project_path`, and `relocate_reads` but was not updated to document the new keyword-only `metadata` parameter.
> * **Fix**: Add a paragraph describing `metadata`'s purpose, type constraints, and its relationship to `CatalogEntry.semantic_metadata`.
> 
> ## Summary
> Adds optional, validated sidecar `semantic_metadata` support to `Catalog.add()`/`CatalogEntry`, threaded through zip/build-dir/expr add paths and exposed via CLI `--json` output, but metadata is silently dropped on the `exist_ok` dedup path and top-level type validation/documentation are incomplete.
> 
> _Review type: default | Agent: claude-code | Job: 1022_

addressed

